PR #93, opened 2017-08-18
added option to apply to another channel other than 1

tp81:


Svidro:
Thanks!  If you don't mind my asking, what do you use this for?  Are you looking for a quicker way to select only tissue stained a certain way?  Dealing with background or artifacts?

tp81:
Hi

Not really, I'm dealing with 2-channel 16-bit fluorescent images and the
DAPI stain (which I also use as 'tissue' marker) is in channel 2!

Cheers
Thomas

Svidro:
That sounds much easier, once integrated, than my method of downsampling the whole image to a size where I could submit it to ImageJ, greyscaling a merged image of the channels, and then sending back the annotation region! I haven't tested it yet, but can it work on multiple channels at once?

Svidro:
Hi again!  I finally went and edited the changes into my code in Eclipse, and while the channel option shows up and does not cause any errors, I am not seeing the channel change have any effect on the result.  Has anyone else tested this and had it work?  I am not sure if I am doing something wrong or missed something.  I tried it with a few files from the OpenSlide site, like Mirax2-Fluorescence-1.

Svidro:
Ah, apparently the channel swap doesn't apply to 8bit images!  So much to learn.

tp81:
Hi

the channel selection works by using the setC method on an ImagePlus, so it
would work whenever the image is recognized as a "composite" or a
"hyperstack". It should work with 8-bit, provided the image at that point
is seen as a composite or a hyperstack. The call to setC needs to be made
before the getByteProcessor call. In particular, it does not work with RGB
images, in the sense that RGB images are not composite/hyperstack by
default, so the setC would have no effect. The way around it is to precede
the call to setC with a call to "ij.plugin.CompositeConverter.makeComposite
<https://imagej.nih.gov/ij/developer/api/ij/plugin/CompositeConverter.html>",
to convert it to a Composite image. I can add a check and the call to
compositeconverter.

Cheers
Thomas

petebankhead:
Oh dear, sorry, I see now I didn't reply to this (although I was away from QuPath at the time and I think we discussed it elsewhere...).

I recall at the time I wondered about how this would impact supporting RGB/non-RGB images with the same command and also maintaining scripting compatibility.

In the meantime, the code has diverged rather a lot and other issues with _Simple Tissue Detection_ have emerged, e.g.
* https://github.com/qupath/qupath/issues/124
* https://github.com/qupath/qupath/issues/248

I'm reluctant to try to resolve the code conflicts to incorporate this small change that may complicate scripting compatibility whenever it looks like the whole simple detection command really needs a thorough overhaul. There are now also other ways to gain more control over tissue detection, e.g.
* https://petebankhead.github.io/qupath/scripting/2018/03/08/script-imagej-to-qupath.html
* the pixel classifier (not yet complete, but I hope it will become a 'standard' way)

The pixel classifier is also being designed to support different kinds of classification, which could eventually also include a simple threshold applied to an original or transformed image. It has the benefit of allowing the classification to be applied at a higher resolution through tiling, and to interactively show preliminary results (e.g. with a threshold slider). I think that this is needed in the longer term, and _Simple tissue detection_ will move into retirement.

Sorry again for not replying here sooner. If this change is still of interest to you, please feel free to reopen the issue. To integrate it, we'd need to
* update the code to be compatible with the current codebase
* test the impact on scripts created before/after the change

PR #134, opened 2018-01-10
Changes arbitrary limit to max pixel size being sent to ImageJ from macro runner

lacan:
Next change should be sending the whole image if no selections exist. 

Best

Oli

### DIFF
--- a/qupath-core-processing/src/main/java/qupath/lib/projects/Project.java
+++ b/qupath-core-processing/src/main/java/qupath/lib/projects/Project.java
@@ -102,7 +102,8 @@ public class Project<T> {
 		
 		boolean changes = false;
 		for (String name : subImages)
-			changes = changes | addImage(new ProjectImageEntry<>(this, server.getSubImagePath(name), name, null));
+			// The sub image name might be the same across images, we should append the server displayed name to it, just to make sure it is unique
+			changes = changes | addImage(new ProjectImageEntry<>(this, server.getSubImagePath(name), server.getDisplayedImageName()+" ("+name+")", null));
 		return changes;
 	}
 	
--- a/qupath-extension-ij/src/main/java/qupath/imagej/plugins/ImageJMacroRunner.java
+++ b/qupath-extension-ij/src/main/java/qupath/imagej/plugins/ImageJMacroRunner.java
@@ -226,7 +226,7 @@ public class ImageJMacroRunner extends AbstractPlugin<BufferedImage> {
 		
 		// Check the size of the region to extract - abort if it is too large
 		double approxPixelCount = (region.getWidth() / region.getDownsample()) * (region.getHeight() / region.getDownsample());
-		if (approxPixelCount > 5000L*5000L) {
+		if (approxPixelCount > 25000L*25000L) {
 			DisplayHelpers.showErrorMessage("ImageJ macro error", "Requested region is too large to send to ImageJ (approx " + (int)(region.getWidth() / region.getDownsample()) + " x " + (int)(region.getHeight() / region.getDownsample()) + " pixels) - try again with a smaller region, or a higher downsample factor");
 			return;
 		}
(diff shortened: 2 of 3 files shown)


petebankhead:
I definitely agree that the current limit is bad.  But the fix could result in more out-of-memory errors.  I think it would be good to good to give some protection from these (or a more informative error if they occur), and aim for a more substantial fix.

Two changes could help:
* Choose a pixel count threshold based upon both the maximum pixel count that ImageJ could conceivably support, i.e. 2<sup>31</sup>-1 (or a little more than 1) *and* also a memory threshold based on estimating the image size and memory available
* Add a scaling factor as a static field to control what proportion of the available memory is permitted for use.

It could be something like this:
```java
static double MEMORY_THRESHOLD = 0.5;
...
long remainingMemory = ... // With the help of Runtime.getRuntime(), after request for garbage collection
long approxMemory = ... // Calculation based on pixel count, bit-depth & number of channels
if (approxPixelCount > 2147480000L || approxMemory > remainingMemory * MEMORY_THRESHOLD) {
   // Show the size error
} else {
   // Have a go at extracting the region
}
```
By making the scaling factor a static field, there is at least the option of overriding it in a Groovy script if absolutely necessary.

The details around the pixel count get a bit fiddly, considering that ImageJ1 will use a packed int array in a `ColorProcessor` for RGB images, while in other cases a new array will be allocated for every channel.  Furthermore, it might be prudent to take parallelization into account, and the result returned by `PathPrefs.getNumCommandThreads()`.

I'm dubious about sending the whole image if no selection exists, because it's really easy to accidentally run the command without an area selected... and then everything can grind to a halt while QuPath tries (and possibly fails) to extract a massive region.  Currently you can use *Objects &rarr; Create full image annotation*, its shortcut `Ctrl + Shift + A` or the scripting command `createSelectAllObject(true);` to create an annotation representing the full image, where the `true` indicates that this annotation should be selected automatically.

lacan:
Hi Peter, 

Thank for the very complete message! I'll go and implement this now and submit it to see if it's to your liking.

Right now I have placed my version of the ImageJMacroRunner in our QuPath, so that I could process the images I am receiving.

> Furthermore, it might be prudent to take parallelization into account, and the result returned by PathPrefs.getNumCommandThreads().

I am not sure how you want to address this, each command thread might take an image of a different size, and these threads to not talk to each other. In a case like this, it would make more sense to find a way to gracefully show some sort of out of memory error in case this happens.

> Currently you can use Objects → Create full image annotation
Perfect, then you're right, no need to send the whole image by default if it's just one command! I was making a RectangleROI with the dimensions of the image server in my script...

lacan:
And _voilà._ 

Works hunky-dory for me. Hope it is to your liking for now. 
Next steps would be adding the `memory_threshold` to the preferences, and making use of it elsewhere, but for now I though it made sense to have it in IJTools.

petebankhead:
Looks good!  I won't get a chance to try the code for a few days, but I will soon - sorry for the delay...

Since you mention adding it to the preferences, this is how it works:
```java
import javafx.beans.property.DoubleProperty;
import qupath.lib.gui.QuPathGUI;
import qupath.lib.gui.prefs.PathPrefs;

DoubleProperty preference = PathPrefs.createPersistentPreference("ijFreeMemory", 0.5);
QuPathGUI.getInstance().getPreferencePanel().addPropertyPreference(
        preference, Double.class, "Extract region memory limit", "ImageJ",
        "The proportion free memory that ImageJ is allowed to use when extracting regions (set > 1 to use as much memory as possible)");
```

And your reference to using it elsewhere reminds me of the command to extract regions [here](https://github.com/qupath/qupath/blob/v0.1.2/qupath-extension-ij/src/main/java/qupath/imagej/gui/commands/ExtractRegionCommand.java#L120).  Looking back at that code, it seems to contain an error in using totalMemory rather than the proper combination of max/total/free that you used.  That is probably another candidate for improvement, and for calling your new method...

petebankhead:
Actually, on reflection, there isn't really anything ImageJ-specific in your method, making it a good candidate to go into one of the 'core' modules.  I could see it being useful in other places - potentially with different memory proportion limits - such as when extracting images to write to disk, or possibly even to throttle parallelization when running a detection command.

lacan:
Well, I decided to put it there based on your discussion that ImageJ used packed int arrays for RGB images, so that perhaps wasn't relevant to other parts of your code. But if you think it can be refactored elsewhere, that's fine by me.

lacan:
this fixes now the incorrect series names with VSI files that have sub images!

PR #324, opened 2019-05-30
Added maven plugin

arvgord:
For compatibility with maven-built extensions.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -161,6 +161,7 @@ allprojects  {
  * Build all projects
  */
 allprojects {
+  apply plugin: 'maven'
   apply plugin: 'java'
   sourceCompatibility = JavaVersion.VERSION_11
   targetCompatibility = JavaVersion.VERSION_11


petebankhead:
Can you elaborate on what this does and why it would help to include in the main repo?

arvgord:
Because I have built-in dependencies in pom.xml(in my extension for QuPath) from qupath-gui-fx. For resolving these dependencies I need to install qupath libraries to local maven repository. But last QuPath version built on gradle and I can't do this without maven plugin in gradle.built.

arvgord:
It will be useful for other developers who maintain maven-based extensions for QuPath.

petebankhead:
I'm not sure that there are (many/any?) other developers maintaining maven-based extensions for QuPath, and gradle is the recommended approach. I'm reluctant to add something like this to the main build script when it isn't actually needed to build QuPath itself, and many people (including me) may prefer not to install the jars to their local maven repository.

In any case, I understand there are other ways to install the jars locally, e.g. https://maven.apache.org/guides/mini/guide-3rd-party-jars-local.html

Currently I am refactoring and aiming to reduce as much of complexity as possible to make the software more maintainable. It may be worth revisiting the use of Maven when this is complete.

PR #425, opened 2020-03-30
Implemented MeasurementExporter + GUI command & created ProjectDialogs

melvingelbard:
- Created MeasurementExporter, to make the export of (annotation/detection) measurements easy.
- Created MeasurementExportCommand, a simple GUI for MeasurementExporter.
- Created ProjectDialogs, which contains a single method for displaying ListSelectionViews with project image entries.
- Refactored DefaultScriptEditor so it now uses the method in ProjectDialogs.

### DIFF
--- a/qupath-gui-fx/src/main/java/qupath/lib/gui/QuPathGUI.java
+++ b/qupath-gui-fx/src/main/java/qupath/lib/gui/QuPathGUI.java
@@ -194,6 +194,7 @@ import qupath.lib.gui.commands.DistanceToAnnotationsCommand;
 import qupath.lib.gui.commands.EstimateStainVectorsCommand;
 import qupath.lib.gui.commands.LoadClassifierCommand;
 import qupath.lib.gui.commands.LogViewerCommand;
+import qupath.lib.gui.commands.MeasurementExportCommand;
 import qupath.lib.gui.commands.MeasurementManager;
 import qupath.lib.gui.commands.MeasurementMapCommand;
 import qupath.lib.gui.commands.MemoryMonitorCommand;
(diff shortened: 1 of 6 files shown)


petebankhead:
This looks good and very useful! Nice touches with auto-updating the image name based on the delimiter, and auto-disabling the column include/exclude as needed :)

Some small things:
* Beware invalid characters in the fields... if the filename contains commas, the entire table becomes shifted if exporting as .csv. It appears to work properly (at least in MS Excel) if the name is enclosed in "inverted commas". There are rules for this [here](https://en.wikipedia.org/wiki/Comma-separated_values#Basic_rules).
* Avoid including spaces when before/after the delimiter.
* Try to get the controls to align at the right, and fill the entire space at the bottom of the dialog. For example, for the combo boxes you'll need to set the max width to Double.MAX_VALUE and (assuming you're using a GridPane) some other grow/fill properties (I forget which).
* Rename 'Choose path' to 'Choose...' and 'Output File' to 'Output file' (for capitalization consistency), and add tooltip text to this row.
* Change 'Apply on' to 'Export type' or similar. Try adding 'Image' as an option, i.e. the root object, and perhaps also TMA cores. 'Image' should probably be the default.
* Add tooltip to the 'Output File' option (and rename to 'Output file' for consistency)
* Fix the tooltip for 'Columns to exclude' (says 'include')
* Revise the javadocs for 'ProjectDialogs'; add for getSourceItems, and standardize for createImageChoicePane (i.e. use the `@param` tags)

Not essential at this point, but after playing with it a bit I think there would be a way to improve the control even further: the choice of columns could be a CheckComboBox, which is populated *if* the user presses a *Populate* button. This needs to be an explicit choice, because it would require opening all the currently-selected images to figure out which column headings are available... and this may be slow. Perhaps not worth the effort for now, having this at all will be a major improvement!

PR #457, opened 2020-04-27
Fixed CompressionType display

melvingelbard:
Fixed display of CompressionType in QuPath while keeping the original 'simple' display in the CLI.


petebankhead:
Looks good! Resolved one of https://forum.image.sc/t/0-2-0-m10-random-musings/36745

PR #472, opened 2020-05-06
Added Escape key to command finder

melvingelbard:


petebankhead:
I think you also need to consume the event, to ensure it doesn't filter down somewhere else.

Issue #465, opened 2020-04-30
Unsaved changes to project are lost without warning when closing QuPath without closing project


cfmoro:
Hi,

When changing e.g. the image name in a project and then exiting QuPath by File > Quit or closing the window (x) the changes are lost without warning. A warning save dialog (as with the individual images) asking to save changes to project would be useful. 

Changes are correctly saved exiting by File > Project > Close project.

Best wishes,
Carlos

petebankhead:
Thanks @cfmoro, I'll close this since it should now be fixed in v0.2.0-m12. Just let me know if you find the issue remains.

Issue #405, opened 2020-02-24
Detection Centroid Distance interacts awkwardly with multiplexing


saramcardle:
I love the new multiplex object classification in m9! However, when you calculate detection centroid distances on objects with multiple classifications, the results are difficult to interpret. It gives a separate measurement for the distance to the nearest cell of each permutation of classes individually. So, it gives the shortest distance to a cell of classA: classB, then separately to classA: classC, etc. But often what I want to know is the distance from a cell to another of classA of any subtype. Can you add a summary measurement for the shortest distance to any cell of classA, maybe like a classA*?

Svidro:
In case this is still an issue for one of your projects, I think the best workaround would be to run each of the individual classifiers and the distance to detections right after it by script using

https://forum.image.sc/t/scripting-json-classifiers-in-qupath-0-2-0-m9/34614/2

followed by detectionCentroidDistances()

Both functions tend to run pretty fast, so it shouldn't be too problematic to write or run.

petebankhead:
This is now included in v0.2.0-m12.

PR #475, opened 2020-05-13
Scriptable pixel classifiers, bug fixes

petebankhead:
* Many pixel classifier improvements
  * Training annotations are used from all compatible images that are currently open
  * New 'Load training' button to support training using annotations from multiple images
  * New 'Measure' button to store measurements after the classifier is closed
  * More control over the regions where the classifier is applied in 'live' preview mode
  * Default classifier is now ANN (often better & much faster than Random Trees)
* Measurement lists are reset if an object's ROI is changed
  * This guards against inadvertently producing invalid measurements by modifying an annotate after measuring
* Viewer no longer centered on selected object when the selection changes or when typing 'Enter'
  * Fixes some annoyances, especially when annotating across multiple viewers
  * Center viewer by double-clicking objects in the 'Annotations' or 'Hierarchy' tab, or in a measurement table
* Improved spatial measurements
  * Optionally split multi-part classifications (e.g. "Class 1: Class 2") for distance calculations (https://github.com/qupath/qupath/issues/405)
* 'Classify -> Object classification -> Set cell intensity classification' now works for all detections, if no cells are present
* LabeledImageServer improvements
  * Supports more than 255 distinct labels
  * New useUniqueLabels() option to support labelling by object, not only classification
* Fixed bug/ambiguity in 'Fill holes' & 'Remove fragments and holes'
  * Handle nested polygons/holes more reliably
  * Changed behavior! Area thresholds now refer to total polygon/hole area ignoring any nested polygons or holes
* Other bug fixes, including:
  * Local normalization now applied before calculating other features (was applied after in m11)
  * Fixed bug in 'Simplify shape' to handle polygons and rectangles properly
  * Projects are automatically saved after changing the image name (https://github.com/qupath/qupath/issues/465)


sberkerss:
Hi mr. Bankhead,

I am having difficulty with pixel classification. I trained a pixel classifier that I want to apply to annotations I made in advance only. However, when im loading a pixel classifier after selecting the annotation of interest, the pixel classifier continues to classify the entire image, rather than the selected annotation. How does it work & what am I doing wrong here?

melvingelbard:
Hi sberkerss,

When you load your pixel classifier, you can choose where to apply it. In this case I believe you want to select 'Any annotations' as Region. Also, it is important to note that the classification preview often goes over the annotations edges, as it uses tile information rather than strict annotation edges. However, the final objects created from the pixel classifier will be restricted to the area of the annotations (if selected in the combo box).

Also, this is a pull request, in the future it would be preferable if you could use the [forum](https://forum.image.sc/tag/qupath) to ask questions. We'd prefer to keep the Github PR page (& issues) as clean as possible for organisation's sake. Plus you might find previous answers from the QuPath community there that migh be useful to you!

sberkerss:
Thank you very much!
Sorry, I'm new here: I will make sure to use the forum next time!

Op wo 8 jul. 2020 10:53 schreef melvingelbard <notifications@github.com>:

PR #491, opened 2020-05-19
Fix typos

camlloyd:
This is non-exhaustive, but I hope it is at least correct. Squash as necessary.


camlloyd:
However, methods such as `suportsImageType` remain unchanged.

petebankhead:
Thanks, do you have a good automated way to find these?

camlloyd:
@petebankhead I set the scope of IntelliJ's spellchecking to run across the whole project. There are options to process code and literals, but ignore comments. Although, it may be worth adding all the instances of say `qupath` to a custom project-level dictionary.

So to _find_ these, I think it works. I did not think it was a good idea to automate _fixing_ everything. Hence there is some human intervention required.

petebankhead:
Good to know, thanks. I'm developing in eclipse, will have a look to see if it has something similar.

camlloyd:
Not sure about built-in, but there are Hunspell plugins for both!

PR #496, opened 2020-05-21
Fix more typos

camlloyd:
Note I have not addressed the following (here be dragons):

- [ ] `actionInterativeAlignment`
- [ ] `bfUseParallization`
- [ ] `CELL_NULCEI_BOTH`
- [ ] `getAncenstorList`
- [ ] `indexerHierearchy`
- [ ] `initializeMatadata`
- [ ] `matricies`
- [ ] `regionstrationType`

Other inconsistencies:
- [ ] Pyramidize/pyramidalize
- [ ] Descendent/descendant

### DIFF
--- a/qupath-core/src/main/java/qupath/lib/roi/GeometryTools.java
+++ b/qupath-core/src/main/java/qupath/lib/roi/GeometryTools.java
@@ -803,7 +803,7 @@ public class GeometryTools {
 				areaTempSigned += 0.5 * (x0 * y1 - x1 * y0);
 				// Add polygon if it has just been closed
 				if (closed && points.size() == 1) {
-					logger.debug("Error when converting area to Geometry: cannot create polygon from cordinate array of length 1!");
+					logger.debug("Error when converting area to Geometry: cannot create polygon from coordinate array of length 1!");
 				} else if (closed) {
 					points.closeRing();
 					if (points.size() <= 3) {
--- a/qupath-experimental/src/main/java/qupath/lib/gui/align/ImageAlignmentPane.java
+++ b/qupath-experimental/src/main/java/qupath/lib/gui/align/ImageAlignmentPane.java
@@ -802,7 +802,7 @@ public class ImageAlignmentPane {
 				motion = opencv_video.MOTION_EUCLIDEAN;
 				break;
 			default:
-				logger.warn("Unknown registraton type {} - will use {}", regionstrationType, RegistrationType.AFFINE);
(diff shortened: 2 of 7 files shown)


petebankhead:
Thanks, I'll look through the ones in the code when I get time, and fix any that are unlikely to anger the dragons too much.

petebankhead:
Most looked pretty innocuous. Only the persistent preference for Bio-Formats has suffered, but I think v0.2.0 may require a new preference key anyway for a clean separation.

Just pyramidize/pyramidalize remains... got a preference?

camlloyd:
I would have to think. I am no expert, but my current understanding is that:

- _to pyramidize_ is to make **x** into a pyramid.
- to _pyramidalize_ is to make **x** pyramidal (pyramid-_shaped_). **x** is not itself a pyramid.

Given the context, there may not be much in it... "Semantics".
Personally, I would go with *pyramidalize*. I could of course be completely mistaken. 

If it's anything to go by, this option also appears more often in the source code.

Or, you could leave it alone for now. It is your call. I wanted to get these polished before v0.2.0, but I know you and Melvin have higher priority tasks at hand*.

*Which is why I wanted to submit these PRs to alleviate you... but they still require your review 🤦

petebankhead:
Sounds good, I feel a slight preference for *pyramidalize* myself... as I guess is reflected in the frequency.

Thanks for submitting these, review was quick :)

v0.2.0 has been a rather intense experience, since almost every part of the software had to be rewritten (after a year away from it), while simultaneously trying to support more and more users navigate the many interrelated and ongoing changes... but I think it was necessary to have a more solid basis for future work. Hopefully, future developments can be a bit less stressful and disruptive for all concerned. It will help not to have to support quite so many misspelled method names :)

camlloyd:
Thank you both for all your efforts. 
It's been a pleasure to follow along and I look forward to seeing what happens next!

camlloyd:
@petebankhead I added another to the list 😅

petebankhead:
Well, it's increasingly clear how much my normal functioning depends upon a spellchecker....

Fixed them now, but unfortunately one instance of `pyramidize` remains - it will turn up on .qpproj files, so changing it without a lot of extra caution would cause compatibility troubles.

camlloyd:
@petebankhead unfortunately `showStarupMesssage` also has one too many `s`

petebankhead:
Fixed! https://github.com/qupath/qupath/pull/498/commits/b797c2cb1d6a974e5f50d01c7bee425a3ca2620e

PR #504, opened 2020-05-27
Added threshold bars to intensity graph

melvingelbard:
Added threshold bars to intensity graph in 'Set cell intensity classifications' (CellIntensityClassificationCommand)


petebankhead:
Looks good! Could this also be added for `SingleMeasurementClassificationCommand`...? I think that's the last one that needs it.

Issue #557, opened 2020-07-16
Hide names for hidden classes


petebankhead:
## Bug report

**Describe the bug**
Objects with specific classes can be hidden in the viewer, but their names linger regardless of the *Show names* setting.

**Expected behavior**
Names are hidden if objects are hidden.

**Additional context**
See https://forum.image.sc/t/cell-density-around-vessel/40391/11

PR #558, opened 2020-07-16
Fixed visibility of hidden classes name

melvingelbard:
- Hidden classes now don't have their name visible in the viewer (#557 )
- Description pane (see `ProjectBrowser` class) doesn't show anymore when setting an entry description with an empty String


petebankhead:
Looks good, but is the renaming of ViewTrackerAnalysisCommand.java intentional?

melvingelbard:
Mmh yes and no.
Git must have added the renaming of my local `ViewTrackerAnalysisCommand.java` to the commit somehow, which will probably be part of a future commit.. I'll close this PR and re-open it without the file renaming (to avoid weird behaviour).

Issue #592, opened 2020-08-14
Confusing behavior when adding a .qpdata file to a project


petebankhead:
## Bug report

**Describe the bug**
Adding an existing .qpdata file to project is a difficult, confusing and buggy process. Basically, it doesn't really work because by default QuPath saves to the path stored in the `ImageData` rather than to the path required for saving the data in the project.

See https://forum.image.sc/t/qpdata-file-in-a-project/41520

**To Reproduce**
1. Create a project
2. Add an *existing* .qpdata file from some other source. The data and objects should load as normal.
3. Open another image
4. Return to the original image. The data and objects are lost.

**Expected behavior**
No surprises. Either the data is retained, or the operation is not permitted.
Ideally, the data would be copied to the project.

**Desktop (please complete the following information):**
 - QuPath v0.2.2 and earlier

**Additional context**
I've written a gist to help resolve this for now: https://gist.github.com/petebankhead/c9f6aed95b38996b2380b09134f2251b

PR #597, opened 2020-09-07
Fix adding .qpdata to an existing project

petebankhead:
See https://github.com/qupath/qupath/issues/592
Solution is to write the ImageData immediately to the project, setting the ID property. This means that calls to 'Save' will now use the 'project' path, rather than the path previously stored within the ImageData.


melvingelbard (approved):
This fixes the problem for me as well.
On another note, it seems that adding a .qpdata from an image hosted on OMERO to an existing project throws an NPE. Not directly related to this fix, but worth investigating.

petebankhead:
Ah, good catch! It looks like the underlying problem is slightly different - I've created a new issue for it at https://github.com/qupath/qupath/issues/598

PR #619, opened 2020-10-21
add rotation knob and remove straighten TMA

mahdilamb:
* Removes the option to straighten by TMA (currently just commented out)
* adds a rotation knob GUI element
* replaces rotation slider in View -> Rotate to rotation knob


mahdilamb:
made in error

PR #623, opened 2020-10-21
Add Rotation slider

mahdilamb:
-add rotation slider
-remove Straighten TMA


petebankhead:
I'll close this for now, since we've switched to another branch and there's a bit more rotation slider stuff to come.

PR #633, opened 2020-10-30
Fixed class saving when exiting QuPath

### LINKED ISSUE #626: Extremely long class list prevents exiting QuPath
## Bug report

**Describe the bug**
QuPath has major problems if the list of available classifications is too long.

See https://forum.image.sc/t/extremely-long-class-list-prevents-exiting-qupath/44381 for details.

**Expected behavior**
QuPath closes elegantly, and retrieves the class list when restarted.

**Additional context**
Presumably we hit the [8192 character limit of what can be stored in preferences](https://docs.oracle.com/javase/8/docs/api/java/util/prefs/Preferences.html#MAX_VALUE_LENGTH).

Note that this also impacts byte arrays (since they use Base64 encoding).

Options include:
* Use the user directory instead (but this might not be set)
* Encode the preferences in a different way, e.g. across multiple entries if needed

melvingelbard:
Fixed #626.


melvingelbard:
Yes good point, the message should probably be more accurate! I'll change this.

PR #636, opened 2020-11-06
Added info to error message for left quotes

melvingelbard:
Left quotes have been known to lead to unexpected behaviours when running scripts.
Now, if a script throws an Exception, QuPath checks whether there is at least one left (single/double) quote and notify the user in the error message.

### DIFF
--- a/qupath-gui-fx/src/main/java/qupath/lib/gui/scripting/DefaultScriptEditor.java
+++ b/qupath-gui-fx/src/main/java/qupath/lib/gui/scripting/DefaultScriptEditor.java
@@ -986,6 +986,14 @@ public class DefaultScriptEditor implements ScriptEditor {
 							sb.append("\n    import " + suggestedClass.getName() + "\nat the start of the script. Full error message below.\n");
 						}
 					}
+					
+					// Check if the error was to do with a special left quote character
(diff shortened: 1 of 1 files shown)


petebankhead:
I'm not exactly sure why, but I get a weird character showing in the error message if I run
```groovy
a = “something”
```
I see
```
At least one left quotation mark (“) was found @ line 1 column 5! You can try replacing it with a regular apostrophe (').

ERROR: MultipleCompilationErrorsException at line 1: startup failed:
Script14.groovy: 2: Unexpected input: '“' @ line 2, column 5.
   a = “something”
       ^

1 error
```
(On a Mac, it's shown as a square in the script editor).

I'd suggest:
* matching the suggestion to the suggestion to the error, i.e. whether it is a single or double quotation mark; otherwise, correcting `“` to `'` would still likely be an error
  * alternatively, give a generic error message `"You can try replacing all instances of fancy quotation marks (‘“’”) with a straight ones (',")"`
* referring to a 'straight quotation mark' rather than apostrophe (which might be curly)
* referring to `At least one invalid quotation mark` rather than `At least one left quotation mark`

melvingelbard:
The weird character appeared to be a typo!
Should now be fixed, along with the suggestions you made.

petebankhead:
Looks good!

Issue #615, opened 2020-10-16
Support for OpenJDK 15


petebankhead:
## Bug report

**Describe the bug**
OpenSlide doesn't work on Windows when QuPath is build with OpenJDK 15.
This appears to come from a change in jpackage - see https://forum.image.sc/t/mrxs-format-not-read-after-building-qupath-0-2-3-with-tensorflow/44122/6

**Expected behavior**
OpenSlide works as before.

**Desktop (please complete the following information):**
 - OS: Windows
 - QuPath Version: Future builds

juliangilbey:
Here's another OpenJDK 15 issue; after building with OpenJDK 15, I get the following error message (on a Linux system):

```
euler:~/qupath (master) $ build/dist/QuPath-0.2.3/bin/QuPath-0.2.3 
Error opening "/home/jdg/qupath/build/dist/QuPath-0.2.3/lib/app/QuPath-0.2.cfg" file: No such file or directory
```

This does not occur when the package is built with OpenJDK 14, so somewhere the path is getting messed up. 

Best wishes,

   Julian

petebankhead:
Ah thanks, good to know. Sounds like it is related to things now being rearranged, but I had been optimistically hoping it would be Windows-only.

petebankhead:
Just for the full set, it fails on macOS too... albeit for a different reason (version numbers that start with 0 seem not to be supported).

anotherBunchOfAtoms:
@juliangilbey I managed to solve this issue by duplicating QuPath-0.2.3.cfg and renaming it to QuPath-0.2.cfg
Best wishes,
Paul

anotherBunchOfAtoms:
however I do get the following error when starting qupath:

`[JavaFX Application Thread] [ERROR] q.l.i.s.o.OpenslideServerBuilder - Could not load OpenSlide native libraries
java.lang.UnsatisfiedLinkError: no openslide-jni in java.library.path: /usr/java/packages/lib:/usr/lib64:/lib64:/lib:/usr/lib
        at java.base/java.lang.ClassLoader.loadLibrary(Unknown Source)
        at java.base/java.lang.Runtime.loadLibrary0(Unknown Source)
        at java.base/java.lang.System.loadLibrary(Unknown Source)
        at org.openslide.OpenSlideJNI.<clinit>(OpenSlideJNI.java:55)
        at org.openslide.OpenSlide.<clinit>(OpenSlide.java:53)
[...]

PR #640, opened 2020-11-13
Fix setting maximum memory via "Help>Show setup options"

ajay1685:
Setting up maximum memory via "QuPath>Help>Show setup options" fails on Windows. Adding the "java-options="  in from of the memory (i.e. -Xmx10240M) fixes the problem. This will match with java option pair values generated by the following line in the "build.gradle" file:
fileTemp << System.lineSeparator() << 'java-options=' << '-XX:MaxRAMPercentage=50' <<

### DIFF
--- a/qupath-gui-fx/src/main/java/qupath/lib/gui/prefs/PathPrefs.java
+++ b/qupath-gui-fx/src/main/java/qupath/lib/gui/prefs/PathPrefs.java
@@ -273,7 +273,7 @@ public class PathPrefs {
 						logger.warn("Cannot set memory to {}, must be >= 512 MB", n);
 						n = 512;
 					}
-					String memory = "-Xmx" + n.intValue() + "M";
+					String memory = "java-options=" + "-Xmx" + n.intValue() + "M";
 					Path config = getConfigPath();
 					if (!Files.exists(config)) {
 						logger.error("Cannot find config file!");


ajay1685:
I just realized that if this was a windows only change can we do this:
```
String memory = "-Xmx" + n.intValue() + "M";
if (GeneralTools.isWindows())
	memory = "java-options=" + memory;
```

petebankhead:
Thanks, can you let me know
* Does it fail already with QuPath v0.2.3?
* Which JDK version are you using?
* Have you checked the impact of the change on Mac / Linux?

petebankhead:
I'm fairly sure this worked at some point, so wondering if it's part of the general issue of [`jpackage` apparently changing in OpenJDK 15](https://github.com/qupath/qupath/issues/615), which will require (different) fixes across all platforms... made more difficult by the fact I haven't been able to find a list of what specifically changed.

ajay1685:
There is no issue with QuPath-0.2.3 from the GitHub releases. When I build from dev-0.3.0 branch using jdk-15.0.0.36-hotspot, it does not correctly set the memory from the help menu.

The config file for 0.2.3 looks like this:
Before setting memory
```
[JavaOptions]
-XX:MaxRAMPercentage=50
```
After setting memory to 50gb, and it works fine.
```
[JavaOptions]
-Xmx51200M
```

With the dev-0.3.0 and jdk-15.0.0.36-hotspot
Before setting memory:
```
[JavaOptions]
java-options=-XX:MaxRAMPercentage=50
```
After setting memory to 45gb
```
[JavaOptions]
-Xmx46080M
```
"java-options=" is missing after setting the memory and QuPath fails to set the correct memory (and defaults to 16gb on my system)

petebankhead:
Ah ok, thanks for investigating - seems to be another of the `jpackage` changes.

Fortunately, the packaging tool is [targeted for Java 16](http://openjdk.java.net/projects/jdk/16/) so should hopefully be stable then.

I'm hesitant about merging this since it really requires a thorough investigation of everything that has changed with Java 15, tested across all platforms, and making a call about whether compatibility with Java 14 needs to be maintained (probably not - but in that case, more changes should be incorporated in the build.gradle to warn the user at build time).

https://github.com/qupath/qupath/issues/615 is the issue intended to track this, but we aren't planning to explore this until our dev branch is a bit more stable and closer to a release. In the meantime, I updated the docs recently to [list Java 14 as required](https://qupath.readthedocs.io/en/latest/docs/reference/building.html#install-a-java-development-kit-jdk), for consistency across branches.

ajay1685:
I will close the pull request, and follow changes at #615

petebankhead:
Thanks - and thanks bringing this up too, I haven't made builds with Java 15 yet so good to know this is another thing to look out for.

PR #642, opened 2020-11-20
Upload object/pixel classifiers

melvingelbard:
- Added tooltip in Pixel classifier UI to notify users that classifiers can't be saved if not working inside a project (tooltip only appears when no project is detected).
- Added ability to Drag & Drop classifier(s) onto the Load object classifier pane to add a classifier from a different source (e.g. a different project).
- Added ability to add classifier(s) to the `comboBox` in Load pixel classifier pane to add a classifier from a different source. (e.g. a different project).

Note (and possibly `TODO`): when adding a `PixelClassifier`, no check is made to ensure that the classifier is valid.

### DIFF
--- a/qupath-extension-processing/src/main/java/qupath/process/gui/commands/ObjectClassifierLoadCommand.java
+++ b/qupath-extension-processing/src/main/java/qupath/process/gui/commands/ObjectClassifierLoadCommand.java
@@ -22,6 +22,7 @@
 package qupath.process.gui.commands;
 
 import java.awt.image.BufferedImage;
+import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.util.ArrayList;
@@ -56,6 +57,7 @@ import qupath.lib.classifiers.object.ObjectClassifiers;
 import qupath.lib.common.GeneralTools;
 import qupath.lib.gui.QuPathGUI;
 import qupath.lib.gui.dialogs.Dialogs;
+import qupath.lib.gui.dialogs.Dialogs.DialogButton;
 import qupath.lib.gui.tools.PaneTools;
 import qupath.lib.images.ImageData;
 import qupath.lib.io.GsonTools;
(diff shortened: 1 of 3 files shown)


petebankhead:
Thanks, I notice the tooltip text only appears if the mouse if over the label - not actually the unavailable text field of button. I only spotted it appears at all because I checked the code; I think it should appear when over any of the components on that row.

petebankhead:
A few more things:
* If I try to load a pixel classifier, I'm still prevented from opening the dialog if there is no classifier in the current project. In this case, I'd expect the dialog to open and allow me to choose a classifier.
* There's a gap to the right of the options now in the dialog that looks like it shouldn't be there.
Screenshot 2021-03-05 at 13 45 30
* It would be a little nicer to determine whether `(s)` is needed or not and adjust the message displayed to the user accordingly when offering to copy the classifier. I also think the wording is a bit redundant; I'd prefer 'Copy classifier to the current project?' and *Yes*, *No*, *Cancel*. Depending upon how the dialog is constructed you may have space to add some further text to explain what it means (but I'm not sure that's really needed).
Screenshot 2021-03-05 at 13 45 38
* I couldn't see a way to work drag & drop with the pixel classifier. I *can* drag a classifier onto QuPath's main window, but it appears that QuPath then tries (and fails) to open the classifier as an image. It doesn't display any meaningful error - or at least not for long enough for me to read it.
* There isn't any text in the *Load object classifiers* dialog that indicates you can load a classifier by drag & drop. I think this text should be added; you might also support adding a classifier by double-clicking and providing a file chooser.

It doesn't have to part of this pull request, but since this will offer support for loading classifiers from elsewhere, perhaps we should relax the strict requirement that classifiers can only be saved within a project. What do you think?

melvingelbard:
I've corrected the little bugs/typos that you mentioned (and an extra one found last minute!).

Regarding easing the restrictions about saving classifiers outside a project: it could be done in the future, even though my gut feeling is to leave this more controlled environment as it is now. The above code handles import (and copy) of classifiers that are outside a project directory, but I think we can only rely on the paths that are within a project. If allowing classifiers from any location, does this mean we will need to start 'resolving paths' if paths change?

petebankhead:
Makes sense - ok, let's leave it as it is and not add any extra complexity for now. Thanks!

PR #645, opened 2020-11-24
Change regular expression matching to unicode escape characters (windows encoding issue)

mahdilamb:
This PR resolves an issue with formatting with Windows encoding


mahdilamb:
This PR resolves an issue with formatting with Windows encoding

PR #647, opened 2020-12-05
Command line option for script arguments

zindy:
This pull request lets a user pass not only a Groovy script name via the qupath-console command line arguments, but also the script's arguments via a `-a` or `--args` option followed by quoted string.

This string is then split on whitespace into a list of strings (unless a block is surrounded by \\" \\" escaped quotes) and passed to the Groovy interpreter as the global variable args.

When no arguments are passed to a script, the global variable args is an empty list. With this change, the integrated script editor / interpreter also gets an empty, global args variable.

I explained in https://forum.image.sc/t/using-command-line-parameters-with-a-groovy-script-in-qupath/45949/4 why this is desirable.

Feel free to merge this pull request, or to adapt the idea, *any way you wish*.

### DIFF
--- a/qupath-gui-fx/src/main/java/qupath/lib/gui/scripting/DefaultScriptEditor.java
+++ b/qupath-gui-fx/src/main/java/qupath/lib/gui/scripting/DefaultScriptEditor.java
@@ -503,17 +503,17 @@ public class DefaultScriptEditor implements ScriptEditor {
 		
 		ScriptEditorTextArea control = new ScriptEditorTextArea(editor);
 		editor.addEventFilter(KeyEvent.KEY_PRESSED, (KeyEvent e) -> {
-			if (e.getCode() == KeyCode.TAB) {
-				handleTabPress(control, e.isShiftDown());
-				e.consume();
-			} else if (e.isShortcutDown() && e.getCode() == KeyCode.SLASH) {
-				handleLineComment(control);
-				e.consume();
-			} else if (e.getCode() == KeyCode.ENTER && control.getSelectedText().length() == 0) {
+	        if (e.getCode() == KeyCode.TAB) {
+	        	handleTabPress(control, e.isShiftDown());
+	        	e.consume();
+	        } else if (e.isShortcutDown() && e.getCode() == KeyCode.SLASH) {
+	        	handleLineComment(control);
+	        	e.consume();
+	        } else if (e.getCode() == KeyCode.ENTER && control.getSelectedText().length() == 0) {
 				handleNewLine(control);
 				e.consume();
 			} 
-		});
+	    });
 
 //		editor.getDocument().addUndoableEditListener(new UndoManager());
 //		// Handle tabs
@@ -646,13 +646,13 @@ public class DefaultScriptEditor implements ScriptEditor {
 		panelList.setCenter(titledScripts);
 		listScripts.getSelectionModel().selectedItemProperty().addListener((v, o, n) -> updateSelectedScript());
 		listScripts.setCellFactory(new Callback<ListView<ScriptTab>, 
-				ListCell<ScriptTab>>() {
-					@Override 
-					public ListCell<ScriptTab> call(ListView<ScriptTab> list) {
-						return new ScriptTabListCell();
-					}
-				}
-			);
+	            ListCell<ScriptTab>>() {
+	                @Override 
+	                public ListCell<ScriptTab> call(ListView<ScriptTab> list) {
+	                    return new ScriptTabListCell();
+	                }
+	            }
+	        );
 		listScripts.setMinWidth(150);
 		runningTask.addListener((v, o, n) -> listScripts.refresh());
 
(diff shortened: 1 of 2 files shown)


zindy:
Sorry I fumbled my commits, it should all be fixed now! There are some indentations in DefaultScriptEditor.java that I inadvertently corrected from spaces to tabs, when all I wanted to do was to make sure my own indentation follow the qupath standard.

I'll just leave this as a note, someone can submit a separate pull request to fully correct the file DefaultScriptEditor.java (if that's important).

Cheers,
Egor

petebankhead:
Thanks @zindy! Just a quick update to say I've been discussing how best to support scripting arguments with @melvingelbard, since he has done most of the command line work. I think a combination of the method you use (creating a global `args` variable) and using picocli for parsing would work well.

Meanwhile, we have a few other related-ish scripting issues to deal with, including
*  supporting the ability to run scripts from other scripts.
* revising the Groovy integration to use an approach other than JSR223: https://docs.groovy-lang.org/latest/html/documentation/guide-integrating.html

The benefit of JSR223 is it potentially enables QuPath to support multiple scripting languages (e.g. Jython), but this comes at the expense of supporting every one of them a bit less than it otherwise could. I'm tempted to commit to Groovy-only and make the support better, rather than attempting to support Jython & more - I'd be very interested to know your thoughts on that.

Anyhow, this is the reason why this pull request will take a bit more thought - I think it would be good to add args support at the same time as we make the other changes, hopefully in the next release.

zindy:
@petebankhead -- Thanks for the feedback!

> [...] and using picocli for parsing would work well.

When I searched how to use picocli to split my argument string into chunks, I could not find an answer.

In the meantime, [this exchange](https://github.com/remkop/picocli/issues/1277) came-up in the picocli issues, and here's a potential solution:

> There is a list-like facility for unmatched parameters: see Unmatched Input.
> 
> The easiest way to use this is with the @Unmatched annotation:

```
class MyApp {
    @Unmatched
    List<String> unmatched;
}
```

I'll have to investigate this, but hopefully this means I can remove the `translateCommandline()` code I pasted into qupath.java and just use the libraries already imported into qupath.java.

On the Groovy side though, I'm still trying to figure out how to parse the `this.args` string list. One way I tried is to use [args4j](https://github.com/kohsuke/args4j) as per [this gist](https://gist.github.com/seanf/6362260), but it means using an external jar library, which I'm not too keen on.

I've been wondering if there is a way to do this with picocli since that can be imported into QuPath's Groovy and found [this page](https://picocli.info/picocli-2.0-groovy-scripts-on-steroids.html) which looks useful, but I haven't explored this much further.

> I'm tempted to commit to Groovy-only and make the support better, rather than attempting to support Jython & more - I'd be very interested to know your thoughts on that.

Me personally, I've always been happy with one / the most used scripting language in a piece of software. I love Python, but my requirements for small scripts never made me want to say, try to use Python to script ImageJ. Having one default scripting language also means there's only one way to do something, and makes it easy to copy-paste code from another script without having to worry about which scripting language the script was written in (I shamefully admit, I do that a lot with your Groovy scripts!).

I'll report back when I get further with picocli.

Cheers,
Egor

melvingelbard:
Hi all, 

I have pushed a solution to this on my fork this morning (link [here)](https://github.com/melvingelbard/qupath-forked/tree/scriptArgs). It uses Picocli to parse the unmatched args (as @zindy mentioned I believe), which are then accessible through the `args` variable in the current running script.

`args` is a simple `String[]`, so no fancy further parsing is performed on it. Not sure whether that was the original plan..
What do you think of this implementation?

zindy:
Hi @melvingelbard,

I see what you did and it makes the additions short and sweet. What you're losing though is the error checking and the ability to separate the arguments for the script from those for QuPath.

That might make tracking some syntax errors (e.g. `--poject` vs `--project`) rather interesting without error checking.

Limiting conflicts between QuPath options and script options could just be something to watch out for, I don't have an immediate feel for how much of an issue that would be (e.g. use `--newproject` rather than `--project` if the goal is to create a new project rather than opening the script on a specific QuPath project).

So, pros and cons: My method ensures isolated arguments between QuPath and the script, but at the expense of having to add the `translateCommandline()` function, and a heavier syntax with quotes around the script commands and escaped quotes for paths with spaces. Yours is a lot simpler, but I think the biggest drawback is losing the error checking and possible collisions between options for QuPath and options for the script.

That's just top of my head, I'm happy to have any kind of command-line control over the script, so I'll let you decide how you want to take this forward.

petebankhead:
I've been thinking about this, and made an attempt to get the best of both worlds: https://github.com/qupath/qupath/pull/676

I remain reluctant to do too much parsing, since as long as the string is passed to Groovy it's quite possible to parse it within the script itself. But I see the risk in allowing misspelled arguments. I hope this approach is a suitable compromise, but I don't do a huge amount of command line stuff so not sure if I've broken some convention along the way.

Curious as to what you think...

petebankhead:
@zindy Do you mind if I close this and merge https://github.com/qupath/qupath/pull/676 in preparation for v0.3?
That would give time for anyone trying the dev branch to test it a bit before release.

PR #651, opened 2020-12-08
Tileexporter fix

### LINKED ISSUE #648: TileExporter requestedPixelSize method gives incorrect results
In the file `qupath-core/src/main/java/qupath/lib/images/writers/TileExporter.java` the definition of the `requestedPixelSize` method has the fraction the wrong way up; it should instead read:

```java
	public TileExporter requestedPixelSize(double pixelSize) {
		this.downsample = pixelSize / server.getPixelCalibration().getAveragedPixelSize().doubleValue();
		return this;
	}
```

Otherwise, if the requested pixel size is 5.0, say, and the average pixel size is 0.25, the downsample value is set to 1/20 instead of the correct 20.

This affects the current GitHub version of QuPath (at commit 3ca92f9026c9c32bcfb28de419673d2beac49c4a).

### LINKED ISSUE #649: A possible (cosmetic?) bug in TileExporter in splitRegionRequests
In `splitRegionRequests`, the working coordinates are in downscaled pixels, and stored in `maxX` and `maxY`.  However, the two nested `for` loops over `x` and `y` have incorrect initialisers: for `y` it is `int y = request.getY()` instead of `int y = (int)(request.getMinY() / downsample)`.  It probably doesn't matter much, though, as presumably `request.getY()` and `request.getMinY()` both return 0 in normal situations.

### LINKED ISSUE #650: In TileExporter, does not correctly loop over t and z values
In the function `getTiledRegionRequests`, there's a nested loop over t and z, but nowhere in the body of the loop are t and z used.  So it would seem that all that would result in the case of an image with nTimepoints > 1 or nZSlices > 1 is multiple identical sets of tiles.  More than that, they would probably just overwrite each other, as in splitRegionRequests, it sets z = request.getZ(), which will presumably always return 0, and likewise for t.  Perhaps the values of z and t should be passed as parameters to splitRegionRequests instead?  That would probably make this work as intended.

juliangilbey:
This set of three patches fixes #648, fixes #649 and fixes #650.


petebankhead:
Thanks Julian, we'll add this changes but don't want to accept the pull request as-is because it brings in some other commits that shouldn't be there, caused by some recent git-related confusion and we try to sanitise our branches for v0.3.

It will take us some time to resolve the problems on our side, in the meantime just logging all related things in a GitHub Issue is great.

juliangilbey:
Sure; I saw weird stuff in the GitHub response to this pull request!  At least you can look at my code if you wish!

Best wishes, Julian

juliangilbey:
I've rebased my patches on main, now that I understand that main is the correct branch to use, so I'll make a separate pull request and close this one.

PR #652, opened 2020-12-08
Main+tileexporter fix

### LINKED ISSUE #648: TileExporter requestedPixelSize method gives incorrect results

### LINKED ISSUE #649: A possible (cosmetic?) bug in TileExporter in splitRegionRequests

### LINKED ISSUE #650: In TileExporter, does not correctly loop over t and z values

juliangilbey:
A clean set of patches this time that fixes #648, fixes #649, fixes #650


petebankhead:
There shouldn't be any need to pass `z` and `t` as separate parameters to `splitRegionRequests` - rather `fullRequest` can be updated to include these, which would be good to reduce the number of necessary `int` parameters.

Sorry, I don't really have the bandwidth to look at this in detail or handle pull requests in the near future, so this is likely to linger for a while.

juliangilbey:
No rush on my part - just trying to be helpful! :-)

Best wishes, Julian

juliangilbey:
On the thought of not passing the parameters: you could indeed do this by saying:
```java
fullRequest = fullRequest.updateT(t)
fullRequest = fullRequest.updateZ(z)
```
in the loop in `getTiledRegionRequests` and then reinstating these two lines in `splitRegionRequests`:
```java
t = request.getT()
z = request.getZ()
```
So it's a balance between four extra lines of code on the one hand and two extra parameters on the other.

petebankhead:
Finally returning to this - I favour the extra lines of code over the parameters. There are already 4 int parameters, and increasing this to 6 makes it harder to work with I think. The `RegionRequest` *should* be used to incorporate `z` and `t`, and it's really a bug if they are ignored.

I'd also rather add this to the `dev-0.3` branch rather than `main`.

Thanks for looking into all this - If you want to make the changes I can merge them, otherwise one of us can repeat the main bits in a separate pull request later this week.

juliangilbey:
Hi Pete!  I've updated my pull request by basing it on `dev-0.3` and changing the parameter passing as requested.

juliangilbey:
Unfortunately, though, I don't know how to tell GitHub that this patch should be against dev-0.3 rather than against main.  So I'll close this PR and submit a new one.

PR #653, opened 2020-12-09
Drag and Drop for URL

melvingelbard:
Implemented the Drag and Drop feature on the viewer for URLs (see https://forum.image.sc/t/feature-proposal-add-url-drag-and-drop-handler/40884).

Note: URL `DropHandler`s take a `List<String>` parameter. This could potentially be changed to take `List<URL>` parameter instead if needed (though the probability that an implementation of a 'String Drag and Drop' is needed in the future is *quite* low).


petebankhead:
Looks good!

PR #665, opened 2021-02-10
Intel OpenVINO integration

dkurt:
Used for deep learning optimization (StarDist algorithm)
1. Download model in OpenVINO Intermediate Representation format:
[dsb2018_heavy_augment.zip](https://mega.nz/file/IY92FbjK#ppzuYyjdxp9lPLZ2J6BnN6bCUFeZkRgbbolT4cSMI6w)
[he_heavy_augment.zip](https://mega.nz/file/1U1wmTYR#HFLXGNjoAS3WR3E64-g7K3QT58cBxbrAJci2iD0EMkE)
INT8: [he_heavy_augment_int8.zip](https://mega.nz/file/NMsgzJgZ#JseSxOTcOeqQbG2X9p3W2e-CXgJ6KjC8n1cs4QS49mY)
INT8: [dsb2018_heavy_augment_int8.zip](https://mega.nz/file/oAkFDKBA#ZjVP4f2Il_tcjsc0iywpVRXLLvh8lBkkAD3wJ7rDdmQ)

Performance:

| Intel(R) Core(TM) i7-6700K | **Test image**: OS-3.ndpi<br>**Model**: he_heavy_augment<br>**Tile size**: 1024x1024<br>**Number of tiles**: 2183 | **Test image**: LuCa-7color_Scan1.qptiff<br>**Model**: dsb2018_heavy_augment<br>**Tile size**: 1024x1024<br>**Number of tiles**: 936 |
|---|---|---|
|  TensorFlow backend  | 23:05 minutes | 14:20 minutes
| OpenVINO backend | 15:28 minutes (x1.5) | 10:39 minutes (x1.36)
| OpenVINO backend (INT8) | 12:29 minutes (x1.87) | 8:43 (x1.68)
| OpenVINO backend (INT8) + NormalizePercentileOp | 11:55 minutes (x1.99) | 8:27 minutes (x1.7)

Build with OpenVINO:
```
./gradlew clean build createPackage -Puse_openvino=true
```

Use `build` argument to enable OpenVINO backend:
```groovy
def stardist = StarDist2D.builder(pathModel)
      .threshold(0.5)              // Prediction threshold
      .normalizePercentiles(1, 99) // Percentile normalization
      .pixelSize(0.5)              // Resolution for detection
      .build("openvino")           // or "tensorflow" to use TensorFlow backend. If empty, use any available backend
```

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,4 +20,4 @@ jobs:
         java-version: '14.0.2'
 
     - name: Build
-      run: ./gradlew clean build createPackage -Ptensorflow-cpu=true -Pskip_tests
+      run: ./gradlew clean build createPackage -Ptensorflow-cpu=true -Puse_openvino=true -Pskip_tests
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,23 @@
+name: CI
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+
+jobs:
+  run_linux:
+    runs-on: ubuntu-18.04
+    name: Build on Linux
+
+    steps:
+    - name: Clone repository
+      uses: actions/checkout@v2
+
+    - name: Install Java
+      uses: actions/setup-java@v1
+      with:
+        java-version: '14.0.2'
+
+    - name: Build
+      run: ./gradlew clean build createPackage -Ptensorflow-cpu=true -Pskip_tests
--- a/build.gradle
+++ b/build.gradle
@@ -236,7 +236,7 @@ allprojects {
     // Optional versions
     tensorflowVersion  = "1.15.2-${javacppVersion}"
     mkldnnVersion      = "0.21.4-${javacppVersion}"
-    openvinoVersion    = "2021.2.4"
+    openvinoVersion    = "2021.2.7"
     
     /*
      * Link to the main external Javadocs if -PlinkJavadoc=true
--- a/build.gradle
+++ b/build.gradle
@@ -273,12 +273,12 @@ allprojects {
     guava
     picocli
     jfreesvg
-	junit
-	logback
-	
-	// Optional
-	tensorflow
-  openvino
+    junit
+    logback
+
+    // Optional
+    tensorflow
+    openvino
   }
 
   dependencies {
@@ -370,7 +370,7 @@ allprojects {
     logback "ch.qos.logback:logback-classic:${logbackVersion}"
     logback "org.slf4j:slf4j-api:${slf4jVersion}"
 
-    openvino "org.intel:openvino:0.0.4"
+    openvino "org.intel:openvino:0.0.6"
 
     tensorflow "org.bytedeco:tensorflow:${tensorflowVersion}"
     if (includeTensorFlowGPU)
--- a/build.gradle
+++ b/build.gradle
@@ -236,6 +236,7 @@ allprojects {
     // Optional versions
     tensorflowVersion  = "1.15.2-${javacppVersion}"
     mkldnnVersion      = "0.21.4-${javacppVersion}"
+    openvinoVersion    = "2021.2.4"
     
     /*
      * Link to the main external Javadocs if -PlinkJavadoc=true
@@ -370,7 +371,7 @@ allprojects {
     logback "ch.qos.logback:logback-classic:${logbackVersion}"
     logback "org.slf4j:slf4j-api:${slf4jVersion}"
 
-    openvino "org.intel:openvino:0.0.6"
+    openvino "org.intel:openvino:${openvinoVersion}-${nativesCPP}"
 
     tensorflow "org.bytedeco:tensorflow:${tensorflowVersion}"
     if (includeTensorFlowGPU)
--- a/build.gradle
+++ b/build.gradle
@@ -278,6 +278,7 @@ allprojects {
 	
 	// Optional
 	tensorflow
+  openvino
   }
 
   dependencies {
@@ -368,7 +369,9 @@ allprojects {
     
     logback "ch.qos.logback:logback-classic:${logbackVersion}"
     logback "org.slf4j:slf4j-api:${slf4jVersion}"
-    
+
+    openvino "org.intel:openvino:0.0.4"
+
     tensorflow "org.bytedeco:tensorflow:${tensorflowVersion}"
     if (includeTensorFlowGPU)
 	    tensorflow "org.bytedeco:tensorflow:${tensorflowVersion}:${nativesCPP}-gpu"
@@ -376,7 +379,7 @@ allprojects {
 	    tensorflow "org.bytedeco:tensorflow:${tensorflowVersion}:${nativesCPP}"
     tensorflow "org.bytedeco:mkl-dnn:${mkldnnVersion}"
     tensorflow "org.bytedeco:mkl-dnn:${mkldnnVersion}:${nativesCPP}"
-    
+
   }
   
   javadoc {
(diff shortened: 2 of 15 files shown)


petebankhead:
Thanks! Adding OpenVINO support to QuPath sounds great.

It may take some time to review this properly, as I'm currently in the middle of various interrelated changes for the next release that may impact this - including a switch to the new 'official' TensorFlow Java distribution, moving StarDist into the main project, and defining suitable interfaces to support models of different kinds (including for classification as well as for segmentation).

To get StarDist in the main project, I intend to use OpenCV's DNN module - see [a preview here](https://github.com/petebankhead/qupath/blob/create-objects/qupath-extension-tensorflow/src/main/java/qupath/tensorflow/stardist/StarDist2D.java#L446), which should work if the `.pb` file is converted to something OpenCV-friendly.

OpenCV has the advantage of already being a dependency, but ultimately we want to be able to support alternatives through extensions to broaden the range of supported models - and perhaps offer better performance. So TensorFlow will remain as an option for specialists. If OpenVINO will help too, it would be very good to include it.

In the meantime, I'd like to ask a few questions:
* Should the [OpenVINO backend in OpenCV 4.5.1](https://github.com/opencv/opencv/wiki/ChangeLog#version451) offer an equivalent level of support and performance, or are there advantages in using this instead / as well?
* Can/could this work across Windows, Linux & macOS - or is it Linux-only?
* Am I right in thinking the dependencies are all Apache (or at least GPL-compatible)?

Thanks again!

dkurt:
@petebankhead, many thanks for fast feedback! I've marked this PR as draft and going to work further.
As you mentioned, there is also an option to use OpenCV API but OpenVINO API will give better performance (at least once I could fix asynchronous efficiency issue).

Briefly answering your questions,

> Should the OpenVINO backend in OpenCV 4.5.1 offer an equivalent level of support and performance, or are there advantages in using this instead / as well?

Yes, it can help to avoid IR conversion, in example. OpenVINO natively can load only IR or ONNX at this moment. However it'll require a separate package with OpenCV linked against OpenVINO.

> Can/could this work across Windows, Linux & macOS - or is it Linux-only?

Yes, no problem at all. I just published a test package for Linux only but going to expand it in future.

> Am I right in thinking the dependencies are all Apache (or at least GPL-compatible)?

OpenVINO is Apache 2.0: https://github.com/openvinotoolkit/openvino/blob/master/LICENSE

dkurt:
Hi! Just wanted to ask if you had a chance to take a look. Also, we just added MacOS package to existing Linux and Windows.

dkurt:
@petebankhead, sorry for reminder, just wanted to ask if you have some requests for this kind of changes?

petebankhead:
Sorry for the delay, it's a crazy grant-writing time for the next month, and any coding time has been spent trying to get v0.3 ready for release soon - so I'm afraid I haven't had a chance for a proper look. Our main changes are happening on the [`dev-0.3`](https://github.com/qupath/qupath/tree/dev-0.3) branch, where we have updated to use JDK 16 and use the newer TensorFlow Java jars.

I think this demonstrates that we really need a way to switch between TensorFlow/OpenCV/OpenVINO/something else without changing any of the StarDist code. That should allow the OpenVINO extension to be added separately just by dragging the jars onto QuPath.

I'm not sure how best to do this. If we can come up with a quick-but-maintainable solution we may be able to include it in v0.3, but otherwise I should have a bit more time to think about it in a month.

petebankhead:
Thanks, the problems are all on the QuPath side that we don't have a good enough API to support extensions very well. We definitely need to change this - I'll try to merge my relevant code into `dev-0.3` in the next few days, which should improve things slightly.

QuPath can/will (and to some extent does...) support models for uses beyond StarDist, so we need something generic.

The easiest way I can think to resolve this temporarily is to add a method to `ImageOps` that allows an extension to register itself as able to create an op from a particular model file. But I imagine that is storing up problems for the future, e.g. if OpenCV and a future ONNX Runtime extension both want to handle .onnx files, there needs to be some way to choose.

This pull request is the first to make us think about it - ideas are very welcome!

dkurt:
@petebankhead, created extensions for StarDist algo and TensorFlow/OpenVINO backends (just moved files to separate locations).

petebankhead:
Sorry this has taken so long, but v0.3.0 has finally been released (within the past few minutes....).

It finally has better support for extensions, and so the [QuPath TensorFlow extension](https://github.com/qupath/qupath-extension-tensorflow) can now be in its own repository.

We now have a pretty general way to support deep learning via [`DnnModel`](https://github.com/qupath/qupath/blob/4eb74946602d8df605402909c753fca027fa3c26/qupath-core-processing/src/main/java/qupath/opencv/dnn/DnnModel.java).

I'm sure the design will need to be improved in the future, but the general idea is to try to require as little as possible to implement a new deep learning extension: basically, implement `DnnModel` and (ideally) also register it with `GsonTools`.
[Here](https://github.com/qupath/qupath-extension-tensorflow/blob/main/src/main/java/qupath/ext/tensorflow/TensorFlowExtension.java#L46) is where that happens for TensorFlow.

[StarDist](https://github.com/qupath/qupath-extension-stardist) is now also in its own extension; the builder can take a `DnnModel` as an input as an alternative to a path, so it's relatively easy to add an alternative way or using the model. There's no longer any need to build QuPath from source because [OpenCV can be used instead](https://qupath.readthedocs.io/en/stable/docs/advanced/stardist.html).

The main reason for all this is that it will make it much easier to use deep learning models within QuPath in different ways, e.g. for object classification, or for semantic segmentation as a drop-in replacement for a pixel classifier. In fact, as long as DnnModel is JSON-serializable (ideally just requires a URI) then this already works, but the process needs to be cleaned up and documented. In the meantime, there are a few videos showing U-Net as live pixel classifier [here](https://twitter.com/petebankhead/status/1406856875732258821?s=20).

@dkurt I completely understand if you've lost interest in this after all this time, but if you ever want to update your code to work as an extension in its own repo (I'll link to it) I'd love to give it a try - and I know there are [users who are very interested in benchmarking and trying to maximize StarDist performance](https://forum.image.sc/t/thoughts-on-0-3-2-rc2-with-stardist/56481) :)

dkurt:
@petebankhead, many thanks for such great progress and letting me know! I'll definitely take a look on current structure and update the proposal.

dkurt:
@petebankhead, do you know an approach to run QuPath as a script? In example, StarDist model worked in this way:
```
./QuPath/bin/QuPath script run_stardist.groovy -i OS-3.ndpi
```
where `run_stardist.groovy` is a script from https://qupath.readthedocs.io/en/stable/docs/advanced/stardist.html.

Now with StarDist as a separate JAR extensions, I have an error "unable to resolve class qupath.ext.stardist.StarDist2D". So the question is how to run the script without GUI now, with 0.3.0?

petebankhead:
Hmmm, if that doesn't work then it sounds like a bug on QuPath's side.

In the meantime, as a short term workaround you could try adding the path to the StarDist jar to the QuPath.cfg file. As long as the extension jar is on the classpath it should (I hope) work.

This discussion may be relevant: https://forum.image.sc/t/main-qupath-v0-3-0-jar-file/57220/4

Hacking the .cfg file *shouldn't* be necessary - I'll investigate what's going wrong.

dkurt:
Hi, @petebankhead! I've created a repository with OpenVINO extension. (PR https://github.com/dkurt/qupath-extension-openvino/pull/1). Going to finish with tests and will merge to main branch soon.

Also, it required bounded PR to StarDist ext: https://github.com/qupath/qupath-extension-stardist/pull/12

PR #666, opened 2021-02-12
Path Object I/O (GeoJSON/serialized/compressed)

melvingelbard:
- Export & import (via GUI and/or scripting) `PathObject`s easily (`.qpdata`/`.geojson`/`.zip`).
- Export & import (via GUI and/or scripting) `ROI`s easily (`.qpdata`/`.geojson`/`.zip`).
- Implemented drag & drop for `PathObject`s files (`.qpdata`/`.geojson`/`.zip`).
- Objects I/O is recorded on the command history.
- Implemented object transform command.
- Objects transformation is recorded on the command history.


petebankhead:
Thanks for doing this! I've looked through the proposed changes, thoughts are below:

* For `getAllObjects()`, I suggest replacing
```java
return objList.parallelStream().filter(e -> e.getClass() != PathRootObject.class).toArray(PathObject[]::new);
```
with
```java
return objList.parallelStream().filter(e -> e != hierarchy.getRootObject()).toArray(PathObject[]::new);
```
because *conceivably* the root object might one day be subclassed or someone may have smuggled a `PathRootObject` wrongly into the hierarchy.

(One could also use `e.isRootObject()`, which would handle the case of multiple root objects differently - the main thing is to avoid relying on class equality)

* `selectAllObjects()` appears in both `QP` and `Commands` - can these be harmonized to avoid repetition?

* Specify in the javadoc for `selectAllObjects` that the root object is excluded

* We should avoid `OperationNotSupportedException` (doesn't seem intended for this use)

* I'm not sure `PathObjectTransform` needs a new class; it could be added to `QP`. At least, it is currently confusing that `transformObject(s)` exists both in `PathObjectTransform` and in `PathObjectTools` (even though I understand the rationale in keeping `PathObjectTools` distinct, since the new method works on a hierarchy). Relatedly:
  * `transformObjects` should take a `PathObjectHierarchy` if it doesn't require an `ImageData`
  * I think 2 variations should exist that accept an `AffineTransform` or an `AffineTransformation` - since a list of `Number` is harder for a user (even if helpful for scripting). Although then I wonder... would a builder become a better way to handle it?

* An important use case is when we may want to transform *all* the objects in a hierarchy *and maintain the hierarchical relationships*. I think this requires a separate code path that recursively transforms all child objects, assigning them to the appropriately-transformed new parent object.
  * A similar use case would be wanting to transform all objects recursively, but *without* them already being part of any hierarchy. For example, during import of objects before they are added to a hierarchy that already contains objects (which don't need to be transformed).
    * I think perhaps `PathObjectTools.transformObject` should either take a 'doRecursive' parameter or have a separate method for recursively applying the transform - also checking for a null ROI. This would mean it can be applied to the root object of a hierarchy to transform the entire hierarchy easily and intuitively.
    
* I don't think we should give a serialize export option in the menus. We should move away from Java serialization as much as possible, and discourage anyone from use it; we should still support it for import though. Groovy make serialization quite easy if it's required in a script anyway... except if things are circular. Which brings me to...

* `exportObjectsAsSerialized` looks like it doesn't actually use the transformed list; it also potentially allows circular references via parents/children, which could be confusing and/or buggy. I don't think we need this method if we are discouraging serialization (although I could be wrong... especially if we find a way to use it internally for convenience).

* *If* `importObjectsFromSerialized` is in the public API, I think it should handle things other than lists, e.g. individual objects, arrays of objects, collections. But I'd prefer to expose it in the public API only if its benefits are clear enough.

* It looks like `.qpdata` is being used for serialized object lists. We really shouldn't add a new file type with the same extension. Rather, we *can* read objects from an existing `.qpdata` file using [`PathIO.readHierarchy(File)`](https://github.com/qupath/qupath/blob/43aad4ecda893a7eb03c30774e64da5b9547bc86/qupath-core/src/main/java/qupath/lib/io/PathIO.java#L410) - this should work even if the server is unavailable itself. The ability to import from old `.qpdata` files is important, but I'd like to avoid encouraging anyone to write `.qpdata` files other than those handled internally within projects (to make it easier for us to replace the format in the future).

* A common use case will be transferring objects between images in the same project. Ideally this would be possible without exporting/importing, but rather simply choosing the project entry for import. Internally, this can use [`ProjectImageEntry.readHierarchy()`](https://github.com/qupath/qupath/blob/43aad4ecda893a7eb03c30774e64da5b9547bc86/qupath-core/src/main/java/qupath/lib/projects/ProjectImageEntry.java#L210).
  * Adding this to the UI is tricky, since navigating long lists of project entries is awkward.... could it be supported by dragging an entry from the project list onto a viewer?

melvingelbard:
Thanks for the feedback!
I'll close this PR and reopen a simpler one [here](https://github.com/qupath/qupath/pull/670) to facilitate integration.

PR #668, opened 2021-02-22
Set focus on magnification/downsample textfield

### LINKED ISSUE #667: Dialogs requesting number input still accept characters
**Describe the bug**
Dialogs requesting number input still accept characters.
This can lead to unexpected behaviour from QuPath (and Exceptions thrown).

**To Reproduce**
1. Put any character in a `TextField` asking for a number (e.g. 'Set magnification').

**Expected behavior**
Not being able to even input wrong characters.

 - OS: All
 - QuPath Version 0.2.3

melvingelbard:
- When setting the magnification/downsample for an image, the `TextField` is now directly in focus.
- Fix #667: All the dialogs requesting for a numeric input (such as the one above) are now restricted to digits and decimal separator only (e.g. "1.20" is permitted, while "1.2f0" is not). N.B.: Scientific notation is accepted only on copy/paste, as the filter will reject any intermediate representation of scientific notation (i.e. 1.23E4 **is** fine whereas 1.23E is **not**).

### DIFF
--- a/qupath-extension-processing/src/main/java/qupath/process/gui/commands/SimpleThresholdCommand.java
+++ b/qupath-extension-processing/src/main/java/qupath/process/gui/commands/SimpleThresholdCommand.java
@@ -237,7 +237,7 @@ public class SimpleThresholdCommand implements Runnable {
 				Bindings.createStringBinding(() -> GeneralTools.formatNumber(sigma.get(), 2), sigma)
 				);
 		labelSigma.setMinWidth(25); // Thanks to Melvin, to stop it jumping around
-		GuiTools.restrictSpinnerInputToNumber(sigmaSpinner, true);
+		GuiTools.restrictTextFieldInputToNumber(sigmaSpinner.getEditor(), true);
 		PaneTools.addGridRow(pane, row++, 0, "Select smoothing sigma value (higher values give a smoother result)", label, sigmaSpinner, labelSigma);
 
 		label = new Label("Threshold");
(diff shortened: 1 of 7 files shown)


petebankhead:
Looks good to me. Not 100% sure if prohibiting text input completely is too aggressive (also with its scientific notation implications), but think we should try it and see how it impacts behavior (I never wanted to use scientific notation until I knew I couldn't...).

Small thing: in testing this, I've found an old bug related to the brightness/contrast display. Basically, if I double-click on min/max and enter a new value *that is out of range*, this is not immediately shown in the label - but *does* appear if I then click once on the label after the input dialog has been closed.

Since it fits with the decimal places change, could you look into this please? If it's not a straightforward fix then let me know and I'll merge this as it is.

petebankhead:
I like it! Seems to work well.

Tiny thing: it looks like `Pattern pattern = Pattern.compile("[a-zA-Z&&[^Ee]]+");` is called on every validation of the text field.

Since `Pattern` instances are immutable (according to the javadocs), I think this should be initialized outside as a `private final static` variable to avoid the unnecessary overhead.

petebankhead:
Looks good, linking to the issues this addresses:
* https://github.com/qupath/qupath/issues/667
* https://github.com/qupath/qupath/issues/669

In testing this, I notice one pre-existing bug: I can input a negative magnification value and tbh I don't know how the viewer decides what to use. But that was already the case in v0.2.3 and can be handled in another pull request.

PR #671, opened 2021-03-05
Added recursive PathObject transform

melvingelbard:
- Added transformObjects(..) in PathObjectTools, which recursively transform a PathObject and all its descendants, 'transforming all the parents-children relationships' along with the newly-created objects.
- TODO: Create tests

### DIFF
--- a/qupath-core/src/main/java/qupath/lib/objects/PathObjectTools.java
+++ b/qupath-core/src/main/java/qupath/lib/objects/PathObjectTools.java
@@ -991,36 +991,12 @@ public class PathObjectTools {
 	 * @param copyMeasurements
 	 * @return
 	 */
-	public static List<PathObject> transformObjects(PathObject pathObject, AffineTransform transform, boolean copyMeasurements) {
-		List<PathObject> newObjs = new ArrayList<>();
-		transformRecursive(pathObject, transform, copyMeasurements, null, newObjs);
-		return newObjs;
-	}
-	
-	/**
-	 * Helper method for {@link #transformObjects(PathObject, AffineTransform, boolean)}.
-	 * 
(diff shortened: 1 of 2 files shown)


petebankhead:
Looks good! One thing: it's not possible to transform any object without a ROI (due to a limitation of the [original transform method](https://github.com/qupath/qupath/blob/e7e46dcc15b66672dfa2c10c2776a9df0a4c2ae9/qupath-core/src/main/java/qupath/lib/objects/PathObjectTools.java#L957)).

I think this becomes very useful now that recursive transforms as possible - since transforming the root would mean the whole hierarchy is transformed.

Would it be possible to include this? Here's a script that can be used for exploration:

```groovy
double dx = getCurrentServer().getWidth()/2
double dy = getCurrentServer().getHeight()/2

def transform = java.awt.geom.AffineTransform.getTranslateInstance(dx, dy)
transform.rotate(10)
transform.translate(-dx, -dy)

def parent = getSelectedObject() // Should work
//def parent = getCurrentHierarchy().getRootObject() // Doesn't work (but should)
def newObject = PathObjectTools.transformObjectRecursive(parent, transform, true)


addObject(newObject)
fireHierarchyUpdate()
```

I *think* the cleanest way to handle this is by supporting objects without ROIs (and root objects) within `transformObject` itself, since then it effectively becomes capable of at least duplicating any object... even if the affine transform itself is ignored. The alternative is to handle missing ROIs *only* within `transformObjectRecursive`... which conceptually makes sense too, but I think makes it harder to code against since special cases need handled elsewhere.

melvingelbard:
Closing this PR and recreating a 'cleaner' PR [here](https://github.com/qupath/qupath/pull/682).

PR #679, opened 2021-03-08
Add CircularSlider for image rotation

mahdilamb:
Replace slider in View -> Rotate image with a circular slider

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -852,7 +852,3 @@ task createPackage(dependsOn:createRuntime, type:Exec) {
   }
 
 }
-task circularSliderTest(type: JavaExec, dependsOn: ['testClasses']) {
-    classpath sourceSets.test.runtimeClasspath
-    main = "Test"
-}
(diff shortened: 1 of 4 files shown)


petebankhead:
Thanks, looks good - although I remember now the things we discussed back in October:
* It takes a lot of space in its own window. This is potentially something a user will have open continually. It would therefore be preferable to have it as a 'floating circle' without the wasted space, moved via click & drag. See e.g. *View &rarr; Show input display*.
* *Straighten TMA* should be removed; *Reset* also doesn't need a separate button - it can be handled some other way (e.g. right-click context menu; whatever the option is, the user needs to be notified at least via tooltip)
* Please fix the consistency of the source file to fit with the existing code, i.e. make sure that GPL notice is at the top of the file, and use the same approach to capitalization in the javadocs.
* Add yourself with an `@author` tag for the `CircularSlider` source if you like (similar to how it is done in other files)

It also doesn't play properly with multiple viewers (right-click on a viewer and select from the 'Multi-view' column); behavior is confusing if different viewers are rotated at different times. If this would take too long to resolve, you could leave it and we'll handle it another time.

PR #685, opened 2021-03-19
Update build.gradle for Java 16

petebankhead:
This involves several changes:
- Use toolchain rather than target version.
  - This is required because Gradle can't currently be run using Java 16, and so the build script needs to be called using a different Java version than the one used to build QuPath. If Java 16 isn't available, Gradle will download it.
- Strip the version number from the application name.
  - This is required to avoid package failures due to the .cfg file name being different. *Unfortunately, this may cause problems for users wanting multiple versions installed simultaneously.* The workaround is to rename/use the .zip installations on Windows.
- Explicitly specify the java.library.path to be $APPDIR
  - This is required to load OpenSlide/JPen. The library path was previously set automatically by jpackage in Java 14 (but not 15).
- Explicitly use java-options in the cfg file when changing the memory settings
- Avoid setting the version on macOS
  - Sadly, we cannot start a version number with 0

These changes should also address the issues that arose when building QuPath v0.2 with Java 14: https://github.com/qupath/qupath/issues/615

Several other changes were made:
- Update gradlew
- JavaCPP now uses platform plugin for better dependency management
- Guava version updated
- License report plugin updated
- include jdk.jsobject


petebankhead:
The updates required for Java 16 were more substantial than I'd hoped... this is my first attempt. It relates to:
* https://github.com/qupath/qupath/issues/615
* https://github.com/qupath/qupath/pull/639
* https://github.com/qupath/qupath/pull/640

I'm not entirely satisfied with the outcome, particularly in dropping the version number and whether this will cause Windows installation pains for any update, but I don't have more time to grapple with it at the moment. And continuing to rely on Java 14 for the next release seems worse.

Feedback / suggestions welcome :)

@EgorBushmelev @yli-hallila @anotherBunchOfAtoms @ertugrulcetin @juliangilbey @ajay1685 @melvingelbard

ajay1685:
I build and tested commit [6b09b5a](https://github.com/qupath/qupath/pull/685/commits/6b09b5aafc63ef2456bb9ba3be4779a1eceadcef) with tensorflow-gpu option, it seems to work well with RTX 2080ti, CUDA 11.0 and cuDNN 8.0.4 on Windows 10. QuPath with previous version of tensorflow java (1.15) was very unpredictable when using GPU.

petebankhead:
Excellent, thanks @ajay1685! I hadn't been able to test tensorflow on Windows yet, that's great news.

Some of the more recent commits have hopefully resolved lingering issues with OpenSlide/Bio-Formats, and restored the version number to the application name on Windows. I *think* this is a good thing for people using the .msi installer, since it means they can easily have multiple versions installed. But I don't use Windows enough to know for sure if that's a desirable feature for most people.

ajay1685:
@petebankhead I use QuPath mostly on Win10/7 with or without GPU. I usually try to always install and use the most recent release version. You have made it very easy since version 0.2.X (by maintaining backward compatibility for projects). It is nice to have the msi installer and the ability to install multiple versions for advanced users. Occasionally that has created confusions for some of our users who are not so familiar or up-to-date with QuPath development and; not realizing consequences of working with the incorrect version. Since it is a per user install, It would be okay if the installer overwrites/upgrades the existing installation instead of creating a parallel install. In the event I do need a different version (Old or recent Snapshot) I can always use the zip version instead of having multiple versions installed. I do not have clear understanding of software versioning but is there any specific reason you are not using 1.X.X version for the next release! and not 0.X.X ?

petebankhead:
Thanks @ajay1685, this is very useful.

The rationale behind the versioning is explained at https://qupath.readthedocs.io/en/latest/docs/intro/versions.html

It is getting to the point where it makes sense to call a version v1.0.0, I'm just apprehensive about doing it with funding running out this year and no time to write another publication that reflects the substantial changes.

PR #687, opened 2021-03-26
Added tests in core

melvingelbard:
<h1>qupath.lib.analysis.images</h1>

- SimpleImages.java
	1. Created tests (TestSimpleImages.java)

<h1>qupath.lib.analysis.stats</h1>

- ArrayWrappers.java
	1. Created tests (TestArrayWrappers.java)
- Histogram.java
	1. Created tests (TestHistogram.java)
	2. Changed bin calculation in buildHistogram() to use getBinIndexForValue()
- RunningStatistics
	1. Created tests (RunningStatistics.java)

<h1>qupath.lib.analysis.stats.survival</h1>

- KaplanMeierData
	1. Created tests (TestKaplanMeierData.java)

<h1>qupath.lib.awt.common</h1>

- AwtTools
	1. Created tests (TestAwtTools.java)
	2. Suppressed default constructor for non-instantiability
- BufferedImageTools
	1. Created tests (TestBufferedImageTools.java)
	2. Suppressed default constructor for non-instantiability
	3. Added 'breaks' to switch statement in setValues(...)

<h1>qupath.lib.classifiers</h1>

- CompositeClassifier

	1. Created tests (TestCompositeClassifier.java)
	2. Added comment in classifyPathObjects() to clarify the value of the returned int
- PathClassifierTools
	1. Suppressed default constructor for non-instantiability
	2. Fixed bug that did not handle missing measurement values (NaN) in setIntensityClassification(...). Now reset the PathClass to the non intensity ancestor class of the current object's one.
	3. Throws an Exception if supplying an empty or null measurement name in setIntensityClassification(), as it can't perform classification without a measurement.
- PathIntensityClassifier
	1. getRequiredMeasurements() with null measurement now returns an empty list to avoid NPE when calling getRequiredMeasurements() in CompositeClassifier
	2. Fixed bug that did not handle missing measurement values (NaN) in setIntensityClassification(...). Now it does not change the class of the object if measurement is missing. Why? Because if we run a CompositeClassifier, one classifier's classification shouldn't be reset by the next one.

<h1>qupath.lib.common</h1>

- ColorTools
	1. Suppressed default constructor for non-instantiability.
	2. Made class final.
- GeneralTools
	1. Suppressed default constructor for non-instantiability.
	2. Made class final.
	3. Added some tests to existing tests.
	4. Fixed blankString(...), which did not use the 'trim' param and was always trimming the input String.
	5. Fixed isMultipartExtension(...) to account for String of length == 1.
	6. almostTheSame(...) now relies on Apache common's equalsWithRelativeTolerance() method.
	7. toURI(...) called with an empty String would return a path leading to QuPath's core module. Now returns an empty URI.
	8. arrayToString(...) always used a space as a delimiter. Now uses the delimiter param instead.
	9. arrayToString(...) always added a delimiter after last element. Now fixed.
	10. Changed arrayToString(Object[], ..)'s javadocs from 'separator' to 'delimiter' to match other methods
	11. generateDistinctName(...) now returns the input string if it is empty (instead of unexpected behaviour if list also had an empty String)

<h1>qupath.lib.object.classes</h1>

- PathClassTools
	1. Created tests (TestPathClassTools.java)
	2. Rephrased part of javadocs for getNonIntensityAncestorClass()
	3. Replace verbose OR operator in if statement with static method in isPositiveOrGradedIntensityClass()
	4. Remove 'distinct' from splitNames()'s javadocs, as the returned list can contain duplicates
	5. Added TODO: Consider moving this to ColorToolsAwt? for averageColors() as it is static and could be repurposed.


petebankhead:
Looks great! Excellent to have more tests & the changes documented here.

I'd like to merge this, there are just a few things:
* Please make sure the GPL copyright header is added to any new files (e.g. `TestSimpleImages.java`)
* Could you add a test for `PathClassifierTools.setIntensityClassifications`? That fix looks a really important one - it would be good to have a few examples that test the behavior to guard against future regressions (including at least one that would fail with the old method).
* There is a merge conflict reported with `GeneralTools`
* I'm not sure about the change to `GeneralTools.generateDistinctName`. If the base is empty, it is returned unchanged - but it still *should* provide a distinct name (for example, this is used when adding images to a project and multiple empty names would be very confusing). The generated name could be `(1)` for example; it shouldn't include a leading space.
I haven't tested the following code, but it might do the job:
```java
public static String generateDistinctName(String base, Collection<String> existingNames) {
		if (!existingNames.contains(base))
			return base;
		
		// Check if we already end with a number, and if so strip that
		if (Pattern.matches(".* (\\([\\d]+\\))$", base)) {
			base = base.substring(0, base.lastIndexOf(" ("));
		}
		
		// Check for the highest number we currently have
		int lastInd = 0;
		var pattern = base.isEmpty() ? Pattern.compile("\\(([\\d]+)\\)") : Pattern.compile(base + " \\(([\\d]+)\\)");
		for (var existing : existingNames) {
			var matcher = pattern.matcher(existing);
			if (matcher.find())
				lastInd = Math.max(lastInd, Integer.parseInt(matcher.group(1)));
		}
                if (!base.isEmpty())
                    base = base + " ";
		return base + "(" + (lastInd + 1) + ")";
	}
```
I think this method also deserves its own test, since it's behavior is quite well-defined - but in practice quite fiddly to implement.

melvingelbard:
Thanks for the feedback! 
I've made the modifications to address all your bullet points and edited the PR description.
If there's anything more I can do, please do tell me!

PR #688, opened 2021-03-30
Support of display settings    &     JPEG support for nChannels>1

iwbh15:
Support of display settings in ImageAlignmentPane.java
(see https://forum.image.sc/t/qupath-scripting-4-image-combiner-register-images-and-create-project-entry/50484/6)

JPEG support for n>1 channels in OMEPyramidWriter.java
(see https://forum.image.sc/t/qupath-jpeg-support-for-multi-channel-8bit-ome-tiff-export/47055)

### DIFF
--- a/qupath-experimental/src/main/java/qupath/lib/gui/align/ImageAlignmentPane.java
+++ b/qupath-experimental/src/main/java/qupath/lib/gui/align/ImageAlignmentPane.java
@@ -124,6 +124,8 @@ import qupath.opencv.tools.OpenCVTools;
  * 
  * @author Pete Bankhead
  *
+ * modified by @phaub , 04'2021 (Support of viewer display settings)
+ * 
  */
 public class ImageAlignmentPane {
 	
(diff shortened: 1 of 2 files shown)


petebankhead:
Thanks, nice idea to provide a way to get the rendering settings!

Happy to merge the rendering change; could you resubmit the pull request targeting it at the [`dev-0.3`](https://github.com/qupath/qupath/tree/dev-0.3) branch please?

There is a small issue with the JPEG export: it can quietly fail and produce an image with all zeros if the user specifies to use >3 interleaved channels.

I realise this takes some effort on the user's part, but I think we need to protect them from this by logging a warning and switching to a valid export option, since in some cases requesting interleaved output may be much more efficient (because of the way in which QuPath requests channels all at once during writing, and planes must be written sequentially). Here is a script to replicate it:

```groovy
// Can use an 8-bit version of ImageJ's 'Neuron' sample image to test this (5 channels)

import qupath.lib.images.writers.ome.*

def path = buildFilePath(PROJECT_BASE_DIR, "my_export.ome.tif")
new OMEPyramidWriter.Builder(getCurrentServer())
    .channelsInterleaved() // Fails, gives image of zeros - channelsPlanar() would work
    .compression(OMEPyramidWriter.CompressionType.JPEG)
    .build()
    .writePyramid(path)
```

PS. How should `::dip` be interpreted in the comments?

iwbh15:
JPEG export
I can confirm the 'small issue'.
It is fixed with a sanity check in build().
@petebankhead Do you think this is suitable?

petebankhead:
Looks good to me, I've merged https://github.com/qupath/qupath/pull/692 so will close this now - thanks.

PR #699, opened 2021-04-06
Add read only mode to QuPathGUI

petebankhead:
This provides a mechanism to suppress 'save' dialogs if QuPath is being run simply as a viewer but without the ability to save changes to any image.

Note that this minimal implementation is for feedback; it may need to be changed in a future version if it proves insufficient in practice.

### DIFF
--- a/qupath-gui-fx/src/main/java/qupath/lib/gui/QuPathGUI.java
+++ b/qupath-gui-fx/src/main/java/qupath/lib/gui/QuPathGUI.java
@@ -265,6 +265,8 @@ public class QuPathGUI {
 			);
 	
 	private BooleanProperty selectedToolLocked = new SimpleBooleanProperty(false);
+
+	private BooleanProperty readOnly = new SimpleBooleanProperty(false);
 	
 	// ExecutorServices for single & multiple threads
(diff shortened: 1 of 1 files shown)


petebankhead:
Hi @yli-hallila a proposed change, as we discussed, to enable QuPath to be used read-only without lots of annoying dialogs appearing.

Would this be useful for you?

It's difficult to know how far to push this: for example, should read-only mode disable save menu items, or prevent adding entries to a project? Since it could become more complex, in this case I went with a minimal implementation as a discussion starter.

yli-hallila:
This would be perfect for our needs!

Integrating it with menu items and other areas of QuPath is perhaps something to think about (is a built-in read-only mode something people want? -- perhaps you want to view a project without making any modifications?). Currently I can disable any (most?) buttons / menu items by using `QuPathGUI.lookupActionByText(...);`, so I personally don't need a more complex implementation.

petebankhead:
Thanks, that sounds good! In that case I prefer the minimal implementation for now - it *might* require a compatibility-breaking change in the future (and therefore an update on your side), but I strongly suspect we will be the only people using it for a while.

PR #711, opened 2021-04-13
Added tests

melvingelbard:
- Added tests for PathClassFactory.
- Made class final and restricted its constructor.
- Added a TODO about PathClass validity


qupath-core/src/main/java/qupath/lib/objects/classes/PathClassFactory.java
```diff
@@ -148,13 +153,14 @@ static boolean classExists(String classString) {
 	 * @param name
 	 */
 	private static void validateName(String name) {
```
petebankhead, line 155: I think the original implementation is 'correct' although admittedly the terms are unclear: the name component must not contain a colon, but a single `PathClass` derived from other `PathClasses` can have a `toString()` method that contains a colon as a separator.

The javadocs should probably reflect this, or rename the method - but I think the colon check should be kept.

qupath-core/src/test/java/qupath/lib/objects/classes/TestPathClassFactory.java
```diff
@@ -0,0 +1,143 @@
...
+
+import qupath.lib.common.ColorTools;
+
+@SuppressWarnings("javadoc")
+public class TestPathClassFactory {
+	
+	@Test
+	public void test_getPathClass1() {
+		sameClass(new PathClass("Ignore*", ColorTools.makeRGB(180, 180, 180)), PathClassFactory.getPathClass(PathClassFactory.StandardPathClasses.IGNORE));
+		sameClass(new PathClass("Image", ColorTools.makeRGB(180, 180, 180)), PathClassFactory.getPathClass(PathClassFactory.StandardPathClasses.IMAGE_ROOT));
```
petebankhead, line 43: This works for me in eclipse, but fails when running with `gradlew` with an `java.lang.UnsupportedOperationException` - which it should, if the class already exists.

I presume the order in which the tests run may be making a difference here.

(`makeRGB` should also now be `packRGB`)

qupath-core/src/test/java/qupath/lib/objects/classes/TestPathClassFactory.java
```diff
@@ -0,0 +1,143 @@
...
+		sameClass(new PathClass("Region*", ColorTools.makeRGB(180, 180, 180)), PathClassFactory.getPathClass(PathClassFactory.StandardPathClasses.REGION));
+		sameClass(new PathClass("Stroma", ColorTools.makeRGB(180, 180, 180)), PathClassFactory.getPathClass(PathClassFactory.StandardPathClasses.STROMA));
+		
+		// Can't use PathClass' constructor for 'Tumor' as getDerivedPathClass() creates an instance of 'Tumor' already.
+		sameClass(PathClassFactory.getPathClass("Tumor"), PathClassFactory.getPathClass(PathClassFactory.StandardPathClasses.TUMOR));
+	}
+	
+	@Test
+	public void test_getPathClass2() {
+		Assertions.assertThrows(IllegalArgumentException.class, () -> PathClassFactory.getPathClass("Class with" + System.lineSeparator() + "new line", Color.RED.getRGB()));
```
petebankhead, line 58: Conceivably using `System.lineSeparator()` could give a test that works on Windows and fails on Linux, or vice versa. Consider instead adding two tests, with `\n` and `\r\n`

PR #714, opened 2021-04-16
Support of display settings in autoAlign()

iwbh15:
from message in image.sc:
... images can be aligned based on specific channels. e.g. brightfield images can be aligned based on the DAB channel or the ODsum …


petebankhead:
Thanks, I'll close this now because v0.3.0 is finally released - and the [QuPath Align extension](https://github.com/qupath/qupath-extension-align) has now been separated out to its own repository. This should make it much easier to create forks and/or develop it at a faster pace, without needing to fit it in with QuPath's (currently rather slow) release cycle.

PR #718, opened 2021-04-28
Created rotation property in QuPathViewer

melvingelbard:
- Created rotation property in `QuPathViewer`.
- Made the `rotationProperty` of the `CircularSlider` in radians, while keeping the displayed values in degrees.
- Moved logic of 'Rotate image' command into its own class (to restrict the creation of the pane to a single instance).
- Bound viewer's rotation property to the `Circular Slider` from 'Rotate image' command.


qupath-gui-fx/src/main/java/qupath/lib/gui/viewer/QuPathViewer.java
```diff
@@ -2749,28 +2758,38 @@ protected void updateAffineTransform() {
 	 * @param theta
 	 */
 	public void setRotation(double theta) {
```
petebankhead, line 2760: This can fail when having two viewers, turning off synchronization, rotating one a little, and then turning synchronization back on. Subsequently when rotating around a full circle, at one point the other viewer will be 'locked' in position until being released after a threshold rotation is reached.

I think the following method works (not very tested!)
```java
        public void setRotation(double theta) {
		if (rotationProperty.get() == theta)
			return;
		while (theta < 0)
			theta += 2 * Math.PI;
		theta = theta % (2 * Math.PI);
		rotationProperty.set(theta);
	}
```

qupath-gui-fx/src/main/java/qupath/lib/gui/commands/RotateImageCommand.java
```diff
@@ -0,0 +1,125 @@
+package qupath.lib.gui.commands;
```
petebankhead, line 1: There should be a license header here

qupath-gui-fx/src/main/java/qupath/lib/gui/commands/RotateImageCommand.java
```diff
@@ -0,0 +1,125 @@
...
+			dialog.setResizable(true);
+		}
+		if (viewerTemp != null)
+			slider.rotationProperty().bindBidirectional(viewerTemp.rotationProperty());
+		return dialog;
+	}
+
+	@Override
+	public void changed(ObservableValue<? extends QuPathViewer> v, QuPathViewer o, QuPathViewer n) {
+		o.rotationProperty().unbindBidirectional(slider.rotationProperty());
```
petebankhead, line 121: I'm not sure where precisely the issue occurs, but there's a problem when having two viewers containing images with different rotations (i.e. not synchronized).

If I then turn synchronization on, whenever I click to activate a different viewer its rotation will immediately snap to match the rotation of the previously-selected viewer.

I think synchronization should always work with relative changes, so that any fixed transform between images is preserved.

PR #719, opened 2021-04-30
Changed getBinIndexForValue method

melvingelbard:
There are now 2 versions of the same method:
- `getBinIndexForValue(double, double)`, which makes the assumption that all bins are of equal width (efficient).
- `getBinIndexForValue(double)`, which does not make assumptions regarding bin width and performs further checks (not efficient).

Histograms now use the former method, as the latter one proved dramatically slower for opening some images (e.g. `CMU-1.svs`).

### DIFF
--- a/qupath-core/src/main/java/qupath/lib/analysis/stats/Histogram.java
+++ b/qupath-core/src/main/java/qupath/lib/analysis/stats/Histogram.java
@@ -203,7 +203,7 @@ public class Histogram { // implements Serializable {
 	 * @return bin index
 	 * @see #getBinIndexForValue(double, double)
 	 */
-	public int getBinIndexForValue(double value) {
+	int getBinIndexForValue(double value) {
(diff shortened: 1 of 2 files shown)


petebankhead:
Hi @melvingelbard if I remember the Histogram code well enough, the bin width is fixed during creation and can't be changed. Therefore I don't think there should be two public methods to get an index, since the index won't be meaningful if the consumer specifies a bin width that doesn't match the actual histogram data.

If there's a need to retain both methods non-publicly, can QuPath automatically choose the 'correct' one based upon knowing whether or not the bins have a fixed width?

Keeping methods at least package private would mean they could both be called from tests, if there's a need to ensure that they behave properly.

melvingelbard:
Hi @petebankhead , you're absolutely right. I don't think there's any chance of having histograms with different bin widths with the current implementation. I'll change the visibility of the non-efficient method to package-private.

PR #720, opened 2021-05-03
Added copy to clipboard action for measurement list

melvingelbard:
Added copy to clipboard action for measurement list (via `Ctrl/CMD` + `C`)

### DIFF
--- a/qupath-gui-fx/src/main/java/qupath/lib/gui/Menus.java
+++ b/qupath-gui-fx/src/main/java/qupath/lib/gui/Menus.java
@@ -134,7 +134,6 @@ class Menus {
 
 		// Copy actions
 		@ActionMenu("Copy to clipboard...>Current viewer")
-		@ActionAccelerator("shortcut+c")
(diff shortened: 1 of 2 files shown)


Svidro:
Yayyyyyyyy

petebankhead:
Looks good but unfortunately doesn't work on a Mac :( Or, rather, it seems to copy fine - but then immediately overwrite the clipboard contents with a screenshot of the current viewer (from the *Edit &rarr; Copy to clipboard... &rarr; Current viewer* shortcut).

Switching to an event filter or to use key released didn't resolve it.

I can think of three workarounds:

1. Add a context menu, so that Mac users have an alternative option
2. Disable copying to clipboard if the viewer isn't directly in focus
3. Turn off the copying a current viewer screenshot to the clipboard shortcut altogether, so that it can't conflict with Ctrl/Cmd + C (I guess copying a screenshot isn't *that* common a task?)

What do you think? I can't use Ctrl/Cmd + C in a measurement table of any kind, I guess because of the eagerness of QuPath to make screenshots.

petebankhead:
Works now, thanks!

PR #725, opened 2021-05-07
Optional args on import, Bio-Formats dimension swapping

petebankhead:
Support specifying optional (currently space-delimited) arguments in the import dialog for images.
Enable DimensionSwapper with Bio-Formats.

Putting these together, one can use a String such as
```
  --dims XYTZC --series 2
```
to specify that only the third series should be read from a file with Bio-Formats and the dimension order should (possibly) be changed.

One can also use pass custom arguments for specific readers, e.g.
```
--bfOptions zeissczi.autostitch=false
```
however be warned that these do not play nicely with memoization, meaning that QuPath may need to be restarted for them to work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ Enhancements:
 * Improved command line
   * Specify script parameters with the --args option
   * Return a non-zero exit code if an exception is thrown (https://github.com/qupath/qupath/issues/654)
+* New 'Optional args' when importing images
+  * This allows options such as `--swap XYTZC --series 2` to be passed to Bio-Formats to customize image import
 * New 'ContourTracing' class to simplify converting thresholded and labeled images to ROIs and objects
 * New PathObjectTools.transformObjectRecursive method to simplify applying an affine transformation to objects
 * Many improvements to ImageOps and OpenCVTools to make scripting with OpenCV much easier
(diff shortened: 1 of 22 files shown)


petebankhead:
It is also finally possible to change the ordering or RGB channels with
```
--order BGR
```
or any other permutation of `RGB` (each letter should appear once).

Other options are
```
--classname bioformats/openslide/imagej
```
to specify the reader, or
```
--rotate ROTATE_90
```
to specify rotation.

petebankhead:
As part of this, the recommended way to get an `ImageServer<BufferedImage>` is now via
```
ImageServer.buildImage(URI, args)
```
This will apply any known transforms required by the specified args. The older `ImageServerProvider.buildServer()` method remains for backwards compatibility, and in any case where someone wants to skip the args logic.

petebankhead:
One more substantial change: JSON is now stored in the `.qpdata` file, rather than a server path.

Storing the server path was sufficient in v0.1.2, since this was all that was used to determine which image should be loaded. However, increasingly `ImageServer` implementations require more information / arguments or even wrap other servers, which means necessitated a switch to use JSON. However, in v0.2 this JSON was only stored in the project, not the `.qpdata` file. The result was that such images could *only* be opened from the project; the `.qpdata` alone wasn't sufficient.

This change updates `.qpdata` to store JSON in the place where previously only a path was used. This should allow more complex images to be re-opened successfully, even if separated from their project.

However, since it's a data file change this requires careful testing...

Issue #758, opened 2021-06-27
Sometimes getSelectedObject() returns null unexpectedly


petebankhead:
## Bug report

**Describe the bug**
In QuPath <= 0.2, it's possible to have selected objects but for `getSelectedObject()` to still return null.

This is kind of a bug, while kind of not. It happens because QuPath distinguishes between the 'primary' selected object, and the set of all selected objects.

However, it often *feels* buggy and understandably causes confusion at times, e.g. https://forum.image.sc/t/error-when-exporting-region-of-the-image-in-script/50269/3

**To Reproduce**
Run the following script in v0.2.3 on an image that contains annotations:
```groovy
resetSelection()
selectAnnotations()
println getSelectedObject()
println getSelectedObjects()
```
You should find that it prints `null` for `getSelectedObject()` selected object, *even though clearly annotations are selected*.

This is most confusing whenever there is only one annotation on the image, and there would be no ambiguity as to which one is the primary.

**Expected behavior**
I don't know... but my feeling is that, if there are multiple selected objects, QuPath should always enforce that one is the primary.

That means a [`PathObjectSelectionModel`](https://github.com/qupath/qupath/blob/88c7cc45648c1d5b09a840bd1e497ea9a46453aa/qupath-core/src/main/java/qupath/lib/objects/hierarchy/events/PathObjectSelectionModel.java#L62) change for whenever a set is selected but the primary is `null`.

I propose:
* If a primary selected object is specified, use that. Otherwise
  * If the current primary selected object remains selected (i.e. it's still in the collection passed to the method), just keep it selected. Otherwise
    * Select the first object in the collection passed to the select method to use as the primary

Another idea was to keep the object with the largest ROI, or at the lowest hierarchy level... but those all sounded more involved.

**Desktop (please complete the following information):**
 - QuPath v0.2.3 and earlier

**Additional context**
I've changed the code already and was planning to merge it, but then figured I'd post here first. If anyone is watching, feedback / opinions / objections welcome!

camlloyd:
> You should find that it prints null for getSelectedObject() selected object, even though clearly annotations are selected. 

Correct.


> I propose:
> 
>     * If a primary selected object is specified, use that. Otherwise
>       
>       * If the current primary selected object remains selected (i.e. it's still in the collection passed to the method), just keep it selected. Otherwise
>         
>         * Select the first object in the collection passed to the select method to use as the primary
> 
> 
This makes sense, but are there use cases where it would not be appropriate?

[...]

PR #759, opened 2021-06-27
Ensure selecting objects also sets a primary selected object

petebankhead:
Aims to resolve https://github.com/qupath/qupath/issues/758
Added some additional logging.


petebankhead:
Closing this in favor of https://github.com/qupath/qupath/pull/768
Discussion convinced me that the change might cause more confusion than it resolves, so for now leaving it up to a very minor logging change (and a bigger logging change that makes setting the log level easier).

Issue #628, opened 2020-10-25
Install on Ubuntu 20.04, image artifacts in slide pane


imgifty:
## Bug report

**Describe the bug**

**To Reproduce**
1. Download release 0.2.3 on  Ubuntu 20.04
2. Extract archive
3. Run qupath from bin directory
4. Open project and import mirax slide

**Expected behavior**
Slide without weird artifacts

**Screenshots**
![Screenshot from 2020-10-25 15-29-30](https://user-images.githubusercontent.com/6561496/97109997-e96d5100-16d6-11eb-9c68-d20863276593.png)

**Desktop (please complete the following information):**
 - OS: Ubuntu 20.04
 - QuPath Version 0.2.3 (also doesn't work on 0.2.1)

petebankhead:
Hi @imgifty, have you checked if it could be the same as https://github.com/qupath/qupath/issues/355?
This is referenced in the docs [here](https://qupath.readthedocs.io/en/latest/docs/intro/installation.html#linux).

See also
* https://github.com/qupath/qupath/issues/627
* https://github.com/libvips/libvips/issues/1862#issuecomment-715237292
* https://forum.image.sc/t/change-image-provider-after-import/42502/3
* https://forum.image.sc/t/install-and-start-qupath-ubuntu-20-04/39580/11

In summary, the specific problem with some .mrxs images isn't something that can be resolved directly inside QuPath - rather, it depends upon OpenSlide and/or its dependencies. You might be able to fix things by installing/updating OpenSlide via a package manager.

(There *is* a more general QuPath issue inasmuch as it needs a better way to support OpenSlide in the future... I'll create another issue to discuss that.)

imgifty:
Hello @petebankhead ,
I reinstalled Pixman-0.40.0 and this indeed fixed my problem, thanks a lot!

Thanks for the speedy help!

phromo:
@imgifty can you help out by providing how you reinstalled pixman? The latest version I can find is 0.38.4-0ubuntu1 for focal.

phromo:
I rebuilt the deb myself. If anyone else goes looking, you might find my packages for ubuntu 20.04-focal here: https://www.lupin.se/~phromo/pixman/ -- use with caution, I'm a noob at debian package building.

imgifty:
> @imgifty can you help out by providing how you reinstalled pixman? The latest version I can find is 0.38.4-0ubuntu1 for focal.

[...]

Issue #629, opened 2020-10-25
OpenSlide support in QuPath


petebankhead:
QuPath supports reading whole slide images using two open source libraries: [Bio-Formats](https://www.openmicroscopy.org/bio-formats/) and [OpenSlide](http://openslide.org).

Of these, Bio-Formats is much more extensive and actively-maintained; OpenSlide only really supports single-plane, 8-bit RGB images and has not been meaningfully updated for a couple of years. Nevertheless, OpenSlide remains important to QuPath users because
* A few common file formats (e.g. .mrxs) are *only* supported by OpenSlide
* Where a format is supported by both Bio-Formats and OpenSlide, OpenSlide is generally much faster at reading image tiles (although may provide less metadata)
* OpenSlide offers alpha information that is relevant if regions are unscanned

The trouble is that OpenSlide is a C library, and getting it to work properly with QuPath is a challenge we haven't managed to solve... thanks to the complexities of how native dependencies - and, more importantly, their sub-dependencies - are handled on different platforms. Bug reports such as
* https://github.com/qupath/qupath/issues/355
* https://github.com/qupath/qupath/issues/628
* https://github.com/qupath/qupath/issues/627

along with breakages caused by `jpackage` updates (https://github.com/qupath/qupath/issues/615) show that the (largely manual) process initially used to try to package up the dependencies hasn't worked.

For these reasons, **we need a new and maintainable approach to include OpenSlide in QuPath in the future, or else drop support entirely**.

Alas, I'm sadly one of those developers who is quite helpless when attempting to build any native libraries with even moderate complexity... spoiled by Java and Python, I haven't been able to figure out a solution. Hence this issue and cry for help.

The biggest problem is getting a truly portable way to build OpenSlide on each platform. If we have that, we can potentially use JNI directly or JavaCPP (https://github.com/bytedeco/javacpp) / Gradle JavaCPP (https://github.com/bytedeco/gradle-javacpp). *Ideally* all dependencies would be packaged into a single .jar file and accessed either with JavaCPP / [native-lib-loader](https://github.com/scijava/native-lib-loader).

An alternative might be to try outsourcing OpenSlide support on macOS and Linux, *requiring* installation with a package manager (?) and only providing the 'official' pre-built binaries on Windows.

In considering this, it is worth asking whether we should be linking to OpenSlide only anyway.
[...]

PR #783, opened 2021-07-15
Optionally set LD_LIBRARY_PATH for Linux

petebankhead:
This avoids the need to create a custom launch script as described at https://github.com/qupath/qupath/issues/628

While still not ideal, it gives an alternative way to avoid using system shared libraries with OpenSlide - which is known to be problematic because of pixman issues on Ubuntu LTS releases (including 20.04).

This works by including a bash script in the jpackage image generated for Linux by default. It also works with
`./gradlew run -Pld-path=true`
and with
`./gradlew installDist -Pld-path=true`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 
 ### Release highlights
 * **New 'Create density map' command** to visualize hotspots & generate annotations based on object densities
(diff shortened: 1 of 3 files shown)


petebankhead:
Better ways to handle this (and to support OpenSlide) would be very welcome.
See also https://github.com/qupath/qupath/issues/629

Issue #799, opened 2021-08-06
Input Display - How do you dismiss it?


Jeff-Haas:
## IMPORTANT!

This section is **only** for bug reports, which **must** follow the template below.

Please post any other questions to the official QuPath forum at https://forum.image.sc/tags/qupath

If in doubt, use [image.sc](https://forum.image.sc/tags/qupath). This helps us keep things organized.

## Before we begin...

Before submitting your bug report, please check the following:

* [X ] I've definitely found a bug (it you're not sure, please use [image.sc](https://forum.image.sc/tags/qupath) instead)
* [X ] I've checked https://qupath.github.io for a new release that might already have fixed the issue
* [X] I've checked the [Changelog](https://github.com/qupath/qupath/blob/master/CHANGELOG.md) to see if the bug has already been fixed in the next release
* [X] I've checked for existing GitHub issues describing the same problem

## Bug report

**Describe the bug**
Unable to turn off Input Display.  Selecting it a second time adds a second version to the screen.  Both versions show keypresses.  Selecting again adds a third one...and so on.

**To Reproduce**
Turn on Input Display (View > Show Input Display)
Notice it shows in lower left of screen.
Try to turn it off by going back to View > Show Input Display.  Note: There is no checkmark showing that it is already active.
Selecting it again adds another one to the screen.
Each additional Input Display can be moved from the lower left corner, and all will show the last key pressed (as long as QuPath was the active application).

**Expected behavior**
The View menu should show a checkmark when the Input Display has been turned on.
Once the Input Display is active, selecting "Show Input Display" will turn off the Input Display and clear the checkmark.

**Screenshots**
![image](https://user-images.githubusercontent.com/27331078/128438550-e734058a-d040-403f-9f5c-507419f24822.png)

I realize these are created as "new" QuPath windows which can be closed from the task bar.  However this fix still seems useful, as the Input Display windows do not have a close box on them.  It took me quite awhile to understand how to close them.

Note that while other windows such as the Memory Monitor act the same way, they have standard title bars with a close box, so it is obvious they can be moved and closed.  It's the special format of the Input Display that confused me.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - QuPath Version 0.3.0 SNAPSHOT.  Also verified in 0.2.3.

**Additional context**

petebankhead:
You can double-click inside to close (this is included in the tooltip text).

[...]

PR #801, opened 2021-08-06
Made 'input display dialog' command a selectable item

melvingelbard:
- Made 'input display dialog' command a 'selectable' item (ticked/not ticked).
- Restricted the amount of input display dialogs to 1.

Spurred by #799 .


petebankhead:
I like it! Behavior seems to me much improved, thanks @melvingelbard and @Jeff-Haas

PR #809, opened 2021-08-26
Simplify DnnModel use with ImageOps

petebankhead:
Reduce support to single-input, single-output models for now, to avoid a strict requirement for input names to be correct.


petebankhead:
Closing in favor of #811

PR #824, opened 2021-09-22
Added predicate parameter to Measurement Exporter

melvingelbard:
Users can now add this parameter to the measurement exporter builder:
```
.filter(obj -> obj.getPathClass() == getPathClass("Tumor"))
```

to filter the path objects that will be exported (in this example, only the path class with 'Tumor' class will be exported).

N.B.: the filter indicates what should be kept, *not* filtered out.

### DIFF
--- a/qupath-gui-fx/src/main/java/qupath/lib/gui/tools/MeasurementExporter.java
+++ b/qupath-gui-fx/src/main/java/qupath/lib/gui/tools/MeasurementExporter.java
@@ -2,7 +2,7 @@
  * #%L
  * This file is part of QuPath.
  * %%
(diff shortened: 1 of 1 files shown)


OstVVind:
Good day. I couldn't find a place to ask a question exactly, I'm sorry for that.
I would like to ask if it is possible to work with slide labels through qupath so that I can export automatically through a script all slide labels from my project to a specific folder?

petebankhead:
Hi @OstVVind the [readme](https://github.com/qupath/qupath) contains links for help and support - the best place to ask questions is the forum at https://forum.image.sc/tag/qupath

PR #827, opened 2021-09-30
Optimize NormalizePercentileOp by histogram calculation

dkurt:
Optimized version gives me about 1 minute less processing time for StarDist on OS-3.ndpi

Proposed to use histogram calculation to compute percentiles. Valid only for RGB images (`[0-256)` pixels range).

### DIFF
--- a/qupath-core-processing/src/main/java/qupath/opencv/ops/ImageOps.java
+++ b/qupath-core-processing/src/main/java/qupath/opencv/ops/ImageOps.java
@@ -510,7 +510,18 @@ public class ImageOps {
 		 * @return
 		 */
 		public static ImageOp percentile(double percentileMin, double percentileMax) {
-			return new NormalizePercentileOp(percentileMin, percentileMax);
+			return new NormalizePercentileOp(percentileMin, percentileMax, true);
+		}
+
+		/**
+		 * Same as {@link #percentile(double, double)}, but optimized for images with values in range [0, 256)
+		 * @param percentileMin
+		 * @param percentileMax
+		 * @param optimized enable optimization
+		 * @return
+		 */
+		public static ImageOp percentile(double percentileMin, double percentileMax, boolean optimized) {
+			return new NormalizePercentileOp(percentileMin, percentileMax, optimized);
 		}
 		
 		/**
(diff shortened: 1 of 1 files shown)


petebankhead:
Thanks! That's an impressive speedup, and a common enough case to deserve an optimized version. I'll check it out (probably for v0.4.0 rather than v0.3.1).

petebankhead:
Hi @dkurt sorry for the long delay, I'm finally getting back to StarDist/ImageOps/normalization before the next release.

In the end, percentile normalization turned out to have a lot of extra complexities that I hadn't originally considered. I've tried to address them in
* https://github.com/qupath/qupath/pull/1130
* https://github.com/qupath/qupath/pull/1146

These are intended to help QuPath fully support the [bioimage.io preprocessing spec](https://github.com/bioimage-io/spec-bioimage-io/blob/gh-pages/preprocessing_spec_0_4.md), which assumes 32-bit input and output. They also add support to normalize channels independently or jointly (since it [turns out some of the StarDist models uses joint channel normalization](https://forum.image.sc/t/stardist-in-qupath-normalization-issue/38912/13)).

I hadn't realised Apache Commons Math provides [10 different methods that can give different results](https://commons.apache.org/proper/commons-math/javadocs/api-3.6.1/org/apache/commons/math3/stat/descriptive/rank/Percentile.EstimationType.html)), so I've now switched to using the one that matches with NumPy (and I think also R, Julia and others).

Along the way, performance should be substantially improved by avoiding full array sorting and using parallelization; I'm seeing reductions from ~1 second to 0.2-0.3s in some tests.

Because of these changes, I'd like to close this PR in favor of keeping the code simpler and not introducing an optimized alternative specifically for 8-bit RGB.

Thanks again though - this helped to show me how slow & in need of improvement the original code was :)

(If you ever want to test how the performance of your optimized approach compares with the new version, I'd be curious about the result - but I realise that probably isn't a priority. In any case, my [timing code is here](https://github.com/qupath/qupath/blob/main/qupath-core-processing/src/test/java/qupath/opencv/tools/TestOpenCVTools.java#L487)).

dkurt:
Thanks, @petebankhead!

PR #830, opened 2021-10-14
Small enhancement that enables Delete or backspace key in the AnnotationPane

tp81:
Hi 

I just thought this might be useful to others too. I found myself pressing the "Del" key to delete annotation in the annotation pane. As a workaround pressing the Delete button or clicking in the viewer and pressing the "Del" key. 

This change allows to use the key too. Code was copied over and slightly modified from the viewer code.

Best
Thomas

### DIFF
--- a/qupath-gui-fx/src/main/java/qupath/lib/gui/panes/AnnotationPane.java
+++ b/qupath-gui-fx/src/main/java/qupath/lib/gui/panes/AnnotationPane.java
@@ -42,6 +42,7 @@ import javafx.beans.property.SimpleBooleanProperty;
 import javafx.beans.value.ChangeListener;
 import javafx.beans.value.ObservableValue;
 import javafx.collections.ListChangeListener.Change;
+import javafx.event.EventHandler;
 import javafx.geometry.Side;
 import javafx.scene.control.Button;
 import javafx.scene.control.ContextMenu;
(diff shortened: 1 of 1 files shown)


petebankhead:
Thanks @tp81! It might take me a bit of time to look at this in detail, because I've switched focus temporarily to finish some papers.

First thoughts: I've also had the issue of having to click on the viewer to activate it for backspace to work, I agree it would be best to avoid this. Minor concerns are
* it introduces public methods, and I'd rather reduce this as much as possible
* if it basically duplicates some code elsewhere, perhaps some refactoring elsewhere could promote reuse

There is also a bit of JavaFX/macOS weirdness when it comes to single-key accelerators (i.e. not requiring Cmd+Something to run a menu item). I'm not sure if it's relevant here but anything accelerator-related takes some extra cross-platform testing.

I've assigned it to a v0.4.0 milestone to remind us to try to get this functionality in that release (tentatively planned for early next year).

petebankhead:
Hi @tp81 I've had a look at this and would like to make a counter-offer :)

Would https://github.com/qupath/qupath/pull/848 be a suitable replacement?

Triggering the 'Delete' button feels safer to me than copying the viewer behavior. I think the main difference is that my proposed alternative will always prompt for confirmation, while backspace within the viewer could delete a single selected annotation immediately.

It also means both buttons (*Select all* and *Delete*) have exact keyboard replacements.

What do you think?

tp81:
I think #848 it's wonderful! Beautifully compact change and also more consistent!

Slightly unrelated: if I find another minor thing like this, would you prefer me to first file a 'bug report' and then mention it in the PR?

Thanks for this @petebankhead !

petebankhead:
Great! Change merged now, I'll close this.

Feel free to make the PR without a bug report. I try to keep bug reports only for things that definitely need fixed / will be changed, PRs for potential code to include, and https://forum.image.sc/tag/qupath for all other discussions. I just added the issue here after making the changes to have something to link to in the changelog... although in retrospect I could just have linked to the PR instead :)

PR #835, opened 2021-11-01
Include measurements when creating annotations from imported GeoJSON data

jameshball:
### Reason for change

- Previously, only 'detections' supported importing measurements from GeoJSON, as the `measurementList` is ignored for annotations
- This fixes this, allowing measurements to be imported from a GeoJSON file

### Changes

- Measurements are now not ignored when creating annotations
- Before this, measurements were fetched but not actually used in `PathObjects.createAnnotationObject`
- This does not change behaviour when the `measurements` array is not provided for a feature
  - `measurementList` is `null` initially, and only populated if `measurements` is in the GeoJSON
  - The previous code didn't pass in any list, which calls the overloaded `PathObjects.createAnnotationObject`, which just returns `createAnnotationObject(roi, pathClass, null);` anyway
  - This means code-flow is unchanged if `measurements` are not provided

#### Custom measurements extracted from GeoJSON:
image
image

### Testing

- Works perfectly fine with GeoJSON files that create annotations both with and without a `measurements` field
- Example GeoJSON file I used for testing: https://pastebin.com/trYDYLXM
  - Tested this file with and without the `measurements`

### DIFF
--- a/qupath-core/src/main/java/qupath/lib/io/PathObjectTypeAdapters.java
+++ b/qupath-core/src/main/java/qupath/lib/io/PathObjectTypeAdapters.java
@@ -435,12 +435,12 @@ class PathObjectTypeAdapters {
 			case ("annotation"):
 			case ("unknown"):
 				// Default is to create an annotation
-				pathObject = PathObjects.createAnnotationObject(roi, pathClass);
+				pathObject = PathObjects.createAnnotationObject(roi, pathClass, measurementList);
 				break;
 			default:
 				// Should be called if the type has been specified as *something*, but not something we recognize
 				logger.warn("Unknown object type {}, I will create an annotation", type);
-				pathObject = PathObjects.createAnnotationObject(roi, pathClass);
+				pathObject = PathObjects.createAnnotationObject(roi, pathClass, measurementList);
 			}
 			if (name != null)
 				pathObject.setName(name);


petebankhead:
Thanks @jameshball for the PR and thorough description. All looks good to me, but I need to find a bit more time to check in detail (and try to remember if there was any logical reason to ignore annotations in the first place... I suspect there was not and it was just an oversight).

The reason I want to check a bit more is that this highlights a bit of weirdness in how [QuPath provides different `MeasurementList` implementations](https://github.com/qupath/qupath/blob/88c7cc45648c1d5b09a840bd1e497ea9a46453aa/qupath-core/src/main/java/qupath/lib/measurements/MeasurementListFactory.java#L44), so the annotation isn't guaranteed to have the same type of list after deserialization as it previously had. This was for reasons of efficiency, and because in an early version there was a 'dynamic measurement' feature that only annotations supported (but this isn't now used).

That's a good chance that doesn't matter, but keeping this PR open is a handy reminder to think about measurement lists and GeoJSON in more detail before the next release.

jameshball:
Thanks for taking a look!

That all makes sense to me, we are using GeoJSON quite heavily in our project so that would be very useful for you to have a more thorough look :)

petebankhead:
Hi @jameshball could you have a look at https://github.com/qupath/qupath/pull/846 please, to see if it fixes the problem for you (or if anything else is missing)?

petebankhead:
Closing now that https://github.com/qupath/qupath/pull/846 has been merged, thanks!

Issue #845, opened 2021-11-20
MeasurementList is ignored for some objects when importing from GeoJSON


petebankhead:
## Bug report

**Describe the bug**
See https://github.com/qupath/qupath/pull/835 from @jameshball

In summary, measurement lists are only included by `PathObjectTypeAdapters` when deserialising detection objects, as can be seen in the code from https://github.com/qupath/qupath/blob/v0.3.0/qupath-core/src/main/java/qupath/lib/io/PathObjectTypeAdapters.java#L412

A second, more subtle bug is that when measurement lists *are* retained, their type can be changed. This is not expected to have any major effect beyond efficiency,  but could potentially be important in some situations.

**To Reproduce**
Run the following script:
```groovy
def roi = ROIs.createRectangleROI(0, 10, 20, 30, ImagePlane.getDefaultPlane())
def annotation = PathObjects.createAnnotationObject(roi)
annotation.setName("Original")
annotation.setPathClass(getPathClass("My custom class"))
annotation.getMeasurementList().addMeasurement("First measurement", 1.0)
annotation.getMeasurementList().addMeasurement("Second measurement", -2.0)

def json = GsonTools.getInstance(true).toJson(annotation)

// Print original annotation, class and measurement list
println "ORIGINAL"
println annotation
println annotation.getPathClass()
println annotation.getMeasurementList().class
println annotation.getMeasurementList()
println "----------------"

// Print deserialized annotation, class and measurement list
def newAnnotation = GsonTools.getInstance().fromJson(json, PathObject)
println "FROM JSON"
println newAnnotation
println newAnnotation.getPathClass()
println newAnnotation.getMeasurementList().class
println newAnnotation.getMeasurementList()
```

The new annotation does not have any measurements in QuPath v0.3.0.

Moreover, switching to use `PathObjects.createDetectionObject(roi)` reveals that the measurements *are* retained, but the class of the measurement list is changed from `FloatList` to `DoubleList` - requiring more memory.

**Expected behavior**
As much information as possible is retained when serialising to/from GeoJSON.

**Desktop (please complete the following information):**
 - QuPath v0.3.0

PR #846, opened 2021-11-20
Improve GeoJSON support for MeasurementLists

petebankhead:
Fixes https://github.com/qupath/qupath/issues/845
See also https://github.com/qupath/qupath/pull/835


jameshball (approved):
Works great with annotations with measurements, and fixes the aforementioned issues! Measurements on a particular annotation post-import:
image

petebankhead:
Great, thanks! I'll merge it for inclusion in the next point release then, I don't think the old behavior can count as anything other than a bug. If you spot any other similar problems please let me know.

PR #851, opened 2021-12-02
Added copy/paste option to workflow

melvingelbard:
- Added copy/paste option to workflow (via keyboard shortcut and via right-click/context menu)
- Fixed bug that would make the workflow command empty until a new command was added


petebankhead:
I like it!

Issue #861, opened 2021-12-14
AbstractTileableImageServer can request the same tile in multiple threads simultaneously


petebankhead:
## Bug report

**Describe the bug**
Many `ImageServer` implementations are derived from `AbstractTileableImageServer`, which breaks requests into fixed tiles and fulfils them using a cache where possible.

Duplicate tile requests occur e.g. when writing an OME-TIFF using multithreading or running a pixel classifier and with a tile size larger than the tile size of the reader. It can result in the same image tiles being read multiple times, potentially adding considerable unnecessary overhead (since tile decompression can be a bottleneck).

Reducing or eliminating this can potentially improve performance substantially, particularly for cases where tile reading is slow.

**To Reproduce**
The bug can easily be reproduced when writing a pyramidal OME-TIFF or training a pixel classifier - although unfortunately it isn't easy to *tell* that it has been reproduced, since the duplicate requests aren't reported.

The problem was identified through VisualVM profiling and running in debug mode, checking for repeated tile requests while trying to improve image write performance.

**Expected behavior**
If a thread wants to obtain a tile that is currently being read by another thread, the first thread should wait until the tile is returned - it should *not* submit a new read request.

**Desktop (please complete the following information):**
 - OS: All
 - QuPath Version 0.3.0 (and earlier)

**Additional context**
This isn't really an issue when only viewing images, since requests are made that perfectly match tiles available through the `ImageServer` and pending requests are checked. However it can be an issue when regions are requested in any other way.

Any solution should avoid generating more threads within an `ImageServer`.

Issue #859, opened 2021-12-14
convert-ome command doesn't report when complete


petebankhead:
## Bug report

**Describe the bug**
When converting an image from the command line, progress may be logged - but no message is given when the conversion is complete.

This is often ok when running from a terminal, but is awkward if running in eclipse - where it is less clear that the process is complete.

**To Reproduce**
Run a command such as
```
QuPath convert-ome /path/to/CMU-1-Small-Region.svs /path/to/CMU-1-Small-Region2.ome.tif 
```

Based only on what is printed, it looks like conversion hangs at 97% - when in fact it did terminate properly.
```
07:48:16.170 [main] [INFO ] q.l.i.writers.ome.OMEPyramidWriter - Writing resolution 1 of 1 (downsample=1.0, 30 tiles)
07:48:16.173 [main] [INFO ] q.l.i.writers.ome.OMEPyramidWriter - Writing plane 1/1
07:48:16.394 [main] [INFO ] q.l.i.writers.ome.OMEPyramidWriter - Written 10% tiles
07:48:16.492 [main] [INFO ] q.l.i.writers.ome.OMEPyramidWriter - Written 20% tiles
07:48:16.597 [main] [INFO ] q.l.i.writers.ome.OMEPyramidWriter - Written 30% tiles
07:48:16.660 [main] [INFO ] q.l.i.writers.ome.OMEPyramidWriter - Written 40% tiles
07:48:16.744 [main] [INFO ] q.l.i.writers.ome.OMEPyramidWriter - Written 50% tiles
07:48:16.800 [main] [INFO ] q.l.i.writers.ome.OMEPyramidWriter - Written 60% tiles
07:48:16.864 [main] [INFO ] q.l.i.writers.ome.OMEPyramidWriter - Written 70% tiles
07:48:16.935 [main] [INFO ] q.l.i.writers.ome.OMEPyramidWriter - Written 80% tiles
07:48:17.036 [main] [INFO ] q.l.i.writers.ome.OMEPyramidWriter - Written 90% tiles
07:48:17.091 [main] [INFO ] q.l.i.writers.ome.OMEPyramidWriter - Written 97% tiles
```
**Expected behavior**
A clear message printed when conversion is successful, ideally also with timing information.

**Desktop (please complete the following information):**
 - OS: All
 - QuPath Version 0.3.0

Issue #860, opened 2021-12-14
ImageJServer is very inefficient with large 2D images


petebankhead:
## Bug report

**Describe the bug**
The performance of `ImageJServer` is much worse than it needs to be - to the extent that it can make some images impossible to work with.

This first appeared at https://forum.image.sc/t/qupath-out-of-memory-problem-when-creating-training-image/60985

I was unable to even open a ~3GB image on my computer (with total 16 GB RAM). It turned out that `ImageJServer` was actually requesting the pixels twice: which was extremely slow, and eventually terminated with an OOM.

This occurs because `ImageJServer` first opens the image as a virtual stack, then computes memory requirements/stack size. Based upon this, it may then open the same image as a regular `ImagePlus`. This should be unnecessary, and also temporarily doubles the memory requirements for 2D images because the original has not yet been closed.

When investigating this, three other issues were noticed:

1. The performance impact of using virtual stacks is *much* worse than I had expected, particularly for large 2D images that will frequently be cropped/duplicated. Extracting a region - even for a 2D plane - causes the data to be read from disk again.
2. [`readBufferedImage(request)`](https://github.com/qupath/qupath/blob/75ec9cebe5e3bc5843fc60b07b455ce1215e1fb9/qupath-core-processing/src/main/java/qupath/imagej/images/servers/ImageJServer.java#L298) is synchronized - but the entire method does not need to be. This prevents calls to resize the extracted image being parallelized, even though this should be safe.
3. Resizing is called once per channel, including for RGB images. However, in ImageJ's world the RGB image is treated as a single channel - and so any resizing is actually performed 3x rather than 1x. The last two operations seem to be entirely unnecessary.

**To Reproduce**
The problem should occur when trying to open any large TIFF that has been written by ImageJ. If it is at the bounds of the available memory, this is likely to fail. If enough memory is available, it should succeed but perform badly.

**Expected behavior**
`ImageJServer` should fully read 2D images - since using a virtual stack brings no benefits - and most multidimensional images, as long as they are small enough to fit comfortably in memory.

Synchronization should be reduced, and resizing limited.

It will sometimes still not be possible to open a non-pyramidal image via ImageJ directly. In this case, running `convert-ome` should be able to efficiently generate an OME-TIFF image.

**Desktop (please complete the following information):**
 - OS: All
 - QuPath Version 0.3.0 (and earlier)

**Additional context**
[...]

PR #862, opened 2021-12-14
Reduce unnecessary tile reads in AbstractImageServer

petebankhead:
Aims to resolve
* https://github.com/qupath/qupath/issues/861
* https://github.com/qupath/qupath/issues/859

Achieves this primarily by changing `AbstractTileableImageServer` to retain pending requests as Futures, which can be reused by other threads who need the same tiles.


petebankhead:
Also addresses
* https://github.com/qupath/qupath/issues/860

Issue #865, opened 2021-12-15
Generating multiple readers with Bio-Formats can be very slow


petebankhead:
## Bug report

**Describe the bug**
`BioFormatsImageServer` lazily creates a new reader for each thread as required. This is ok if the readers are fast to initialize, but can be a major bottleneck if they are not.

This was noticed when working with large CZI images (>30 GB). Initializing a reader took ~3-5s. However, because the method was synchronized and 32 threads were wanting to get pixels for the viewer, many other threads were blocked. This meant that even viewing the image properly was not possible for well over a minute.

Once the readers were created, performance was fine.

**To Reproduce**
Unfortunately, I'm not aware of any public images images that can be used to test this. It might be evident with any large Axioscan images (I'm not entirely sure).

Once the image is open, zoom in and wait for tiles to appear. If experiencing the problem, this will take an unreasonable amount of time. VisualVM indicates that the bottleneck is initializing readers.

**Expected behavior**
No major delay in requesting tiles once the image has been opened.

**Desktop (please complete the following information):**
 - OS: macOS (but likely to be all)
 - QuPath Version 0.3.0

**Screenshots**
Compare the 'total time' (>100s) with the actual time spent using the CPU (3.6s) for a thread requesting image tiles.

![Screenshot 2021-12-15 at 12 49 12](https://user-images.githubusercontent.com/4690904/146189714-d4197439-8b67-4b79-bf64-f400de9b7b72.png)

**Additional context**
A few things could help:
* Reduce the maximum number of tile request threads
* Reduce synchronization when creating readers
* Limit the number of readers Bio-Formats can create, independently of the number of threads making tile requests
* Reduce the calls to `isThisType()` when creating a reader (when the class of the reader can be known)

petebankhead:
Upon further investigation, memoization can greatly reduce the severity of the problem - which is probably why it hasn't generated more complaints.

PR #867, opened 2021-12-17
Update multithreading support in BioFormatsImageServer

petebankhead:
Aims to fix https://github.com/qupath/qupath/issues/865
This takes a different approach to parallelization, managing a pool of ImageReaders with each tile-requesting thread taking the next available reader.
If there are no readers available, and the total number is less than some maximum value (based upon the number if available processors), a new reader is generated on another thread and added to the queue when ready.

This should
* avoid generating more readers than needed, with a limit separate from the number of tile requesting threads
* avoid attempting to initialize multiple readers simultaneously, which can be a bottleneck

In addition, more tests have been added.


NicoKiaru:
Just in case that's useful (I may be late...) I use a readerpool in bigdataviewer-playground. I used the class from here:

https://www.dbtsai.com/blog/2013/java-concurrent-dynamic-object-pool-for-non-thread-safe-objects-using-blocking-queue/

It's working great so far.

One thing to take care of is to put tile reading in between a `pool.acquire()` and `pool.recycle(reader)` statements:

https://github.com/BIOP/bigdataviewer-bioformats/blob/bfb48be52694ffebaa03fbbe6339ac509aab66ea/src/main/java/ch/epfl/biop/bdv/bioformats/bioformatssource/BioFormatsBdvSourceUnsignedShort.java#L119-L157

petebankhead:
Thanks @NicoKiaru that sounds pretty similar to what I've tried to do here. As far as I can see, my implementation is working ok so I'll stick with it for now.

One minor difference is that my version restricts creating new readers to a single background thread, since I found that sometimes (although rarely) constructing many reader simultaneously had really bad performance, as all the constructors got stuck parsing XML.

By ensuring readers are created sequentially, I avoid blocking for too long when tile reading is substantially faster than reader initializing: if a thread wants a reader, it puts in its request. If there's a reader available, it gets that immediately. If not (and the max reader limit isn't reached) a background thread goes off to create a new reader. The original thread will then wait for the next available reader - which *might* be the one initialized because of its request, but it might also be one that has since become available. Waiting time is minimized in any case.

I also added some `Cleaner` support from Java 9 to handle the fact that it's easy to miss closing an `ImageServer` properly.

PS. Since the recycling is really important, should you do it in a `finally` block?

NicoKiaru:
> Since the recycling is really important, should you do it in a finally block?

I need to check this! An the cleaner thing as well, but for now, Fiji = Java 8, so yeah, let's see if I can avoid it. 
I'm limiting the number of reader as well, it's a property of my dataset, and it's true that limiting to the number of parallel reader to around 10 usually gives the best result (usually lower than the number of cores).

> One minor difference is that my version restricts creating new readers to a single background thread, since I found that sometimes (although rarely) constructing many reader simultaneously had really bad performance, as all the constructors got stuck parsing XML.

Right! In bdv, a first reader is created to parse metadata and build the memo file, I guess it's similar in QuPath. Then extra readers are creating for tile reading, once the memo file is already created. 

There may be some optimisation possible for these 'extra readers', since metadata is not required anymore: https://forum.image.sc/t/open-czi-in-bio-formats-large-ram-requirements/45592/5, but I did not test it.

PR #931, opened 2022-02-28
Major refactoring of Script Editor

melvingelbard:
_TODO: Description_


qupath-gui-fx/src/main/java/qupath/lib/gui/scripting/ScriptEditorDragDropListener.java
```diff
@@ -0,0 +1,80 @@
...
+			var list = dragboard.getFiles();
+			if (list == null)
+				return;
+			
+			List<File> jars = list.stream().filter(f -> f.getName().toLowerCase().endsWith(".jar")).collect(Collectors.toList());
+			if (!jars.isEmpty() && qupath.canInstallExtensions())
+				qupath.installExtensions(list);
+			
+			List<File> remainingFiles = list.stream().filter(f -> !f.getName().toLowerCase().endsWith(".jar")).collect(Collectors.toList());
+			var supported = ScriptLanguageProvider.getAvailableScriptLanguages().stream().flatMap(l -> Arrays.stream(l.getExtensions())).collect(Collectors.toList());
```
petebankhead, line 67: This could be troublesome - see javadoc at https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/stream/Collectors.html#toList()
There's no guarantee we'll get a mutable list back, so I think this should be changed.

One way would be to just immediately add the list to a `new HashSet<>()` (there's bound to be a more efficient way to add to an existing collection with streams, but I don't remember it...) - since it's really a set rather than a list that we want anyway.

qupath-gui-fx/src/main/java/qupath/lib/gui/scripting/languages/ScriptLanguage.java
```diff
@@ -0,0 +1,88 @@
...
+package qupath.lib.gui.scripting.languages;
+
+/**
+ * Abstract class to represent languages supported by the script editor.
+ * @author Melvin Gelbard
+ * @since v0.4.0
+ */
+public abstract class ScriptLanguage {
+	
+	protected String name;
```
petebankhead, line 33: I think these should be `private` rather than `protected`

qupath-gui-fx/src/main/java/qupath/lib/gui/scripting/languages/ScriptLanguage.java
```diff
@@ -0,0 +1,88 @@
...
+	protected ScriptAutoCompletor completor;
+	
+	/**
+	 * Default constructor for a {@link ScriptLanguage}.
+	 * @param name		the language name
+	 * @param exts			the possible extensions for this language
+	 * @param syntax		the syntax object for this language
+	 * @param completor	the auto-completion object for this language
+	 */
+	protected ScriptLanguage(String name, String[] exts, ScriptSyntax syntax, ScriptAutoCompletor completor) {
```
petebankhead, line 45: I think `ScriptSyntax` and `ScriptAutoCompletor` shouldn't be required here - don't store them in the abstract class, and allow subclasses to return them.

Main reason: subclasses might want to initialise them within the constructor. But the `String` for name and extensions should always be available, so it's ok to make them required by the constructor.

With the current implementation, I don't think this needs to be an abstract class at all (?)

qupath-gui-fx/src/main/java/qupath/lib/gui/scripting/languages/RunnableLanguage.java
```diff
@@ -0,0 +1,63 @@
...
+	 */
+	Object executeScript(final String script, final Project<BufferedImage> project, final ImageData<BufferedImage> imageData, final Collection<Class<?>> defaultImports, final Collection<Class<?>> defaultStaticImports, final ScriptContext context) throws ScriptException;
+
+	/**
+	 * Get the default imports as a String, to add at the beginning of the executed script.
+	 * @param defaultClasses  		a collection of the classes to import 
+	 * @param defaultStaticClasses	a collection of classes to import as static classes
+	 * @return default imports
+	 */
+	String getDefaultImports(Collection<Class<?>> defaultClasses, Collection<Class<?>> defaultStaticClasses);
```
petebankhead, line 62: I suggest separating into two methods: `getImportStatements(Collection<Class<?>> classes)` and `getStaticImportStatments(Collection<Class<?>> classes)`.
Better to avoid referring to 'default', since the method itself is really independent of however the 'default classes' are defined within QuPath.

qupath-gui-fx/src/main/java/qupath/lib/gui/scripting/ScriptEditorDragDropListener.java
melvingelbard: Good catch! I think adding the list to a `HashSet` as you mentioned should fix this. The next commit will fix this.

PR #1039, opened 2022-08-18
fixed RoiTools.computeTiledROIs

crobbins327:
Noticed that byColumn option did not work (the bottom tiles would never be generated when only using prepared geometries) and that the geometryLocal was not different from geometry (because of typo in the index for row/columnParents).

Made some changes, hope they work.

### DIFF
--- a/qupath-core/src/main/java/qupath/lib/roi/RoiTools.java
+++ b/qupath-core/src/main/java/qupath/lib/roi/RoiTools.java
@@ -742,7 +742,7 @@ public class RoiTools {
 													x,
 													envelope.getMinY(),
 													w + overlap*2,
-													envelope.getMaxX());
+													envelope.getMaxY());
 											if (!prepared2.intersects(col))
 												return empty;
 											else if (prepared2.covers(col))
@@ -765,13 +765,13 @@ public class RoiTools {
 
 			double y = yMin + yi * h - overlap;
 			if (rowParents != null)
-				geometryLocal = rowParents.getOrDefault(y, geometry);
+				geometryLocal = rowParents.getOrDefault(yi, geometry);
 
 			for (int xi = 0; xi < nx; xi++) {
 
 				double x = xMin + xi * w - overlap;
 				if (columnParents != null)
-					geometryLocal = columnParents.getOrDefault(x, geometry);
+					geometryLocal = columnParents.getOrDefault(xi, geometry);
 				
 				if (geometryLocal.isEmpty())
 					continue;


petebankhead:
Thanks, can you describe a bit more how to replicate the original bug and check it is fixed?

We need something we can link to in https://github.com/qupath/qupath/blob/main/CHANGELOG.md to describe the fix (either an issue or a PR).

Also, I think that `settings.gradle` was included here in error.

crobbins327:
Hi Pete,

I stumbled on it when trying to tile very large annotations for a simple plugin I am making. 
 
I basically set the bounds to be the entire image and then had a very large annotation (converted to ROI) that I computed the tiled ROIs for. I don't have a clean example to share, but maybe I can write some code that is commented out and then you can step through it with a debugger and check the behavior out for yourself?

Thank you again for your continued work on QuPath, I've learned so much through your source code.

Also, yes. The settings.gradle file was not supposed to be committed.

petebankhead:
Thanks @crobbins327 could you update the PR (or create a new one) to remove the settings file please?

> I basically set the bounds to be the entire image

Ah, I think that's the key bit - I see the problematic behavior for a full image annotation.

crobbins327:
Ok I reverted the changes and recommitted. Maybe I should just start this PR over...

petebankhead:
Thanks, will try to check it out soon.
Seems I can squash it into a single commit when merging. Never done that, wonder if it works as well as it sounds like it should...

crobbins327:
Thanks for the tip. Haha I need to refactor how I setup the qupath source code....  now I've squashed the commits but I had to ignore the folders for my extensions... sorry for all the complications, maybe you can cherry pick just the "fixed ROITools.computeTiledROIs" commit? I'm still learning git.

petebankhead:
Probably best make a new PR, there shouldn't be changes to `.gitignore` included either :)

Meanwhile, it seems this doesn't fix the bug I thought it was going to fix, i.e. extra tiles are generated along x but not along y:

Screenshot 2022-08-18 at 16 15 08

The objects shown in the screenshot can be reproduced using

```groovy
clearAllObjects()
createSelectAllObject(true)
runPlugin('qupath.lib.algorithms.TilerPlugin', '{"tileSizeMicrons":200.0,"trimToROI":false,"makeAnnotations":true,"removeParentAnnotation":false}')
```

Creating objects outside the image is probably not a good idea anyway, so it's not clear to me whether the x or y behavior is less bad. The inconsistency doesn't seem good... but I wouldn't want to add a 'fix' unless it's totally clear that it's an improvement, since any change will impact the reproducibility of scripts across QuPath versions.

Anyhow, I'm not seeing a change of behavior when using the code from the PR. Could you please fill in the info from the bug report template? It's still not clear to me what the bug is or whether the fix fixes it.

crobbins327:
Yea, that's from recentering the tiles. If you comment out the portion of the code that centers the tiles then I get the behavior that I wanted for my extension.
```
// Center the tiles
// xMin = (int)(bounds.getCenterX() - (nx * w * .5));
// yMin = (int)(bounds.getCenterY() - (ny * h * .5));
```
For a byRow image:
row no center roitools computeTiledROIs

For a byColumn image:
column no center roitools computeTiledROIs

The original fix is mainly to adjust the behavior of the code so that it actually is using the prepared geometries that are calculated. If you step through the code with a debugger, you can see that the rowParents or columnParents are not indexed by y or x but by yi and xi, so the geometries are never retrieved from these maps. And the columnParents geometries are sometimes not correct because the envelope for intersection did not go across the entire column.

The final result depends on if you want the tiled objects to be centered or not within the parent annotation. Perhaps you could fix it by adding an additional "center tiles" argument for the places that need tiling, but I'm sure this would break a lot of things.

I do not have the time to fill out a bug report now, but just wanted to let you know about this.

petebankhead:
Thanks, although I remain quite confused. If you find time to write up a bug report that clearly shows the broken and fixed behavior then that would be really helpful.

Otherwise, I'll be on the lookout for problems with the tiling, but I'm not sure I'll have time to look in much detail any time soon without a demonstration that the current behavior is critically broken. As far as I can tell, any current bugginess may lead to inefficiency but I'd like an example where it gives a wrong result to understand what to fix.

Ideally there would be a test added to [`TestRoiTools`](https://github.com/qupath/qupath/blob/dddad1fe5c567015e07bed8c05940c4403b8d07f/qupath-core/src/test/java/qupath/lib/roi/TestRoiTools.java) that fails before and passes afterwards.

Issue #1032, opened 2022-08-02
Distance to the inner surface of annotations


Svidro:
Calculate a negative distance to the inner surface of annotations rather than saving the distance from the detection to annotation as zero.

As mentioned in several places on the forum, while it is possible to invert an annotation in order to measure the distance to annotations (2D) for detections within a particular annotation, it requires some coding and is a bit awkward. As inverting the annotation no longer inherits the class of the inverted annotation, that also adds an extra step that might be difficult for newer users. 
Finally, large complex annotations may stress some systems to invert, so it would be convenient to avoid this step if possible. 
Current workaround: https://forum.image.sc/t/distance-of-detections-to-border-of-parent-annotation/70199/9?u=research_associate
Some other instances of this coming up
https://forum.image.sc/t/need-help-with-troubleshooting-distance-to-annotation-calculation/48051/4?u=research_associate
https://forum.image.sc/t/qupath-distance-to-annotation-border-from-both-inside-and-outside-of-the-annotation/40643/3?u=research_associate
https://forum.image.sc/t/qupath-distance-between-annotations/47960/7?u=research_associate

Hopefully this could be as simple as replacing the logic that sets inside values to zero with multiplying by negative one, but I'm not sure. https://github.com/qupath/qupath/blob/053efeff6d941e7a73beab5445cf0d6238ed97b7/qupath-core/src/main/java/qupath/lib/analysis/DistanceTools.java#L187
I didn't see anything obvious, so guessing it has to do with how the geometries are being used.

Additionally, it would probably be nice if it "smartly" took into account TMA core borders as well. Though _that_ would probably be a bit more complicated.

petebankhead:
Thanks @Svidro - and also for linking to the code. That reminds me why I didn't do it before... the code is *very* involved. I vaguely remember it going through some rounds of optimization and having to deal with corner cases (like mixtures of lines, areas and points).

I agree it's necessary but may be tricky, not sure where I'll get time to delve into it again.

petebankhead:
This should hopefully address it: https://github.com/qupath/qupath/pull/1042

PR #1042, opened 2022-08-18
Add 'Signed distances to annotations 2D'

petebankhead:
Addresses this (frequent) feature request:
* https://github.com/qupath/qupath/issues/1032

Adds a new *Signed distances to annotations 2D* command, which should be scriptable via `QP` and `DistanceTools`.

distances

distances-signed


MichaelSNelson:
👍

PR #1043, opened 2022-08-18
Fix RoiTools.computeTiledROIs()

petebankhead:
Thanks to @crobbins327 for pointing out the problem and solution.

Slightly modified from proposed code at https://github.com/qupath/qupath/pull/1039

Testing is somewhat convoluted. I originally (wrongly thought) that the *Create tiles* command used this method, but it doesn't. However it's used internally for plugin commands (like cell detection) - so is potentially quite important.

Testing is difficult, but a script like this can help:
```groovy
def annotation = getSelectedObject()
int s = 64
def size = qupath.lib.geom.ImmutableDimension.getInstance(s, s)
def rois = RoiTools.computeTiledROIs(annotation.getROI(), size, size, true, 0)
def newAnnotations = rois.collect {r -> PathObjects.createAnnotationObject(r)}
annotation.clearPathObjects()
annotation.addPathObjects(newAnnotations)
fireHierarchyUpdate()
```

Using a rather convoluted example, and reducing the `geometry.getNumPoints() > 1000` threshold in `RoiTools` to trigger the code, tiles were previously missing:

wrong

With the adapted code they are included:

right

The problem hadn't surfaced before, probably because (as @crobbins327 pointed out) the wrong key was used with the map, and so the 'optimization' of handling row/column geometries didn't do anything.

With that in mind, if this turns out to cause problems then the whole method could just be simplified and the unnecessary attempt at optimization from https://github.com/qupath/qupath/commit/a3366633851740e0d675b118b48133ce61211101 removed.

@crobbins327 could you check if the fix looks ok to you as well?

### DIFF
--- a/qupath-core/src/main/java/qupath/lib/roi/RoiTools.java
+++ b/qupath-core/src/main/java/qupath/lib/roi/RoiTools.java
@@ -717,7 +717,7 @@ public class RoiTools {
 											var row = GeometryTools.createRectangle(
 													envelope.getMinX(),
 													y,
-													envelope.getMaxX(),
+													envelope.getMaxX() - envelope.getMinX(),
 													h + overlap*2);
 											if (!prepared2.intersects(row))
 												return empty;
@@ -742,7 +742,7 @@ public class RoiTools {
 													x,
 													envelope.getMinY(),
 													w + overlap*2,
-													envelope.getMaxX());
+													envelope.getMaxY() - envelope.getMinY());
 											if (!prepared2.intersects(col))
 												return empty;
 											else if (prepared2.covers(col))
@@ -765,16 +765,19 @@ public class RoiTools {
 
 			double y = yMin + yi * h - overlap;
 			if (rowParents != null)
-				geometryLocal = rowParents.getOrDefault(y, geometry);
+				geometryLocal = rowParents.getOrDefault(yi, geometry);
 
 			for (int xi = 0; xi < nx; xi++) {
 
 				double x = xMin + xi * w - overlap;
 				if (columnParents != null)
-					geometryLocal = columnParents.getOrDefault(x, geometry);
+					geometryLocal = columnParents.getOrDefault(xi, geometry);
 				
 				if (geometryLocal.isEmpty())
 					continue;
+				
+//				if (geometry != geometryLocal)
+//					System.err.println("Using row or column geometry!");
 				
 				// Create the tile
 				var rect = GeometryTools.createRectangle(x, y, w + overlap*2, h + overlap*2);


crobbins327:
Yep, this works for me Pete. Clever example.

I also messed around with consolidating the last portion of the code into a parallel stream, but it is not much faster than the original code for one parentROI. In my case, I want to calculate the intersections for multiple ROIs with the tile, so it was easier for me to write the code similar to the snippet below:

```
//		have to make all these "final" temp variables.... probably a better way
		double finalYMin = yMin;
		double finalXMin = xMin;
		boolean finalByColumn = byColumn;
		Map<Integer, Geometry> finalColumnParents = columnParents;
		boolean finalByRow = byRow;
		Map<Integer, Geometry> finalRowParents = rowParents;
		List<ROI> tileROIs = Collections.synchronizedList(new ArrayList<>());
		var plane = parentROI.getImagePlane();
		AtomicInteger nullInterExcepetions = new AtomicInteger(0);
		IntStream.range(0, nx).parallel().forEach(xi -> {
			double x = finalXMin + xi * w - overlap;
//			A very hacky way to consolidate the code into 1 loop.
//			Atomic Reference doesn't behave when getting hit by multiple streams setting potentially different values for each stream...
			Geometry outerGeometryLocal = finalByColumn ? finalColumnParents.getOrDefault(xi, geometry) : geometry;
			IntStream.range(0, ny).parallel().forEach(yi -> {
				double y = finalYMin + yi * h - overlap;
				Geometry geometryLocal = finalByRow ? finalRowParents.getOrDefault(yi, geometry) : outerGeometryLocal;

				// Create the tile
				var rect = GeometryTools.createRectangle(x, y, w + overlap * 2, h + overlap * 2);
				Geometry inter = intersect(rect, geometryLocal);
				if(inter==null) {
					nullInterExcepetions.incrementAndGet();
					return;
				}
				ROI roi = GeometryTools.geometryToROI(inter, plane);
				tileROIs.add(roi);
			});
		});

		if (nullInterExcepetions.get() > 0) {
			logger.warn("Tiles lost during tiling: {}", nullInterExcepetions.get());
			logger.warn("You may be able to avoid tiling errors by calling 'Simplify shape' on any complex annotations first.");
		}
```

crobbins327:
This is kinda related but probably deserves it's own question. But I was wondering if there was a way to show the child objects inside the viewer for Tiles?

Here is a picture of my object hierarchy and current viewer:
children objects of tiles are not shown in viewer

I want all the children objects (which are detections/tiles) to be shown inside each tile:
selected children objects of tiles are not shown in viewer

I tried changing child objects from tiles to detections, but they won't show in the viewer unless I click on each of the children individually. Is there a setting I am missing to show all child objects inside of tiles?

Thank you again for your help.

PR #1049, opened 2022-09-08
Fix IllegalArgumentException for network share URIs on windows

ap--:
Dear QuPath developers,

Quite some time ago a user of `paquo` reported a bug with network share URIs on windows, see: https://github.com/bayer-science-for-a-better-life/paquo/issues/65

This PR get's rid of the immediate issue in ImageServerProvider.getServerBuilders, but I'm not sure if there's more bugs further downstream using network share URIs.

The fix boils down to using `java.nio.Paths.get(uri).toFile()` instead of `java.io.File(uri)` directly. (see: https://github.com/bayer-science-for-a-better-life/paquo/issues/65#issuecomment-970711915)

I am currently unable to test this myself due to lack of access to a windows machine. 

Cheers,
Andreas 😃

### DIFF
--- a/qupath-core/src/main/java/qupath/lib/images/servers/ImageServerProvider.java
+++ b/qupath-core/src/main/java/qupath/lib/images/servers/ImageServerProvider.java
@@ -27,6 +27,7 @@ import java.io.File;
 import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -170,7 +171,7 @@ public class ImageServerProvider {
 
 	private static <T> List<UriImageSupport<T>> getServerBuilders(final Class<T> cls, final String path, String...args) throws IOException {
 		URI uri = legacyPathToURI(path);
-		if ("file".equals(uri.getScheme()) && !new File(uri).exists()) {
+		if ("file".equals(uri.getScheme()) && !Paths.get(uri).toFile().exists()) {
 			throw new IOException(uri.toString() + " does not exist!");
 		}
 
--- a/qupath-core/src/main/java/qupath/lib/images/servers/ImageServerProvider.java
+++ b/qupath-core/src/main/java/qupath/lib/images/servers/ImageServerProvider.java
@@ -4,7 +4,7 @@
  * %%
  * Copyright (C) 2014 - 2016 The Queen's University of Belfast, Northern Ireland
  * Contact: IP Management (ipmanagement@qub.ac.uk)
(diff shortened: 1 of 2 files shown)


qupath-core/src/main/java/qupath/lib/images/servers/ImageServerProvider.java
```diff
@@ -170,7 +171,7 @@ public static URI legacyPathToURI(String path) throws IOException {
 
 	private static <T> List<UriImageSupport<T>> getServerBuilders(final Class<T> cls, final String path, String...args) throws IOException {
 		URI uri = legacyPathToURI(path);
-		if ("file".equals(uri.getScheme()) && !new File(uri).exists()) {
+		if ("file".equals(uri.getScheme()) && !Paths.get(uri).toFile().exists()) {
```
petebankhead, line 174: If you are creating a `File` from a `Path` only to check it exists then I guess using [`Files.exists(path)`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/nio/file/Files.html#exists(java.nio.file.Path,java.nio.file.LinkOption...)) would make more sense?

qupath-core/src/test/java/qupath/lib/images/servers/TestImageServerProvider.java
```diff
@@ -0,0 +1,44 @@
+/*-
+ * #%L
+ * This file is part of QuPath.
+ * %%
+ * Copyright (C) 2018 - 2020 QuPath developers, The University of Edinburgh
```
petebankhead, line 5: Just 2022 here :)

ap--:
> but I guess I don't understand when and why that matters, or if the URI could simply be given in some other way?
> 
> Unfortunately, I use Windows very rarely, network shares on Windows even less, and don't know much about more exotic URIs...

tbh it's the same for me. So far as I understand, it's currently impossible to use valid file URIs on windows that are not mapped to a network drive. i.e. 

```
file://networkshare/c$/file.svs
```

I am not 100% sure if it's possible to use files on network shares, when they are mapped to a network drive, i.e.
```
file:///M:/file.svs
```

And I should actually try this on a windows laptop first to verify. Initially, I was hoping that this might be some issue you're already aware of, and that the source of the bug might be obvious. I'll report back once I've tested this manually with QuPath on windows.

petebankhead:
Ah, thanks. I think the second scenario is the only one I've ever worked with myself - I'll also see if I can replicate it somehow.

ap--:
I think this PR can be closed.
Providing networkshare uris in the form of `"file:////localhost/x$/some/path/file.ext"` seems to work. Thanks again for your help!

https://github.com/bayer-science-for-a-better-life/paquo/blob/832817f6b0a877a57afd8e7e008a287af0564f5e/paquo/tests/test_projects.py#L183-L190

PR #1056, opened 2022-09-13
Update dark.css

finglis:
Update to search box prompt text in dark mode as it was previously poorly visible due to styling from JavaFX modena that needed to be overridden.

Before: 
image
After:
image

Attempts to make the style more specific to override the external style but was unsuccessful so the less preferable !important tag was used. The colour is derived from the current background colour at that time.


petebankhead (commented):
Looks good to me!

Issue #1069, opened 2022-10-05
Object connections cause moderate viewer lag when downsample factor is less than 1


MarkZaidi:
## IMPORTANT!

* [X] I've definitely found a bug (it you're not sure, please use [image.sc](https://forum.image.sc/tags/qupath) instead)
* [X] I've checked https://qupath.github.io for a new release that might already have fixed the issue
* [X] I've checked the [Changelog](https://github.com/qupath/qupath/blob/master/CHANGELOG.md) to see if the bug has already been fixed in the next release
* [X] I've checked for existing GitHub issues describing the same problem

## Bug report

**Describe the bug**
Object connections cause moderate viewer lag when downsample factor is less than 1, evident when panning. Very minor and niche issue, but I figured I'd open a ticket anyways.

**To Reproduce**
1. Run cell detection across a whole-slide annotation
2. Generate object connections via "Delaunay cluster features 2D"
3.  Set viewer downsample factor to 0.99 or less
4. Observe lag when panning. Unchecking "show object connections" or setting a downsample factor greater than or equal to 1 will resolve this issue.

**Expected behavior**
Visualizing objects overlaid onto images shouldn't result in substantial lag

**Screenshots**
Video recording: https://www.youtube.com/watch?v=qOBp2X6HVUc

**Desktop (please complete the following information):**
 - OS: Windows 10
 - QuPath Version: Version: 0.4.0-SNAPSHOT
Build time: 2022-01-24, 13:48

**Additional context**
A 24-class composite classifier, annotation-generating pixel classifier, and several other spatial analyses were also performed.

petebankhead:
I've added a PR that is intended to address this... although its main focus is on another related bug: https://github.com/qupath/qupath/pull/1070

> Visualizing objects overlaid onto images shouldn't result in substantial lag

Ideally it shouldn't, but if it does I wouldn't say it's necessarily a bug... since QuPath is already needing to do a *lot* of stuff to get acceptable performance across a wide range of scenarios. Specifically here:
* For a downsample >= 1, repainting detections caches tiles and multiple resolution levels for performance - this is why QuPath can handle millions of objects.
* For downsample < 1, repainting happens for all detections in the field of view (like for annotations) for improved appearance without nasty bitmap-upsampling artefacts. This is inevitably laggier than using cached tiles, but caching itself has considerable overhead in terms of memory and worse appearance.

[...]

PR #1070, opened 2022-10-05
Improve connection painting performance

petebankhead:
Addresses https://github.com/qupath/qupath/issues/1069 by avoiding drawing each connecting line twice.

Also incorporates image plane information when painting (previously, connections would only be shown if z=0 and t=0).

There is still a problem with connections being lost if both objects are outside the current field of view, but their connecting line crosses the view.


finglis (approved):

petebankhead:
More details about the fix at https://github.com/qupath/qupath/issues/1069#issuecomment-1268167189

PR #1072, opened 2022-10-10
Add ImageServer.readRegion() methods

petebankhead:
This PR proposes a moderately significant change to the `ImageServer` interface, which I'd like to discuss (and sanity-check) with the people most affected.

@iwbh15 @NicoKiaru @lacan @ap-- @sdvillal @finglis (sorry if I have missed anyone...)

The background is that pixels are currently accessed in QuPath via something like
```java
RegionRequest request = RegionRequest.createInstance(server.getPath(), downsample, x, y, width, height);
BufferedImage img = server.readBufferedImage(request);
```

Creating a `RegionRequest` made sense in the early days, since it was used for caching. It remains useful for accessing corresponding regions in paired images, or for creating requests from ROIs, and for avoiding a plethora of parameters (particularly if `z` and `t` indices should be returned as well).

However, often it just feels like unnecessary overhead, e.g. when writing simple scripts... or accessing from another language, like Python.

It is pretty easy to retrofit support for
```java
BufferedImage img = server.readBufferedImage(downsample, x, y, width, height, z, t);
BufferedImage img2 = server.readBufferedImage(downsample, x, y, width, height);
```
by simply adding default methods to the interface. I don't think this should break anything.

However this leads to another consideration... `readBufferedImage` isn't a great name, since the method really returns an instance of whatever the generic parameter `T` stands for in `ImageServer<T>`.

It happens that this is pretty much always `BufferedImage` in QuPath, but the original idea was that an `ImageServer` could exists that returns something else that might be more convenient (e.g. something from ImgLib2, or an OpenCV `Mat`).

So rather than doubling-down on `server.readBufferedImage` I would like to switch to `server.readRegion(RegionRequest)` instead. This feels to me more accurate and intuitive.

The tricky bit is not breaking everything else... like parts of paquo or warpy, and all existing scripts that need pixels.

The approach I've taken here is to mark `readBufferedImage` as deprecated (and planned for removal), while also providing a default implementation that calls `readRegion`. These means that new classes won't need to implement the deprecated method.

Then - and here it gets awkward - I've tried to allow older `ImageServer` implementations to still work by *also* providing a default implementation of `readRegion`. It will throw an `UnsupportedOperationException` *unless* it is called from a subclass that overrides the `readBufferedImage` method... in which case it will fall back to using that. The first time that happens, a warning will be logged.

You can see the code at https://github.com/petebankhead/qupath/blob/server/qupath-core/src/main/java/qupath/lib/images/servers/ImageServer.java#L213

The outcome is that, with the current PR, *all* of these should work in a Groovy script:
```groovy
def server = getCurrentServer()
double downsample = 100.0
def request = RegionRequest.createInstance(server, downsample)
println server.readBufferedImage(request)
println server.readRegion(request)
println server.readRegion(downsample, 0, 0, server.getWidth(), server.getHeight())
println server.readRegion(downsample, 0, 0, server.getWidth(), server.getHeight(), 0, 0)
```
where the `readBufferedImage` call logs a warning *once* (for each `ImageServer` class) and the rest don't.

My tentative plan would be to
* include this in QuPath v0.4.0 (hopefully later this month)
* remove the default implementation of `readRegion` in v0.5.0, forcing any subclass of `ImageServer` to be updated to work... but still allowing scripts to use `readBufferedImage`, with the deprecation message logged
* remove `readBufferedImage` later, e.g. in v0.6.0

As far as I can tell, everything still works in QuPath itself - but I don't know how your projects might be affected.

What do you all think?


lacan:
Hi Pete, and thanks for putting us in the loop!
It is true that the` RegionRequest` bits are a bit verbose, and any ease on that part would be welcome. 

The proposed changes will have little effect on our scripts, as most calls are for `IJTools.createPathImage()` in any case...

Concerning Warpy, and other extensions, we are happy to modify them and to update them no problem, and make sure to freeze the previous versions.

On a personal note, breaking changes let us keep an eye on scripts that are still being used (often for projects that have little to do with the original requirements), so we kind of see it as a good thing when people come to us because something "no longer works". Helps us make sure that not too many weird things are being done with our scripts :) 

I am pinging @rdornier, for completion, as he was working on the OMERO-RAW part we have here and might need to make some changes there, if we keep that extension after the 0.4 update. 

One argument for making more intensive changes, without too much backwards compatibility is that it really forces us to keep up with the new developments, rather than let it linger on until it's too late. So I'm more in favor of "breaking" than keeping full backwards compatibility. Though admittedly, our user base is not as large as yours, so we won't see as much of a tidalwave...

>  include this in QuPath v0.4.0 (hopefully later this month)

Awesome

> remove the default implementation of readRegion in v0.5.0, forcing any subclass of ImageServer to be updated to work... but still allowing scripts to use readBufferedImage, with the deprecation message logged

OK for us, though if the scripts end up breaking, this would not be a problem, kind of a nice extra way of announcing that things have changed.

> remove readBufferedImage later, e.g. in v0.6.0

Not in v0.5.0 already? 

Thanks again for the discussion! Looking forward to notes and comments from others!

petebankhead:
Thanks @lacan ! I'm glad we have a similar view of breaking changes having advantages :)

> > remove readBufferedImage later, e.g. in v0.6.0
> 
> Not in v0.5.0 already?

Could be... certainly tempting if there are no objections!

The awkwardness of the default [`readRegion`](https://github.com/petebankhead/qupath/blob/server/qupath-core/src/main/java/qupath/lib/images/servers/ImageServer.java#L213) method makes me more uncomfortable... that could be skipped entirely in v0.4.0, but then I think that anything else that subclasses `ImageServer` would *definitely* break.

This potentially also affects @j-hudecek and @yli-hallila (should have mentioned you before, sorry!)

ap--:
Hi Pete,

I just ran the `paquo` test suite against the `qupath` snapshot version from this PR and all tests pass. So there are no concerns from our side.

Also, since the current release of `paquo` supports any QuPath from `0.2.0-m9` to (now tested) `0.4.0+snapshot` (https://github.com/qupath/qupath/pull/1072/commits/a91c5bfc865710308dc22f11fd6a3f4709eaf799) having a small breaking change would also be fine. We're maintaining a bunch of checks to support the different QuPath versions anyways.

So what I am saying is: please don't keep any cruft around just because it might break `paquo`.

Cheers,
Andreas

j-hudecek:
Hi Pete, 

Thanks for thinking of Slide Score with breaking changes. Our ```SlideScoreImageServer``` extends ```AbstractTileableImageServer``` and only overrides ```readTile```. So, these changes don't cause any problems if I understand it correctly. I don't see any deprecation warnings at compile- or run-time. The new JDK version, modular JavaFX and mandatory JLS caused more issues ;-)

Jan

petebankhead:
Thanks @ap-- and @j-hudecek :)

The acceptance of this API change has me now thinking about going a step further... I am missing a method like
```java
public T readLevel(int level, ImageRegion region);
```
where the `level` is a specific level of the pyramid (0 for full-resolution).

Currently, if I want to get a rectangle of pixels from level 2, I need to figure out the coordinates in the full-resolution image (i.e. multiply by the downsample value). Then QuPath should find them again after dividing by the downsample.

However, the coordinates need to be integers - which makes me uneasy about what happens if the downsample is something like 4.23452345.
Should I be rounding or flooring when I scale up the coordinates?
And can I be sure that QuPath will do the right thing when it scales them back down, so that I get the original coordinates I wanted again...?

If not, then it seems I might get off-by-one errors and slightly unexpected results. If I want a 256 x 256 pixel region, I might end up with a 255 x 256 pixel region... which would be annoying.

It seems that I can round or floor when scaling up, and then round or floor when scaling down, but I wasn't completely sure which I should be doing so I created a quick Python simulation to test what happens:  https://gist.github.com/petebankhead/2d4a21cb69f3b68c8f8fa14475723647

Based on this, it seems I need to round in both directions... which was maybe a predictable conclusion for the more mathematically confident, but I wasn't sure of it's what QuPath does internally* and we can't count on users necessarily knowing that.

(*However*, note that if the downsample is < 1 then all the methods fail...)

I think it would be nice to resolve the ambiguity somehow, although adding an extra method
```java
public T readLevel(int level, int x, int y, int width, int height, int z, int t);
```
looks like a horrible explosion of ints.

A possibly-simpler alternative might be to support something like this:
```java
public T readScaledRegion(double downsample, int outputX, int outputY, int outputWidth, int outputHeight, int z, int t);
```
where the coordinates are defined according to the output image, but the downsample isn't fixed to *have* to be a pyramidal level.

Thinking about it, that could potentially be implemented more simply with another default method added to the interface that ultimately just called `readRegion`. The main trouble is that I don't know what to call the method.

I'll keep thinking and try to add something to this PR....

*- I think that QuPath is doing the right thing in most places (based on [this](https://github.com/qupath/qupath/blob/053efeff6d941e7a73beab5445cf0d6238ed97b7/qupath-core/src/main/java/qupath/lib/images/servers/TileRequest.java#L142) and [this](https://github.com/qupath/qupath/blob/053efeff6d941e7a73beab5445cf0d6238ed97b7/qupath-core/src/main/java/qupath/lib/images/servers/TileRequest.java#L142)). But I strongly suspect the [`TileExporter`](https://github.com/qupath/qupath/blob/9eaf034e2cd0325d38967bde5ac43900ddc15e3f/qupath-core/src/main/java/qupath/lib/images/writers/TileExporter.java) isn't always doing the right thing.

ap--:
Just as another suggestion:

In our pathology dataset library we implemented functionality that allows you to retrieve a tile at a location with a certain size a **specific MPP** (see [here](https://github.com/Bayer-Group/pado/blob/738ed0f814297f22ba005d5b30d254244e23ef82/pado/images/image.py#L506-L508)).

Because after talking to a few researchers, they usually don't want to think about the downsample values of levels in the file. They want tiles at a specific magnification.

petebankhead:
Thanks @ap-- it's tempting since QuPath generally µm/px values where it can, but the trouble is that we can't count on them always being available. When I know it's there, I end up doing a lot of
```groovy
double downsample = requestedPixelSizeMicrons / server.getPixelCalibration().getAveragedPixelSizeMicrons();
```

petebankhead:
Thanks everyone for the feedback - I'll merge this for now to avoid making too much of a mess while merging other PRs, but might return to it later.

PR #1076, opened 2022-10-14
Improve pixel classifier measurement performance

petebankhead:
Also fixes a possible bug when making measurements from point annotations.
Inspired by this discussion: https://forum.image.sc/t/qupath-measure-pixel-classifier-area-per-cell-detection-for-wsis/72701

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ This is a work-in-progress.
     * Smart parentheses and (double/single) quotes (https://github.com/qupath/qupath/pull/907)
     * Comment block handling (https://github.com/qupath/qupath/pull/908)
   * New 'Edit -> Wrap lines', 'Edit -> Replace curly quotes' and 'Edit -> Zap gremlins' options
-* Making measurements from a pixel classifier should (usually) be much faster
+* Making measurements from a pixel classifier is *much* faster in some circumstances (https://github.com/qupath/qupath/pull/1076)
 * New 'Analyze -> Spatial analysis -> Signed distance to annotations 2D' command (https://github.com/qupath/qupath/issues/1032)
 * Better support for opening/importing from files containing multiple images
   * New 'Show image selector' option when adding images to a project
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ This is a work-in-progress.
     * Smart parentheses and (double/single) quotes (https://github.com/qupath/qupath/pull/907)
     * Comment block handling (https://github.com/qupath/qupath/pull/908)
   * New 'Edit -> Wrap lines', 'Edit -> Replace curly quotes' and 'Edit -> Zap gremlins' options
+* Making measurements from a pixel classifier should (usually) be much faster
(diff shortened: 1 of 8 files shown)


petebankhead:
I tested performance using CMU-1.svs.
I used a very basic thresholder and simple classifier trained for 3 classes, saved for both classification and probability output - then ran the script at the bottom.

The processing time was:

| v0.3.0  | v0.4.0-SNAPSHOT |
| ------------- | ------------- |
| 593.9 s  | 60.1 s  |

Results identical as far as I can tell. So... quite a substantial difference :)

```groovy
def checkpoints = [:]


setImageType('BRIGHTFIELD_H_E')
setColorDeconvolutionStains('{"Name" : "H&E default", "Stain 1" : "Hematoxylin", "Values 1" : "0.65111 0.70119 0.29049", "Stain 2" : "Eosin", "Values 2" : "0.2159 0.8012 0.5581", "Background" : " 255 255 255"}')

clearAllObjects()

checkpoints << ['Tissue detection': System.currentTimeMillis()]

createAnnotationsFromPixelClassifier("Tissue detection", 10000.0, 0.0, "INCLUDE_IGNORED")

checkpoints << ['Cell detection': System.currentTimeMillis()]

selectAnnotations()
runPlugin('qupath.imagej.detect.cells.WatershedCellDetection', '{"detectionImageBrightfield": "Hematoxylin OD",  "requestedPixelSizeMicrons": 1.0,  "backgroundRadiusMicrons": 8.0,  "medianRadiusMicrons": 0.0,  "sigmaMicrons": 1.5,  "minAreaMicrons": 10.0,  "maxAreaMicrons": 400.0,  "threshold": 0.1,  "maxBackground": 2.0,  "watershedPostProcess": true,  "cellExpansionMicrons": 5.0,  "includeNuclei": true,  "smoothBoundaries": true,  "makeMeasurements": true}')

for (classifier in ['Some probability', 'Some classification']) {

    // Create annotation measurements
    checkpoints << ["Annotation measurements for $classifier": System.currentTimeMillis()]
    selectAnnotations()
    addPixelClassifierMeasurements(classifier, classifier)
    
    // Create cell measurements
    checkpoints << ["Cell measurements for $classifier": System.currentTimeMillis()]
    selectCells()
    addPixelClassifierMeasurements(classifier, classifier)
}
checkpoints << ["Done": System.currentTimeMillis()]
resetSelection()
println 'Done!'

def entries = checkpoints.entrySet() as List
println "Total time: \t${entries[-1].value - entries[0].value} ms"
for (int i = 0; i < entries.size()-1; i++) {
    println "    ${entries[i].key} \t${entries[i+1].value - entries[i].value}"
}
```

petebankhead:
Latest commit adds more options to restrict where live pixel classifier prediction is calculated.

Previously, it could be restricted to annotations - but using their full bounding box. This could sometimes still result in very large regions being processed.

annotations_bounds

Now it's also possible to restrict using the annotation ROI directly (i.e. the ROI shape intersects the tiled region that may be processed). This can reduce the amount of processing required substantially in some cases.

annotations_only

Both options still exist, since the more complex calculations to restrict the predicted regions could *potentially* slow things down in some cases.

Issue #1088, opened 2022-10-22
Add Show/hide names toolbar button


petebankhead:
## Feature request

**Is your feature request related to a problem? Please describe.**
When PathObjects have names, it's not at all obvious to a user how to show/hide them in the viewer - since the option is hidden under *View &rarr; Show names*.

**Describe the solution you'd like**
Add another toolbar button, probably next to *Show/hide annotations*.

**Describe alternatives you've considered**

* Just document it better and hope people check readthedocs
* Add it via a less obtrusive 'more' button next to the show/hide annotations toolbar button

**Additional context**
One way or another, it needs to be easier to control name visibility. This is likely to become particularly relevant if ImageJ ROIs are imported from ROI Manager.

PR #1090, opened 2022-10-23
New 'Measure > Grid views' commands

petebankhead:
Based on the old 'TMA &rarr; TMA grid summary view', but now with support for annotations.

Also measurement table improvements:
* improved thumbnail painting
* moved object ID to the last column, with new preference to turn it off completely


petebankhead:
Added 'Show names' toolbar button (with an `N` icon), to fix
* https://github.com/qupath/qupath/issues/1088

Issue #1085, opened 2022-10-22
Replace MeasurementList with Map for more intuitive scripting


petebankhead:
### Is your feature request related to a problem? Please describe.

Most measurements for a `PathObject` are managed in a `MeasurementList`. This effectively maps between `String` keys and primitive `double` values, and can work in a reasonably efficient way by:

* Supporting both `double[]` as `float[]` as backing arrays
* Sharing the list of `String` measurement names between many objects
* Storing a map between `String` names and array indexes

The last two optimisations only come into effect whenever `MeasurementList.close()` is called, but all together can help considerably when working with millions of objects. A less efficient alternative implementation can be used with smaller numbers of objects.

However, there are a few problems:

* Originally envisaged as a list, it's possible to have duplicates... but then what happens when requesting a measurement by name is undefined
  * To overcome the above issue, it's advised to use `putMeasurement(String, double)` rather than `addMeasurement(String, double)`... but this isn't enforced
* Consumers need to remember to close the list at the end... but can often forget
  * The lists now implement `AutoCloseable` so can be used with try-with-resources... but it looks weird
* If a measurement is unavailable, `Double.NaN` is returned... but is indistinguishable from that being the *correct* value
* It's a whole non-standard API to learn when scripting

It's all pretty cumbersome, as can be seen in a Groovy script:

```groovy
def allCells = getCellObjects()

// Get an individual measurement for the first cell
def cell = allCells[0]
println cell.getMeasurementList().getMeasurementValue('Nucleus: Area')

// Get all the measurement names and values
println cell.getMeasurementList().getMeasurementNames()
def values = []
for (int i = 0; i < cell.getMeasurementList().size(); i++) {
    values << cell.getMeasurementList().getMeasurementValue(i)
}
println values
```

### Describe the solution you'd like

*Ideally* we'd have a `Map<String, double>` that preserves the efficiency of `MeasurementList`... if only this was permitted by Java.

In practice, `Map<String, Double>` is likely to give adequate performance in Groovy - as long as the values are stored as primitives for efficiency in Java.

In fact, to maintain compatibility we could initially implement an `AbstractMap` that wraps around an existing `MeasurementList` - with the latter still used internally. We'd need a method added to `PathObject`

```java
public Map<String, Double> getMeasurements();
```

This would help by:

* Providing a standard API, which explicitly doesn't permit duplicate keys
* Fitting well with GeoJSON export
* Permitting `MeasurementList` to be deprecated
[...]

Issue #1087, opened 2022-10-22
Support setting object classifications by name or names


petebankhead:
### Is your feature request related to a problem? Please describe.

Classifications are represented in QuPath by `PathClass` objects. Therefore if you want to set a classification you need something like:

```java
pathObject.setPathClass(PathClassFactory.getPathClass("Some name", optionalColorInteger));
```

In a script, with default static imports from `QP`, that can be somewhat shortened to either of the following
```groovy
pathObject.setPathClass(getPathClass("Some name"))
pathObject.pathClass = getPathClass("Some name") // Alternative using some Groovy magic
```

### Describe the solution you'd like

Adding a method to `PathObject`
```java
public void setPathClass(String name) {
    setPathClass(PathClassFactory.getPathClass(name);
}
```
would make it possible to use the simpler script

```groovy
pathObject.pathClass = "Some name"
```

I feel this further lowers the barrier to scripting... except for one thing: it makes `pathObject.setPathClass(null)` ambiguous - and liable to throw an exception.

This could be slightly mitigated in several ways, e.g.

1. adding a constant field to `QP`
```java
PathClass UNCLASSIFIED
```
2. adding a `resetPathClass()` method as an alternative to setting null
3. taking an empty string as an alternative way to set the classification to null
4. using different names, e.g. `setPathClass(PathClass)` and `setClassification(String)` - but both ultimately causing the same `PathClass` to be set.

The final option has the problem of potential confusion by seeming to have two variables, but in reality only having one.

However there might be an intriguing opportunity in this... although it will take a bit more thought.

One criticism of QuPath's classification model is that it is [restricted to a single `PathClass`, even if further subclassifications can be derived](https://qupath.readthedocs.io/en/stable/docs/concepts/classifications.html).

The API could be adapted to something like:

```java
public void setPathClass(PathClass);

public PathClass getPathClass();

public void setClassifications(Collection<String>);

public Set<String> getClassifications();
```

Internally, `get/setClassifications()` would be working with the `PathClass`, but from a user perspective they seem only to have an ordered set. This would mean they could switch between using either the `PathClass` or string-based classification model easily, depending upon which makes most sense for the application. For example:

* `PathClass` may be helpful in dealing with single classifications, possibly with intensity-subclasses (e.g. scoring a single IHC marker). Working with more than one subclassification is probably uncommon.
* `Set<String>` makes sense for multiplexed analysis, where each 'classification' could be seen as being like a tag that a cell is positive for a particular marker.

[...]

PR #1094, opened 2022-10-24
Improve working with measurements and classifications

petebankhead:
See https://github.com/qupath/qupath/issues/1085 for more details

This involved deprecating `MeasurementList.addMeasurement(String, double)` to enforce uniqueness (so the list was really always more of a map anyway). This is important for interpreting measurements anyway.

### DIFF
--- a/qupath-core-processing/src/main/java/qupath/imagej/detect/cells/ObjectMeasurements.java
+++ b/qupath-core-processing/src/main/java/qupath/imagej/detect/cells/ObjectMeasurements.java
@@ -50,11 +50,11 @@ class ObjectMeasurements {
 	 * @param prefix
 	 */
 	public static void addShapeStatistics(MeasurementList measurementList, PolygonROI roi, double pixelWidth, double pixelHeight, String prefix) {
-		measurementList.putMeasurement(prefix + "Area", roi.getScaledArea(pixelWidth, pixelHeight));
-		measurementList.putMeasurement(prefix + "Perimeter", roi.getScaledLength(pixelWidth, pixelHeight));
-		measurementList.putMeasurement(prefix + "Circularity", RoiTools.getCircularity(roi, pixelWidth, pixelHeight));
+		measurementList.put(prefix + "Area", roi.getScaledArea(pixelWidth, pixelHeight));
+		measurementList.put(prefix + "Perimeter", roi.getScaledLength(pixelWidth, pixelHeight));
+		measurementList.put(prefix + "Circularity", RoiTools.getCircularity(roi, pixelWidth, pixelHeight));
 //		measurementList.putMeasurement(prefix + "Convex area", roi.getScaledConvexArea(pixelWidth, pixelHeight));
-		measurementList.putMeasurement(prefix + "Solidity", roi.getSolidity());
+		measurementList.put(prefix + "Solidity", roi.getSolidity());
 	}
 
 	public static void addShapeStatistics(MeasurementList measurementList, Roi roi, ImageProcessor ip, Calibration cal, String prefix) {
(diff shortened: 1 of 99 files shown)


finglis (approved):

petebankhead:
Continued to work on this, and now the PR addresses both

* https://github.com/qupath/qupath/issues/1085
* https://github.com/qupath/qupath/issues/1087

**Please see the links for more detail. It's a fairly substantial change, so feedback would be very welcome!**

To see it in action, here is a script for multiplex classification - written for the famous [LuCa image](https://qupath.readthedocs.io/en/stable/docs/intro/acknowledgements.html) after running cell detection.

The choice of threshold is dodgy, but the point is to show how this PR changes the way the scripter can interact with measurements and classifications, using a (I think) far simpler and more intuitive syntax, and fewer lines of code.

```groovy
// Get the server & cells
def server = getCurrentServer()
def cells = getCellObjects()

// Reset our classifications
cells.each { cell -> cell.resetPathClass() }

// Loop through channels
for (def channel in server.getMetadata().getChannels()) {
    // Extract the channel name
    def channelName = channel.name
    
    // Skip some channels
    if ('DAPI' in channelName || 'Autofluorescence' in channelName)
        continue
        
    // Create a classification name from the channel
    // Here, I take the first bit up until any whitespace
    def classificationName = channelName.split()[0]

    // Define the measurement we want
    def measurementName = "Cell: $channelName mean"
    
    // Calculate some threshold from the measurement
    // Here, just the mean
    double threshold = cells.measurements[measurementName].average()
    
    // Append a classification to all the cells above the threshold
    cells.each { cell ->
        if (cell.measurements[measurementName] > threshold)
            cell.classifications += [classificationName]
    }
}

// Figure update (could do this automatically...)
fireHierarchyUpdate()
```

Closer inspection reveals a few Groovy tricks at work:

* `List.each { }` as a shorter alternative to for loops
* Avoid explicitly calling getters (e.g. `channel.name` rather than `channel.getName()`, `cell.measurements` rather than `cell.getMeasurements()` or - previously in QuPath - `cell.getMeasurementList()`
* Getting from a map using `map[key]` syntax (rather than `map.get(key)`
* Calling a method on a list to get it applied to all elements... so `cells.measurements[measurementName].average()` first extracts all the measurements, then averages them
* Sneakily applying a `setClassifications()` again by accessing it like a property - and using collection concatenation with `=+` to join the old classifications to the new ones
  * this works in the opposite direction, with `-=` to remove one or more classifications

The changes have been made without introducing any changes to QuPath's data files, since `PathClass` and `MeasurementList` are still doing the main work, it's just that there are now alternative ways to interact with them. They are designed to map well with Groovy tricks, while still behaving sensibly from Java.

There are a few outstanding questions to be resolved, including:
* how much should classification names be validated?
* should duplicate classification names be permitted?

Duplicate names are currently allowed, but can also easily be restricted, e.g. see
```groovy
// Add the same class multiple times
def p = getSelectedObject()
for (int i = 0; i < 10; i++) {
    p.classifications += ['Class 1']
}
println p.classifications

// Calling this enforces uniqueness
p.classifications = p.classifications as Set
println p.classifications

fireHierarchyUpdate()
```
which prints
```
INFO: [Class 1, Class 1, Class 1, Class 1, Class 1, Class 1, Class 1, Class 1, Class 1, Class 1, Class 1]
INFO: [Class 1]
```

**If anyone wants to try the PR, I'd appreciate feedback - or any bugs or things I've failed to think about**

finglis (approved):

petebankhead:
The latest commit deprecates `PathClassFactory`. To get a `PathClass`, now use one of the static methods in `PathClass` itself.

Issue #1083, opened 2022-10-21
Colors of the table in TMA Data Viewer do not match with "Moderna Dark" theme.


ajay1685:
## Bug report

**Describe the bug**
1. TMA Data viewer table colors does not match with the "Moderna Dark" theme
2. TMA Data viewer table has missing values for column Object ID, Name, Missing, Centroid X, and Centroid Y
3. Vertical scrolling for the TMA Data Viewer is sluggish (only when loading from the entire project which has a lot of detections) 

**To Reproduce**
1. Open a project with TMA images, open an image with TMA Data
4. Click on File -> "TMA Data" -> "Launch TMA Data Viewer"
5. In the TMA Result Viewer window unchecking the "Group by ID" results in a table with light background and gray text. See the screenshot
6. Values from the following columns are missing:  Object ID, Name, Missing, Centroid X, and Centroid Y
7. Test vertical scrolling in the TMA Data Viewer

**Expected behavior**
1. TMA Data viewer colors should match "Moderna Dark" theme
2. Populate values for the following columns: Object ID, Name, Missing, Centroid X, and Centroid Y
3. Improve table for smooth vertical scrolling - if possible!

**Screenshots**
<p align="center">
<img src="https://user-images.githubusercontent.com/10900214/197264041-a4c298d4-b3b3-43e8-a142-9d8d9b8ff47a.png" width="800">
</p>

**Desktop (please complete the following information):**
 - OS: Windows 10
 - QuPath Version: 0.3.2 & 0.4.0-SNAPSHOT

petebankhead:
Thanks @ajay1685 the PR above should address most of this, and a couple of other bugs I spotted (like measurements not updating if cores were moved on the image). Please let me know if you notice any significant problems with this.

Because the TMA data viewer is unmaintained, I think there is an above-average risk of unreported/unnoticed bugs so I've added a warning now as well. You can double-click on it to make it go away.

<img width="1006" alt="TMA data viewer" src="https://user-images.githubusercontent.com/4690904/197327977-4298edd4-13f3-41d3-b99c-6f95be821ff9.png">

In the longer term, I think it should be removed or rewritten.

Rewritten would be preferable, but it's really a matter of capacity... as the software gets bigger, there are so many different parts to maintain and I haven't worked with TMAs myself in years.
(In that regard, new QuPath jobs to be advertised next week - please share them widely!)

[...]

PR #1096, opened 2022-10-25
Fixes: columnImage is null

ajay1685:
This seems to address the ```"columnImage" is null``` error mentioned in #1083 when loading the qptma file in the TMA Data Viewer.

### DIFF
--- a/qupath-gui-fx/src/main/java/qupath/lib/gui/tma/TMASummaryViewer.java
+++ b/qupath-gui-fx/src/main/java/qupath/lib/gui/tma/TMASummaryViewer.java
@@ -1460,9 +1460,9 @@ public class TMASummaryViewer {
 				columnOverlay.widthProperty().addListener((v, o, n) -> {
 					if (n.doubleValue() == columnOverlay.getPrefWidth())
 						return;
-					columnImage.setPrefWidth(n.doubleValue());
 					if (hasImages)
-						table.refresh();
+						columnImage.setPrefWidth(n.doubleValue());
+					table.refresh();
 				});
 				columns.add(columnOverlay);
 			}
--- a/qupath-gui-fx/src/main/java/qupath/lib/gui/tma/TMASummaryViewer.java
+++ b/qupath-gui-fx/src/main/java/qupath/lib/gui/tma/TMASummaryViewer.java
@@ -1441,10 +1441,9 @@ public class TMASummaryViewer {
 				columnImage.widthProperty().addListener((v, o, n) -> {
 					if (n.doubleValue() == columnImage.getPrefWidth())
 						return;
-					if (hasOverlay){
+					if (hasOverlay)
 						columnOverlay.setPrefWidth(n.doubleValue());
-						table.refresh();
-					}
+					table.refresh();
 				});
 				columns.add(columnImage);
 			}
@@ -1461,10 +1460,9 @@ public class TMASummaryViewer {
 				columnOverlay.widthProperty().addListener((v, o, n) -> {
 					if (n.doubleValue() == columnOverlay.getPrefWidth())
 						return;
-					if (hasImages){
+					if (hasImages)
 						columnImage.setPrefWidth(n.doubleValue());
-						table.refresh();
-					}
+					table.refresh();
 				});
 				columns.add(columnOverlay);
 			}
--- a/qupath-gui-fx/src/main/java/qupath/lib/gui/tma/TMASummaryViewer.java
+++ b/qupath-gui-fx/src/main/java/qupath/lib/gui/tma/TMASummaryViewer.java
@@ -1441,9 +1441,10 @@ public class TMASummaryViewer {
 				columnImage.widthProperty().addListener((v, o, n) -> {
 					if (n.doubleValue() == columnImage.getPrefWidth())
 						return;
-					if (hasOverlay)
+					if (hasOverlay){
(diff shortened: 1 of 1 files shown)


petebankhead:
Thanks! Not certain because I haven't run it, but looks like it should be
```java
if (hasImages) {
    columnImage.setPrefWidth(n.doubleValue());
    table.refresh();
}
```
since likely no need to call refresh unless there are images. Does that work?

ajay1685:
I tested your suggestion and included the fix. Now both column ```columnImage``` & ```columnOverlay``` sizes are in sync. Adjusting one also adjusts the other proportionally. I believe this is the correct behavior.

ajay1685:
The commit [f37dcfc](https://github.com/qupath/qupath/pull/1096/commits/f37dcfc63cc65a43b8a0eba76e7b7db86c3d75bc) assumes both the ```columnImage``` & ```columnOverlay``` are available and works well when it is true. Table does not seem to refresh when only one of them is available. 
Hence the table may need to be refreshed outside the ```if``` statement. Does this mean that the table may refresh twice when both column are available ? 
This one is trickier that I had initially thought.
 
```java
		// Add columns to show images, if we have them
		if (hasImages || hasOverlay) {
			TreeTableColumn<TMAEntry, TMAEntry> columnImage = hasImages ? new TreeTableColumn<>("Thumbnail") : null;
			TreeTableColumn<TMAEntry, TMAEntry> columnOverlay = hasOverlay ? new TreeTableColumn<>("Overlay") : null;

			if (hasImages) {
				columnImage.setCellValueFactory(new Callback<CellDataFeatures<TMAEntry, TMAEntry>, ObservableValue<TMAEntry>>() {
					@Override
					public ObservableValue<TMAEntry> call(CellDataFeatures<TMAEntry, TMAEntry> p) {
						return p.getValue().valueProperty();
					}
				});
				columnImage.setCellFactory(c -> new ImageTableCell(imageCache, false));
				columnImage.maxWidthProperty().bind(maxSmallWidth);
				columnImage.widthProperty().addListener((v, o, n) -> {
					if (n.doubleValue() == columnImage.getPrefWidth())
						return;
					if (hasOverlay)
						columnOverlay.setPrefWidth(n.doubleValue());
					table.refresh();
				});
				columns.add(columnImage);
			}

			if (hasOverlay) {
				columnOverlay.setCellValueFactory(new Callback<CellDataFeatures<TMAEntry, TMAEntry>, ObservableValue<TMAEntry>>() {
					@Override
					public ObservableValue<TMAEntry> call(CellDataFeatures<TMAEntry, TMAEntry> p) {
						return p.getValue().valueProperty();
					}
				});
				columnOverlay.setCellFactory(c -> new ImageTableCell(imageCache, true));
				columnOverlay.maxWidthProperty().bind(maxSmallWidth);
				columnOverlay.widthProperty().addListener((v, o, n) -> {
					if (n.doubleValue() == columnOverlay.getPrefWidth())
						return;
					if (hasImages)
						columnImage.setPrefWidth(n.doubleValue());
					table.refresh();
				});
				columns.add(columnOverlay);
			}
		}
```

This might be better for now! Since it works in both cases:
1. If both ```columnImage``` and ```columnOverlay``` are available.
2. When one of the ```columnImage``` or ```columnOverlay``` is not available.

I discovered this because when TMA Data is exported for the multichannel image (non-RGB) it is exported as TIF. The TMA viewer only seems to load the image if it is saved as .jpg. For the overlay it load either .jpg or .png.

Here is TMA Data exporter code that exports the images:
https://github.com/petebankhead/qupath/blob/fca7fa5fd9efd815f4c317705fbe2a0f9c39270f/qupath-gui-fx/src/main/java/qupath/lib/gui/tma/TMADataIO.java#L107
```jave
	String coreExt = imageData.getServer().isRGB() ? ".jpg" : ".tif";
```

and Here is TMA Data Viewer code that loads the images:
https://github.com/petebankhead/qupath/blob/fca7fa5fd9efd815f4c317705fbe2a0f9c39270f/qupath-gui-fx/src/main/java/qupath/lib/gui/tma/TMASummaryViewer.java#L1714
```
	File fileImage = new File(dirData, name + ".jpg");
```

petebankhead:
Refreshing slightly more than necessary is probably ok, it's only a problem if you notice it slowing things down. Certainly better than missing a necessary `refresh()` call :)

There's no .tif support because then that brings in channel colors/brightness & contrast and so on, so I guess that would require a more core change to export with those settings. But that gets even more awkward if the visualization settings might not be available if calling the export by batch processing...

petebankhead:
Incidentally, if you're enjoying it and want to tackle the sluggishness you could try setting the cell height to be fixed - although it's tricky because it will depend upon the image size. So you'll need to bring to some other property, and make sure it never gets too small (or fails horribly if images are missing).

I introduced that recently in the regular measurement tables and it improved performance quite a bit. The code I used is at https://github.com/qupath/qupath/blob/main/qupath-gui-fx/src/main/java/qupath/lib/gui/commands/SummaryMeasurementTableCommand.java#L210

See here for more info: https://openjfx.io/javadoc/12/javafx.controls/javafx/scene/control/TableView.html#fixedCellSizeProperty()

(Not necessary though - just if you're curious!)

ajay1685:
I think this will do. let's not make this pull request any messier than it already is. Any new issues will need to be a separate issue with it's own PR :)

PR #1145, opened 2022-11-17
In recent projects menu, added paths tooltips and fixed OS dependent separators

zindy:
Hello,

this PR is to make it easier to recognise which QuPath project is which in the "Recent projects..." menu:

image

If you let your mouse pointer sit over one of the shorten project paths, a tooltip appears with the full project path.

Also, since I'm on Windows, I changed the hard-coded slashes to `File.separator` so they become OS dependent (i.e. '/' on Linux or Mac, and '\\' on Windows).

Cheers,
Egor

### DIFF
--- a/qupath-gui-fx/src/main/java/qupath/lib/gui/tools/GuiTools.java
+++ b/qupath-gui-fx/src/main/java/qupath/lib/gui/tools/GuiTools.java
@@ -1535,7 +1535,6 @@ public class GuiTools {
 					continue;
 				String name = getNameFromURI(uri);
 				name = "..." + File.separator + name;
-				//MenuItem item = new MenuItem(name);
 				CustomMenuItem item = new CustomMenuItem(new Label(name));
 				Tooltip tooltip = new Tooltip(Paths.get(uri).normalize().toString());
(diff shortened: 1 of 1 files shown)


petebankhead:
Hi @zindy I'm afraid this can't be included, as it breaks the menus on the Mac. In fact I've had to stay away from `CustomMenuItem` almost entirely for that reason (it's ok in popup menus).

The problem is that QuPath uses the system menubar where possible. You can change this in the preferences, but it's on by default and that means that the app feels a lot more like a native one; on the Mac it means the menubar at the top, rather than anchored in the window.

The downside on the Mac is that icons don't appear, and `CustomMenuItem` isn't supported. When I try this PR it causes the text of the recent menu items to disappear entirely. It reappears if I turn off the system menubar option... but then blocks me from ever turning it on again.

I can't check it on Linux right now, but I suspect there might also be issues.

Re. the separator, I can see an argument for changing it - but I'd tend towards leaving it as it is. My thinking is that it really represents the end part of the URI, so the slashes should be forward. Additionally, I'd encourage the use of forward slashes in paths for QuPath scripts - even when written in Windows - since they work consistently, while users need to know to escape backslashes.

PR #1147, opened 2022-11-18
Middle button

petebankhead:
@zindy what do you think?

Reduced middle button

Possibly it's just middle button envy from a Mac user, but I was uncomfortable with giving it equal prominence in the input display (although I think it's a good idea to have it represented).


zindy:
Looks really nice @petebankhead!

Definitely looks like the wheel button on a modern mouse, rather than the middle button on a Sun-3 workstation mouse from the 80s!

PR #1222, opened 2023-02-01
Show ImageLocation Panel on Max Border Pixels

matty234:
At the moment, the ImageLocation overlay panel doesn't show when the cursor is at the maximum x or y values:

image

This pull request makes it visible:

image


petebankhead:
Thanks!

PR #1270, opened 2023-06-21
Further Brightness/Contrast improvements

petebankhead:
Summary of changes in this PR (and recent ones):
* Add checkbox to the top of the 'Show' column to show/hide all
* Replace color rectangles with minimal color pickers (that look the same, but respond to clicks)
* Automatically bring text field into focus when double-clicking to set a channel name
* Rename columns
* Reorder the lower controls
* Support regular expressions in the channel name filter for advanced filtering (optional)

Before (left) and after (right).

Brightness:Contrast before changes

Brightness:Contrast dialog updated


petebankhead:
The "Pixel value" label has since been removed as well, to avoid wasting more vertical space.

Issue #1272, opened 2023-06-22
Make SVG export include overlays, not just the original image


zindy:
## Feature request

**Is your feature request related to a problem? Please describe.**
At the moment, SVG exports [1] behave differently from their bitmap counterparts [2], in the sense that the bitmap exports include whichever overlays are currently displayed, but the SVG exports only include the original image.

[1] File -> Export images -> Rendered SVG and File -> Export snapshot -> Current viewer content (SVG)
[2] File -> Export images -> Rendered RGB (with overlays), File -> Export snapshot -> Main window content and File -> Export snapshot -> Current viewer content

**Describe the solution you'd like**
I would *both* types of exports to include overlays if the user so wishes.

See the discussion on https://forum.image.sc/t/export-snapshot-current-view-content-svg-does-not-include-the-density-map/82740

**Describe alternatives you've considered**
I have a two-step workaround in the meantime:
1. Export images -> Rendered SVG (Linked raster)
2. Hide the annotations and detections, then Export images -> Rendered RGB to overwrite the linked raster image, using PNG and the same downsample as used in step 1.

**Additional context**
SVG is a great format. The files can be edited in [Inkscape](https://inkscape.org/) and also be used as-is in (recent versions of) Microsoft Office Word and Powerpoint. And they look great when printed, which is why I think it's important for QuPath to have the best possible SVG export functions.

MichaelSNelson:
I would like to second this, and add some references that QUAREP-LiMi is pushing for! Make good quality exports easy for users, and we can hope/expect better quality figures in publications:
https://arxiv.org/abs/2302.07005
Work in progress video from @biovoxxel https://my.hidrive.com/lnk/Mfimluif#file where the latter parts discuss creation of SVG image overlays and their advantages in publication.

biovoxxel:
I have to add that this is the first and inofficial version of the video but hope it helps. The location might soon change and I will post it here. I am also happy to help in case the QuPath team has some questions on the implementation. To have comparable and interchangeable SVGs at the end I would plea for embedding the images as a base64 png in the SVG.
Here is the ImageJ implementation: https://github.com/biovoxxel/BioVoxxel-Figure-Tools

petebankhead:
Thanks, I've merged some changes into the `dev` branch for v0.5.0.

Basically, SVG export will now add the active overlay layers. It does this using either a vector representation (e.g. for objects or gridlines) or separate images with opacity set according to the overlay display. This means that individual layers should remain editable later.

[...]

PR #1273, opened 2023-06-23
Add overlay support in SVG export

petebankhead:
See https://github.com/qupath/qupath/issues/1272

This PR:
* Includes visible overlays in SVG export, as long as images are included
* Attempts to render vector overlays as such (e.g. grid & TMA overlays)
* Makes the SVG export parameters persistent


petebankhead:
I should add that it **doesn't** handle scalebars etc. as these are rendered in a completely different way, which isn't very amenable to the current approach to SVG export. But that's probably not easily fixable without changing a lot of stuff.

PR #1275, opened 2023-07-05
make getNameWithoutExtension return value consistent

hubutui:
`GeneralTools.getNameWithoutExtension(String)` returns the whole string without the extension for absolute path, while
`GeneralTools.getNameWithoutExtension(File)` returns only the file name part without the extension. eg: `/home/demo/myfile.tif`, the former method returns `/home/demo/myfile` while the latter one returns `myfile`. This commit make them return `myfile` for consistent. A test is also added.

### DIFF
--- a/qupath-core/src/main/java/qupath/lib/common/GeneralTools.java
+++ b/qupath-core/src/main/java/qupath/lib/common/GeneralTools.java
@@ -296,8 +296,7 @@ public final class GeneralTools {
 	 * {@link #getExtension(File)}
 	 */
 	public static String getNameWithoutExtension(String name) {
-		var ext = getExtension(name).orElse(null);
-		return ext ==  null ? name : name.substring(0, name.length() - ext.length());
+		return getNameWithoutExtension(new File(name));
 	}
 	
 	/**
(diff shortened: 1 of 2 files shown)


petebankhead:
I think this is an issue of naming/documentation, not behavior.

If the caller knows that their string can be converted to a `File`, then it's very easy for them to just call `new File(String)` first and use `getNameWithoutExtension(File)`. But the current behavior is consistent and useful for cases where this isn't the desired result.

Although the naming isn't great, `getNameWithoutExtension(String)` does not specify the name to be a *file*name. Although it's natural for the user to think it would be, I don't think the method should silently assume that.

Also, QuPath can work with images where the URL doesn't relate to a file system file, or this method could be called with a String that contains characters that are invalid for a filename. I don't *think* `new File(String)` throws an exception in that case, but I'm not certain and behavior might be platform-dependent.

For these reasons I would prefer to either
* Rename the method to `stripExtension(String)` (but deprecate the original for removal in the future)
* Keep the original but improve the javadoc

(There is also a major refactoring currently under way, so I'm reluctant to merge other PRs in the short term to avoid conflicts.)

hubutui:
Yeah, people would expect these two methods return the same result. It would also be a great choice to rename `GeneralTools.getNameWithoutExtension(String)` to another proper one and deprecate it.

petebankhead:
@hubutui I've made a proposed fix at https://github.com/qupath/qupath/pull/1276 - what do you think?

PR #1279, opened 2023-07-17
Add changelistener to downsamplefactor

alanocallaghan:
This probably only come in handy in weird edge cases like clicking the "adjust zoom to fit images to viewer" button when zoomed in.

Maybe a binding would be more suitable, though

### DIFF
--- a/qupath-gui-fx/src/main/java/qupath/lib/gui/ToolBarComponent.java
+++ b/qupath-gui-fx/src/main/java/qupath/lib/gui/ToolBarComponent.java
@@ -309,6 +309,9 @@ class ToolBarComponent {
 			setMaxWidth(60);
 			setTextAlignment(TextAlignment.CENTER);
 			setOnMouseEntered(e -> refreshMagnificationTooltip());
+			viewer.downsampleFactor().addListener((v, o, n) -> {
+				setText(GuiTools.getMagnificationString(viewer));
+			});
 			setOnMouseClicked(e -> {
 				if (e.getClickCount() == 2)
 					promptToUpdateMagnification();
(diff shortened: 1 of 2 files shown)


petebankhead:
When trying to run this I get
```
qupath.fx.dialogs.Dialogs - QuPath exception: Exception in Application start method
java.lang.RuntimeException: Exception in Application start method
	at com.sun.javafx.application.LauncherImpl.launchApplication1(LauncherImpl.java:893)
	at com.sun.javafx.application.LauncherImpl.lambda$launchApplication$2(LauncherImpl.java:195)
	at java.base/java.lang.Thread.run(Thread.java:833)
Caused by: java.lang.NullPointerException: Cannot invoke "qupath.lib.gui.viewer.QuPathViewer.downsampleFactor()" because "this.viewer" is null
```

But even if this is fixed, would it have any visible impact? The label text should already be updated.

I agree that downsample factor should be exposed as a property, but I think it will require a cautious approach; if it's not a `ReadOnlyDoubleProperty` then the user should be able to assume that they can also *set* the downsample via the property, but I expect that would require a bigger refactoring of `QuPathViewer`.

It's old-fashioned and suboptimal, but for now adding a [`QuPathViewerListener`](https://github.com/qupath/qupath/blob/007b18108b9f43954279403cb1f796d7ada4b521/qupath-gui-fx/src/main/java/qupath/lib/gui/viewer/QuPathViewerListener.java) is the expected way to identify any kind of viewer change.

Definitely a move towards a more property-based approach would be welcome, but then so would a viewer that uses JavaFX more natively (rather than lots of legacy Graphics2D stuff)... so I think the question is whether to invest time in refactoring the current viewer, or take on the task of writing an entirely new viewer?

I definitely think a new viewer would be nice at some point - would be good to discuss what to do along the way.

alanocallaghan:
Ah, sorry, I made this change on 0.4.3 where I was working on the extension and switched to main to cherry-pick without testing.

The context I spotted it is definitely an edge case.

1. zoom in an arbitrary amount, say past 1x
2. Click "Adjust zoom for all images to fit the entire image in the viewer"
3. Image downsample factor changes but displayed magnification doesn't

But this is fixed in main already it seems. Sorry!

PR #1282, opened 2023-07-24
Use collectors toList instead of bare toList in refreshExtensions

alanocallaghan:
Problem: `refreshExtensions` always throws `UnsupportedOperationException`, even if it succeeds.
Solution: use `Collectors.toList()` instead of the `Stream` `toList` method (which should be equivalent anyways imo...).

From: https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/stream/Stream.html#toList()

> The returned List is unmodifiable; calls to any mutator method will always cause UnsupportedOperationException to be thrown.

### DIFF
--- a/qupath-gui-fx/src/main/java/qupath/lib/gui/ExtensionManager.java
+++ b/qupath-gui-fx/src/main/java/qupath/lib/gui/ExtensionManager.java
@@ -177,7 +177,7 @@ public class ExtensionManager {
 		if (showNotification) {
 			// A bit convoluted... but try to show new servers that have been loaded by comparing with the past
 			List<String> serverBuilders = serverBuildersBefore.stream().map(s -> s.getName()).toList();
-			List<String> serverBuildersUpdated = ImageServerProvider.getInstalledImageServerBuilders().stream().map(s -> s.getName()).collect(Collectors.toList());
+			List<String> serverBuildersUpdated = ImageServerProvider.getInstalledImageServerBuilders().stream().map(s -> s.getName()).collect(Collectors.toCollection(ArrayList::new));
 			serverBuildersUpdated.removeAll(serverBuilders);
 			for (String builderName : serverBuildersUpdated) {
 				Dialogs.showInfoNotification("Image server loaded",  builderName);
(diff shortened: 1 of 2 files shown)


petebankhead:
Yeah, I was over-enthusiastic in converting `Collectors.toList()` to simply use `toList()` when updating to Java 17 - this problem has emerged a few times.

See https://github.com/qupath/qupath/commit/1710f51e7846b9b85367da2a716a8b67c8805e92 for more info.

The reason I didn't just undo it is that `Collectors.toList()` *also* seems to be wrong - at least inasmuch as it makes no guarantees of mutability: https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/stream/Collectors.html#toList()
So the fact it worked previously was kind of incidental or accidental.

Therefore my thinking was that we could stick with `Stream.toList()` where that works, and then where mutability is required we use instead
```java
stream.collect(Collectors.toCollection(ArrayList::new))
```

I hadn't spotted this as a failure case, so we certainly should change it - but I think better to make the change that explicitly generates an `ArrayList` to avoid relying upon the happenstance of `Collectors.toList()` being mutable.

alanocallaghan:
Ah, not guaranteeing is incredible frustrating. Yeah, best not to rely on the quirks of the underlying methods, have changed to use arraylist

alanocallaghan:
While I'm at it, the v0.4 extensions go by default to `v0.4/extensions/my-extension.jar` while in current main they go to `v0.5/my-extension.jar`, not sure if that's intended

petebankhead:
Doesn't sound intended... the `main` branch now is in a pretty unstable state mid-refactoring that I haven't had time to get back to, but which affected the extension stuff.

The relevant aspect is probably that the 'user directory' is increasingly used for other stuff (including css, localization strings, colormaps....) and not just extensions - so I created a new [`FileCopier`](https://github.com/qupath/qupath/blob/main/qupath-gui-fx/src/main/java/qupath/lib/gui/FileCopier.java) class to help with copying files over into the directory (i.e. when someone drags relevant files onto QuPath, they are copied to the appropriate subdirectory of the user directory... of which 'extensions' is just one). I suspect I messed up something there.

alanocallaghan:
Sensible, I'll do my best to fix any of the small bugs I spot when testing the current build

petebankhead:
Probably should, but it also needs to be v0.4.3 compatible so can't use `FileCopier`.

Not sure it matters *too* much which directory it uses, as long as we don't surprise users by writing stuff in unexpected places (or, even worse, overwriting anything...).

Sidenote: perhaps the biggest unresolved thing in `main` is that `qupath-fx` should be brought out into a separate repo, and switched to Apache. And *possibly* split even further, so that localization/preference stuff can be used elsewhere (e.g. log viewer, script editor...).

PR #1286, opened 2023-08-08
Fix PathClass singleton creation when a derived PathClass is requested

carlocastoldi:
When opening a project with derived classes in QuPath 0.4.0+, it currently wrongly adds them to `existingClasses`.

This simple PR aims to fix this

### DIFF
--- a/qupath-core/src/main/java/qupath/lib/objects/classes/PathClass.java
+++ b/qupath-core/src/main/java/qupath/lib/objects/classes/PathClass.java
@@ -825,6 +825,10 @@ public final class PathClass implements Comparable<PathClass>, Serializable {
 		// This can occur during deserialization
 		if (!pathClass.isDerivedClass() && pathClass.getName() == null)
 			return NULL_CLASS;
+		if (pathClass.isDerivedClass()) {
+			PathClass parent = getSingleton(pathClass.getParentClass());
+			pathClass = PathClass.getInstance(parent, pathClass.getName(), pathClass.getColor());
+		}
 		var previous = existingClasses.putIfAbsent(createCacheString(pathClass), pathClass);
 		return previous == null ? pathClass : previous;
 	}


petebankhead:
Thanks, could you describe in more detail what this fixes?

Can you provide any minimal example of the bug - ideally [with a unit test](https://github.com/qupath/qupath/blob/main/qupath-core/src/test/java/qupath/lib/objects/classes/TestPathClass.java)?

Based on the other thread I tried running
```groovy
pc1 = QP.getPathClass("Right")
pc2 = QP.getDerivedPathClass(pc1, "hello").getParentClass()
println pc1 === pc2
println pc1.hashCode()
println pc2.hashCode()
```
but that behaved properly, so I'm not sure what exactly is wrong.

carlocastoldi:
this is the minimum example I could write:

```groovy
import com.google.gson.Gson
import com.google.gson.stream.JsonReader
import qupath.lib.objects.classes.PathClass

String json = "{ \"name\": \"Example\", \"parentClass\": {\"name\": \"Parent\"}, \"colorRGB\": 0}";
JsonReader jin = new JsonReader(new StringReader(json));
Gson gson = new Gson();
PathClass pathClass = gson.fromJson(jin, PathClass.class);
pathClass = PathClass.getSingleton(pathClass);
PathClass parent1 = pathClass.getParentClass();
PathClass parent2 = PathClass.fromString("Parent");
println("getSingleton: "+parent1.hashCode()+" ("+parent1+")");
println("fromString: "+parent2.hashCode()+" ("+parent2+")");
println("Are equal: "+(parent1===parent2))
```

carlocastoldi:
Also: a pretty similar code could be used for this empty test
https://github.com/qupath/qupath/blob/ff797d180fbe180bf59e376270e3bfe32c424c82/qupath-core/src/test/java/qupath/lib/objects/classes/TestPathClassFactory.java#L188C37-L188C37

petebankhead:
Thanks, I understand better now that the bug is connected to deserializing.

I see that your JSON uses an 'old' (pre-v0.4) syntax:
```json
{ "name": "Example", "parentClass": {"name": "Parent"}, "colorRGB": 0}
```
If I run the following in v0.4.3
```groovy
def pc = getDerivedPathClass(getPathClass("First"), "Second")
println GsonTools.getInstance(true).toJson(pc)
```
I see
```json
{
  "names": [
    "First",
    "Second"
  ],
  "color": [117, 210, 222]
}
```
which was changed in v0.4 for better readability and easier support in other languages.

Also, for deserializing you should use `gson = GsonTools.getInstance()` rather than `new Gson()`, because QuPath installs its own `TypeAdapters` to handle custom classes, including `PathClass`.

So I implemented these changes in your script, randomizing the class names to avoid being thwarted by caching:
```groovy
import qupath.lib.objects.classes.PathClass

for (useLegacyJson in [true, false]) {

    String parentName = Math.random() as String
    String childName = Math.random() as String
    String json
    if (useLegacyJson)
        json = """{ "name": "${childName}", "parentClass": {"name": "${parentName}"}, "colorRGB": 0}"""
    else
        json = """{ "names": ["${parentName}", "${childName}"], "color": [0, 1, 2]}"""
    
    def gson = GsonTools.getInstance()
    PathClass pathClass = gson.fromJson(json, PathClass.class)
    pathClass = PathClass.getSingleton(pathClass)
    PathClass parent1 = pathClass.getParentClass()
    PathClass parent2 = PathClass.fromString(parentName)
    
    if (useLegacyJson)
        println "With legacy JSON"
    else
        println "With v0.4 JSON"
    println("getSingleton: " + parent1.hashCode() + " (" + parent1 + ")")
    println("fromString: " + parent2.hashCode() + " (" + parent2 + ")")
    println("Are equal: " + (parent1===parent2))    
    println "----------"
}
```
From a typical run I see
```
INFO: With legacy JSON
INFO: getSingleton: 1714766410 (0.07271672325316691)
INFO: fromString: 2014375544 (0.07271672325316691)
INFO: Are equal: false
INFO: ----------
INFO: With v0.4 JSON
INFO: getSingleton: 489320231 (0.016118546223767072)
INFO: fromString: 489320231 (0.016118546223767072)
INFO: Are equal: true
INFO: ----------
```

I think this implies that the bug *only* occurs when deserializing a PathClass written using the 'earlier' syntax, and doesn't affect a PathClass that was written by v0.4 itself.

Does that sound correct?

I'm on leave at the moment, but will explore the PR in more detail in a couple of weeks, and before the next QuPath release. In the meantime I'm keen to understand when the bug strikes.

carlocastoldi:
>I think this implies that the bug only occurs when deserializing a PathClass written using the 'earlier' syntax, and doesn't affect a PathClass that was written by v0.4 itself.
Does that sound correct?

Yeah, I think that is a correct deduction!

> In the meantime I'm keen to understand when the bug strikes.

This bug occurs when I open an image in a project with annotations with derived `PathClass`. These sort of project/images are made with [ABBA's extension](https://github.com/BIOP/qupath-extension-abba), however the bug occurs when opening such images in a QuPath v0.4.3 installation without installed extensions, too.
I am not a developer of ABBA extension, but I'm willing to investigate if you think it's necessary. I just need to understand where the serialized `PathClass`es of a project are saved to file!

> I'm on leave at the moment, but will explore the PR in more detail in a couple of weeks, and before the next QuPath release.

Thanks a ton!

NicoKiaru:
Hi all!

I'm trying to understand the bottom line of what seems to be a tricky bug. I do not really understand all of it... what I can tell is simply that Oli (@lacan) helped me to make a 'derived path class' in order to label if a Roi is on the left or on the right part of the brain:

https://github.com/BIOP/qupath-extension-abba/blob/d9384ba7e198a454cb8e9da4661988eac7ee285c/src/main/java/qupath/ext/biop/abba/AtlasTools.java#L248

After, the serialisation was delegated to QuPath, there was nothing custom here.

The abba extension has been in use since the v0.3, so both projects with the 0.3 serialization way and projects with the 0.4 serialisation way exist in the wild.

Since there won't be any new release of QuPath 0.4, I think I should look at solving this issue in the extension. Is there a way to do that ?

Again, sorry for my lack of understanding, I think I need to spend a full hour to understand it and I do not know when I'll be afford it.

petebankhead:
I've merged this because it definitely seems to fix an important bug, even if it's hard to replicate in isolation because it tends to require QuPath to be restarted in between.

I've confirmed that it can affect classifications that are deserialized either from JSON or from Java's serialization (e.g. a .qpdata file).

To demonstrate the latter, I set a opened and image and set a derived `PathClass` in a script as follows:
```groovy
def pathClass = getDerivedPathClass(getPathClass("Right"), "Here")
getSelectedObject().setPathClass(pathClass)
```
After saving the data and restarting QuPath, I selected the same object and ran
```groovy
def right = getPathClass("Right")
println right === getSelectedObject().getPathClass().getParentClass()
```
Before the change this prints `false`, indicating that the `PathClass` has been deserialized wrongly. With this PR, it prints `true`.

I can't think of a good way to unit test this, but the fix seems important.

Thanks @carlocastoldi !

carlocastoldi:
Finally we got to the bottom of the problem. Thanks you for the patience and for merging it!

When can we expect the next release?
I'm interested in updating all the old (v0.3) installations of QuPath in my laboratory so that we can all use some scripts i wrote taking advantage of a couple of features of v0.4+ ^^

petebankhead:
The aim is late September/early October, before [the San Diego Workshop](https://forum.image.sc/t/upcoming-qupath-training-course-october-9-11-san-diego/83751/12). There will be other changes...

Issue #1154, opened 2022-11-21
Long(ish) load times for QuPath project containing multiplexed images


MarkZaidi:
Currently building the latest version of QuPath; will close ticket if it ends up being fixed.

## Before we begin...

Before submitting your bug report, please check the following:

* [X] I've definitely found a bug (it you're not sure, please use [image.sc](https://forum.image.sc/tags/qupath) instead)
* [X] I've checked https://qupath.github.io for a new release that might already have fixed the issue
* [X] I've checked the [Changelog](https://github.com/qupath/qupath/blob/master/CHANGELOG.md) to see if the bug has already been fixed in the next release
* [X] I've checked for existing GitHub issues describing the same problem

## Bug report

**Describe the bug**
QuPath projects with several hundred IMC images take a notably longer time to load. Opening the images themselves is comparatively faster, yet are still read into memory. No noticeable change in memory usage is occurring when the QuPath project is being loaded, indicating that these images may not be pre-loaded (i.e. cached) during loading of the project itself. My guess is that the thumbnails are being reconstructed each time the QuPath project is loaded (computer must be restarted to reproduce, closing and relaunching QuPath is not sufficient). Given that thumbnail generation was updated in the changelog of the newest unofficial release, I'll build from source now and close the ticket if I can't reproduce the bug.

It's a minor inconvenience at best, especially since it's only a delay of two minutes. But keep in mind, the CPU here is a Ryzen 5950X with among the highest single thread speeds of desktop CPUs, and only 1 of the 32 threads is being used during this period. Depending on the root cause and a user's hardware configuration, load times may scale linearly or exponentially as project sizes move into the range of thousands of images.

If this is caused by thumbnail generation, then this may only be affecting multiplexed IHC images such as those acquired through IMC or CODEX.

**To Reproduce**
Video demonstration: https://www.youtube.com/watch?v=q4Jn9UTKUMw
1. Create a QuPath project
2. Load ~300 IMC images. Example IMC images can be found here: https://drive.google.com/file/d/1UcQmiIcjIfYdBVK1v6Phb32FBjHXciuH/view
3. Close QuPath and restart system
4. Launch QuPath and open the aforementioned project. Observe relatively long project load times

**Expected behavior**
Loading of a QuPath project should take a few seconds, especially if no measurements are present.

**Screenshots**
[...]

PR #1298, opened 2023-08-24
Replace all uses of Files.exist(path) with path.toFile().exists()

alanocallaghan:
I did this with `egrep -rl "Files\.exists\(\w+\)" | xargs sed -E -i 's/Files\.exists\((\w+)\)/\1.toFile().exists()/g'` so there's a non-zero chance I missed some usages (anything with a method or field access in the something of `Files.exists(something)` will be missed) and a small chance I broke something, but I didn't want the regex to get too complex.

Based on apparently slow performance, see
https://rules.sonarsource.com/java/RSPEC-3725/

https://github.com/qupath/qupath-extension-wsinfer/pull/30
https://github.com/qupath/qupath/issues/1154#issuecomment-1327044045

### DIFF
--- a/qupath-core-processing/src/main/java/qupath/lib/scripting/QP.java
+++ b/qupath-core-processing/src/main/java/qupath/lib/scripting/QP.java
@@ -4052,7 +4052,7 @@ public class QP {
 			if (classifier == null) {
 				try {
 					var path = Paths.get(name);
-					if (Files.exists(path))
+					if (path.toFile().exists())
 						classifier = ObjectClassifiers.readClassifier(path);
 				} catch (Exception e) {
 					exception = e;
@@ -4097,7 +4097,7 @@ public class QP {
 		}
 		try {
 			var path = Paths.get(name);
-			if (Files.exists(path))
+			if (path.toFile().exists())
 				return DensityMaps.loadDensityMap(path);
 		} catch (Exception e) {
 			exception = e;
@@ -4278,7 +4278,7 @@ public class QP {
 		}
 		try {
 			var path = Paths.get(name);
-			if (Files.exists(path))
+			if (path.toFile().exists())
 				pixelClassifier = PixelClassifiers.readClassifier(path);
 		} catch (Exception e) {
 			exception = e;
--- a/qupath-core-processing/src/main/java/qupath/opencv/ml/BioimageIoTools.java
+++ b/qupath-core-processing/src/main/java/qupath/opencv/ml/BioimageIoTools.java
@@ -124,13 +124,13 @@ public class BioimageIoTools {
 					// We can't currently handle zipped weights - so see if we can find unzipped weights
 					if (relativeSource.toLowerCase().endsWith(".zip")) {
 						pathWeights = basePath.resolve(relativeSource.substring(0, relativeSource.length()-4));
-						if (!Files.exists(pathWeights)) {
+						if (!pathWeights.toFile().exists()) {
 							logger.warn("Please unzip the model weights to {}", pathWeights);
 							continue;
 						}
 					} else {
 						pathWeights = basePath.resolve(relativeSource);
-						if (!Files.exists(pathWeights)) {
+						if (!pathWeights.toFile().exists()) {
 							logger.warn("Can't find model weights at {}", pathWeights);
 							continue;
 						}
--- a/qupath-core/src/main/java/qupath/lib/common/GeneralTools.java
+++ b/qupath-core/src/main/java/qupath/lib/common/GeneralTools.java
@@ -124,9 +124,9 @@ public final class GeneralTools {
 		// v0.2, less reliable way
 		if (version == null) {
 			var path = Paths.get("VERSION");
-			if (!Files.exists(path))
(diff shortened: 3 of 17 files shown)


petebankhead:
Nice regex :)

Makes me a bit nervous, since I'm not aware of anything currently being broken. Definitely worth considering `File.exists()` instead of `Files.exists(path)` when it's called a lot in a loop (and I guess anything *could* be called a lot in a loop by other code).

Any way to quickly benchmark the difference?

I guess we also want to know if one method is substantially faster or slower for checking files over a network, as well as local files.

(Conceivably we might also add our own method, e.g. `GeneralTools.checkFileExists()` or something and then use our preferred implementation in one place... but not sure if that's just adding awkwardness.)

alanocallaghan:
I'll benchmark Files.exists(file) vs file.exists() locally later, though it'll probably be worth checking performance on a samba mount as well as nfs/sshfs

alanocallaghan:
So on linux on 3000 files (random content, random size), `Files.exist(path)` takes 0.035 seconds while `file.exists()` takes 0.047 seconds. Pretty weird stuff, it may be a Windows issue

petebankhead:
I got curious too and tried an extremely rough and messy benchmarking as well - `Files.exist(path)` was quicker for me too (although I checked the same files multiple times, and lots of files that didn't exist).

Maybe a Windows, Java version, or network issue...

petebankhead:
I was seeing a bigger difference (`Files.exist(path)` taking about 1/3 the time)

alanocallaghan:
The linked issue seems very extreme, unless you are checking almost the whole HD then 260 seconds is unreasonable, and if switching to file.exists() solves that then I think it's worth taking a much smaller hit on other platforms

petebankhead:
From the other thread I'm not sure if it does solve the issue though :/ 

There have been obscure problems in the past through the use of `File` instead of `Path`, so nothing seems straightforward: https://github.com/Bayer-Group/paquo/issues/65#issuecomment-1312398115

alanocallaghan:
On windows:

```
INFO: Same: true
INFO: Paths 	37.279 s
Files 	17.887 s
Total duration	55.166 s
Average per checkpoint: 27.583 s
```

alanocallaghan:
Slightly updated script showing there is indeed some badness going on (note the counts don't match!)

```groovy
import groovy.transform.CompileStatic
import qupath.lib.common.Timeit

import java.nio.file.Files
import java.nio.file.Path

List<File> files = new File(System.getProperty("user.home")).listFiles() as List

for (int i = 0; i < 10; i++)
    files.add(new File(Math.random() as String))

List<Path> paths = files.collect(f -> f.toPath()) as List

int nIterations = 1000
println("Same: " + (countFilesExist(files) == countPathsExist(paths)))
println("Count files: " + countFilesExist(files))
println("Count paths: " + countPathsExist(paths))


println new Timeit()
        .checkpointAndRun("Paths ", () -> countPathsExist(paths, nIterations))
        .checkpointAndRun("Files ", () -> countFilesExist(files, nIterations))
        .stop()
        .summarizeCheckpoints()


@CompileStatic
int countFilesExist(List<File> files, int iterations=1) {
    int count = 0;
    for (int i = 0; i < iterations; i++) {
        for (def f : files) {
            if (f.exists())
                count++;
        }
    }
    return count;
}

@CompileStatic
int countPathsExist(List<Path> files, int iterations=1) {
    int count = 0;
    for (int i = 0; i < iterations; i++) {
        for (def f : files) {
            if (Files.exists(f))
                count++;
        }
    }
    return count;
}
```

Windows:
```
INFO: Same: false
INFO: Count files: 51
INFO: Count paths: 38
INFO: Paths 	4.273 s
Files 	1.838 s
Total duration	6.111 s
Average per checkpoint: 3.055 s
```

petebankhead:
Does it persist if you switch the order that you compare `Files` and `File`?

Also... why do they not match?!?

alanocallaghan:
Ubuntu/PopOS results, pretty much what I expected, more files less time, no inconsistency:

```
INFO: Starting script at Thu Aug 24 14:47:27 BST 2023
INFO: Same: true
INFO: Count files: 208
INFO: Count paths: 208
INFO: Paths 	598 ms
Files 	319 ms
Total duration	917 ms
Average per checkpoint: 458 ms
INFO: Total run time: 2.68 seconds
```

> Also... why do they not match?!?

The unholy combination that is Windows and Java, I guess

alanocallaghan:
The files that mismatch are things like `C:\Users\alano\Application Data`, which doesn't seem to exist, at least when I paste it into explorer I get `[directory] is not accessible. Access is denied`, and they're not visible when viewing the parent dir in explorer or cmd. Somewhat bizarre behaviour, can chalk it up to Windows I think

petebankhead:
Hmmm, in that case though we might prefer `paths` for not creating the impression that they exist?

I have the impression there isn't a clear winner here, and so I tend towards prefering `Files.exist` for its modern nature (and I see the implementation is at least partly different in an upcoming JDK) - although that could change it one turns out to have exceptionally horribly performance, e.g. for a network share.

Issue #1306, opened 2023-08-30
Derived classifications can be read incorrectly from data files and json


petebankhead:
## Bug report

> **TLDR:** There's a bug in how QuPath v0.4.0 - v0.4.3 saved/reloads (derived) classifications that *could* impact analysis results based upon counting classified objects.
> 
> In most cases, it shouldn't matter and won't affect results; in fact, at the time of writing I'm not aware of any real-world examples where it has negatively impacted QuPath's analysis. But because it *could* matter, particularly for complicated / non-standard workflows, we plan to include a fix in the next QuPath release.
> 
> In the spirit of openness, the following (somewhat technical) description attempts to explain the issue as clearly as possible.

**Describe the bug**
There is a subtle bug in QuPath v0.4.0 - v0.4.3 that strikes when reading ['derived' `PathClass` objects](https://qupath.readthedocs.io/en/0.4/docs/concepts/classifications.html#classifications-derived-classifications) from data files or (some) JSON representations.

Earlier QuPath versions are not affected.

Derived classifications include examples such as `Tumor: Positive` and `Tumor: Negative` - where both come from a base `Tumor` classification (represented by a `PathClass` object).

QuPath aims to ensure that every classification and derived classification is a singleton, i.e. it exists only once.

This means that requesting the parent class for `Tumor: Positive` and `Tumor: Negative` should always return the exact same classification object, and that should be identical to what is returned by `QP.getPathClass('Tumor')`.

The issue is that whenever a derived classification is *reloaded* in QuPath v0.4.x, a necessary check is missed and this makes it possible for the parent classification of `Tumor: Positive` and `Tumor: Negative` to *not* be identical to the result of `QP.getPathClass('Tumor')`.

This could potentially cause confusion when attempting to count objects that share the same classification.

**To Reproduce**
The bug is hard to reproduce, and I don't know of any way to reproduce it within a single QuPath session. It requires data to be saved, and then loaded into a new QuPath session (i.e. after closing QuPath and reopening it, or running QuPath on a different computer).

1. Perform a workflow that creates derived classifications, e.g. [positive tumor cell classification](https://qupath.readthedocs.io/en/0.4/docs/tutorials/cell_classification.html)
2. Save the data, restart QuPath, reload the data
3. Run a script such as the following to count the objects with a specified base classification
```groovy
println getCellObjects().count {
    it.getPathClass().getBaseClass() === getPathClass('Tumor')}
```
[...]

PR #1307, opened 2023-08-30
Potential v0.4.4 release

petebankhead:
Fixes https://github.com/qupath/qupath/issues/1306


carlocastoldi:
This still fixes #1306 for me!

May I ask you why isn't `petebankhead:0.4.4-fixes` rebased to `qupath:main`? Do you think it is possible to do the rebase without too much work?
It's not a big deal, but I'd like to have my rather small and insignificant commit 😁

Otherwise no problem! I get it and it's no biggie ^^

petebankhead:
Because I'm not great with git :) Could probably have cherry-picked it, but `main` has a *lot* of other changes, since the intention was that the next release would be v0.5.0 - and these should not be part of any v0.4.4 release.

My aim this morning was to create tests, working off the original v0.4.3 code to ensure that the tests were failing initially. This revealed another issue that wasn't covered through your fix (namely that deserializing a `PathClass` only works properly via a `PathObject`). So I approached it a different way by first introducing a new `PathClass.readResolve()` method, which solved that problem and also most of the original issue. I then found it was still possible to work around the use of `readResolve()` alone when deserializing legacy JSON - so I followed your approach to update the `getSingleton` method as you'd suggested as well to try to cover all scenarios.

The end result is that there are now 4 different deserialization tests, but it required the combination of approaches to pass all of them.

carlocastoldi:
Oh okay. Gotcha!
Thank you for the explanation. Looks super good to me

PR #1315, opened 2023-09-06
Toolbar, Menu and icon updates

petebankhead:
* Added icons & toolbar buttons for script editor and log viewer
* Fixed bug that meant menuitem graphics didn't update their color when the style changed
* Created new preference to control if the system menubar is used for all windows, the main window only, or no windows

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,10 @@
 This is a work-in-progress for the next QuPath release.
 
 ### Enhancements
+* New toolbar buttons for script editor `</>` and log viewer
 * *File &rarr; Export snapshots* supports PNG, JPEG and TIFF (not just PNG)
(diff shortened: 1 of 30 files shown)


finglis:
Looks good!
One sneaky icon that doesn't seem to be behaving as expected is a "C" under the "View" menu and "Show pixel classification". My need to hover over it to see it.
I'm not sure if the icon for the script editor may need slightly more space between each of the elements (eg "< / >" rather than "</>") It may just be by eyes but the forward slash makes the less-than and greater-than sign look different levels?

petebankhead:
Well-spotted with the `C`, I'll check that out later.

For the scripting icon, I used the one from Font Awesome so can't get any more fine-grained control, unfortunately. I assume it's an optical illusion...

PR #1316, opened 2023-09-06
Update ViewTrackerControlPane.java

petebankhead:
Fix exception when trying to open view tracking twice. More refactoring should be done in the future to avoid the use of the static dialog.


petebankhead:
Potentially, although I'm not seeing any difference on my Mac. 

Made a small guess and something that would help - can you pull down the latest code and try it out please?

Issue #1317, opened 2023-09-06
Detachable viewers (good or bad?)


petebankhead:
### Is your feature request related to a problem? Please describe.
QuPath was designed to have one main window. If there are to be multiple viewers, these are embedded in a grid.

The original thinking was that we couldn't handle too many images simultaneously, since they tend to be big.

But QuPath is now used for lots of different things, including with images of different sizes... and often with multiple monitors. The 'single window' could become very limiting. 

### Describe the solution you'd like
Not so much the 'solution I'd like' as the one I'm proposing: adding support for detaching viewers.

Here's a **very rough** example of it in action, hacked together with minimal code changes:

<img width="1188" alt="Detachable viewer demo" src="https://github.com/qupath/qupath/assets/4690904/933e5528-4bf9-461b-855d-04eeb6fe8ad2">

### Additional context
This isn't so much a feature request as an invitation for feedback.

My feeling is that it's worth exploring, but I'm not sure if it helps or just creates confusion.

My main questions are:
* Is this a feature anyone really wants?
* Is it likely to be more helpful or confusing for new users?
* If implemented, should it work by
   * Detaching one viewer (and leaving the rest of the 'viewer grid' intact), or
   * Breaking up the UI entirely, so that there is no longer a grid and instead we have separate windows (including for the toolbar and analysis pane, more like ImageJ)
* Is there a need to have the *same* image open in multiple viewers?
   * Note that this wouldn't permit having different objects or display modes (e.g. different kinds of overlays), but rather it would just allow different image regions to be visualized at the same time. That is already kind of possible with *View &rarr; Show mini viewer* - except the mini viewer doesn't support all the kinds of interaction that are possible with a full viewer.

Rylern:
I think it can be helpful because this feature is already present in several software like ImageJ, and I don't think it will be confusing for new users as I guess it will just require to add a new button to implement this.

Is that a request that has already been mentioned on the Image.sc forum? If not, shouldn't we ask the opinions of the QuPath users there? I'm not using QuPath like a regular user so my opinion here may not reflect theirs.

alanocallaghan:
> Is this a feature anyone really wants?

The answer to that is almost always "yes"; the better question is whether it is worthwhile to implement given the refactoring/maintenance burden it entails.

> Is it likely to be more helpful or confusing for new users?

It seems like the sort of thing that would be confusing if done accidentally, so we should try to make that difficult if possible, but as Leo says in imaging software it's common to have multiple images open.

[...]

Issue #6, opened 2016-10-17
Both main menubars displayed when ImageJ open but QuPath in focus in macOS


petebankhead:
Whenever ImageJ is opened from within QuPath (e.g. to send a region), if QuPath is brought back into focus both menubars are displayed together in macOS.

![imagej_double_menus](https://cloud.githubusercontent.com/assets/4690904/19434276/61c64a00-945c-11e6-9b60-082e25731428.jpg)

This appears to be related to ImageJ being AWT-based, and its request to use the system menubar conflicts awkwardly with QuPath's JavaFX-based use of the system menubar.

petebankhead:
Closing this because:

* I see no way to fix it completely
* there's now a preference to toggle whether the system menubar is used by QuPath... which provides a way in which to deal with the issue

PR #1318, opened 2023-09-06
Make viewers 'detachable'

petebankhead:
See discussion at https://github.com/qupath/qupath/issues/1317

This PR has become much more extensive and cover a few (unexpectedly) interrelated issues.

## Viewer grids & detached viewers

The PR aims to improve the use of multiple viewers more generally:

* The *View &rarr; Multi-view...* main menu and *Multi-view* popup menu (accessed by right-click on a viewer) now contain commands to set a fixed grid size of viewers: 1x1, 1x2, 2x1, 2x2, 3x3.
   * This makes it much easier to create a viewer grid for common sizes, without needing multiple steps to add rows/columns.
   * Text and icons show the expected grid shapes so they should be easy to interpret
* New commands are added to *Detach viewer from the grid* and subsequently to *Attach viewer to the grid*, so that viewers can be moved to their own floating windows and back again.

## PathPlugin & PluginRunner

When making the above changes, it became clear that a major refactoring & simplification of `PathPlugin` and `PluginRunner` was needed.

The original problem was that `PluginRunnerFX` proved to be buggy with detached windows, because the `ImageData` being handled was accessed via `PluginRunner.getImageData()`. In most cases that's fine, but `PluginRunnerFX` was always returning the *current* `ImageData` from QuPath, and if this *changed* before the plugin was complete (and had fired any hierarchy updates /  logged any workflow steps) then exceptions or unexpected behavior could occur.

It might conceivably have happened previously when using multiple viewers, but I don't know of any cases where it did - because QuPath locks the UI when plugins run, and this seemed to be working ok.

So this PR also includes a new design greatly simplifies `PluginRunner` to simple run tasks in parallel with progress notification - with some special logic for `PathTask` implementations.

Alongside that,
```java
PathPlugin.runPlugin(PluginRunner pluginRunner, String arg)
```
has been updated to
```java
PathPlugin.runPlugin(PluginRunner pluginRunner, ImageData<T> imageData, String arg)`
```
so that there's no need to request the `ImageData` from the plugin runner - and it's possible to ensure this remains unchanged while all the work of the plugin is completex, no matter what might go on in the UI.

Early indications are that the code is just simpler and more robust. However, it's still a core change that *could* potentially have broken lots of things, so most analysis-related commands ought to be checked (e.g. cell detection, or anything else involving `PathPlugin`).

## ImageJ, macOS & menubars

The ImageJ macro runner was an implementation of `PathPlugin`.

When exploring how this was affected, I dug into a longstanding bug since QuPath's beginning: basically, ImageJ and QuPath's menus don't play together nicely on macOS: https://github.com/qupath/qupath/issues/6

Shortcuts/accelerators are particularly badly affected: the QuPath ones either don't work, or would trigger both the accelerator within QuPath *and* within ImageJ.

This PR attempts to finally fix this, by
1. overriding any system menubar request from QuPath while ImageJ is running, so that QuPath's menu becomes 'demoted' to within the app
2. blocking ImageJ's last menubar whenever it no longer has any focussed stage (i.e. whenever a JavaFX window is activated)

This seems to work on macOS, but the menubar thing failed badly on Windows - so now is *only* applied with Mac.

@alanocallaghan @Rylern it would be good if you could check what happens on Linux if you:
* Run *Extensions &rarr; ImageJ &rarr; Send snapshot to ImageJ*
* Try to interact and use accelerators with both ImageJ and QuPath windows in focus. For example, pressing `shortcut+L` *should*
   * launch the ImageJ command finder if an ImageJ window is in focus
   * launch the QuPath command list if a QuPath window is in focus
   * *not* launch either the wrong window, or both at once
@finglis if you could also check this on Windows and the older MacBook that's be good, in case I've missed something.

## ImageJ extension toolbar button

The ImageJ extension now appears on the toolbar as a `MenuButton` rather than a `Button`. What it *does* (reveal a popup menu) remains unchanged.

The bad thing is that it takes more space. The good thing is that the menu now appears properly even if the toolbar is truncated, and which happens when the window size is small (like on my 13" Windows laptop screen).

## Revert new system menubar option

Along the way, I had to undo one of the improvements in https://github.com/qupath/qupath/pull/1315 - specifically

* Created new preference to control if the system menubar is used for all windows, the main window only, or no windows

It looked promising, but sadly using the system menubar for the main window and local menubars for other windows failed on macOS, inasmuch as accelerators run on a local menubar would then also be passed to the system menubar... meaning that `shortcut+r` in a script editor would not only run the script, but also offer to reload the data in the current viewer.


alanocallaghan:
> Run Extensions → ImageJ → Send snapshot to ImageJ
> Try to interact and use accelerators with both ImageJ and QuPath windows in focus. For example, pressing shortcut+L should
>    launch the ImageJ command finder if an ImageJ window is in focus
>    launch the QuPath command list if a QuPath window is in focus
>    not launch either the wrong window, or both at once

This is how it behaves, but I checked on a previous version and it also behaved this way.

"Send snapshot to ImageJ" always sends the image on the "primary" viewer, not the selected one. Is this intended?

Side note: movement/zoom is coupled between viewers, even when different images are open. Is this intended? Seems counterintuitive to me. Also is not great if you have an image highly zoomed and then open another, or if two images have different levels of detail available

PR #1319, opened 2023-09-08
Extract masked pixels

Rylern:
New method (+ corresponding test) to extract pixels from a ROI using OpenCV.
 
It extracts pixels from the input with non-zero values in the corresponding mask, limited to the requested (zero-based) channel index. The mask doesn't have to be multichannel; it could be, but also could be just a single channel that should be applied to all channels.
 
The corresponding unit test generates a binary image of a circle, and then duplicate it to use as the main input image. It checks if the mean value is 1 with the mask, and less than 1 without it.

### DIFF
--- a/qupath-core-processing/src/main/java/qupath/opencv/tools/OpenCVTools.java
+++ b/qupath-core-processing/src/main/java/qupath/opencv/tools/OpenCVTools.java
@@ -2678,8 +2678,11 @@ public class OpenCVTools {
 			opencv_core.extractChannel(mask, maskChannel, channel);
 		}
 
-		if (inputChannel.total() != maskChannel.total()) {
-			throw new IllegalArgumentException("The input and the mask must have the same dimensions");
+		if (inputChannel.rows() != maskChannel.rows()) {
+			throw new IllegalArgumentException("The input and the mask don't have the same number of rows");
+		}
+		if (inputChannel.cols() != maskChannel.cols()) {
+			throw new IllegalArgumentException("The input and the mask don't have the same number of columns");
 		}
 
 		int height = inputChannel.rows();
--- a/qupath-core-processing/src/main/java/qupath/opencv/tools/OpenCVTools.java
+++ b/qupath-core-processing/src/main/java/qupath/opencv/tools/OpenCVTools.java
@@ -2668,52 +2668,51 @@ public class OpenCVTools {
 	 * @throws IllegalArgumentException when the input and the mask don't have the same dimensions
 	 */
 	public static float[] extractMaskedPixels(Mat input, Mat mask, int channel) {
-		Mat inputChannel = new Mat();
-		opencv_core.extractChannel(input, inputChannel, channel);
+		try (PointerScope scope = new PointerScope()) {
+			Mat inputChannel = new Mat();
+			if (inputChannel.channels() == 1) {
+				inputChannel = input;
+			} else {
+				opencv_core.extractChannel(input, inputChannel, channel);
+			}
 
-		Mat maskChannel = new Mat();
-		if (mask.channels() == 1) {
-			maskChannel = mask;
-		} else {
-			opencv_core.extractChannel(mask, maskChannel, channel);
-		}
+			Mat maskChannel = new Mat();
+			if (mask.channels() == 1) {
+				maskChannel = mask;
+			} else {
+				opencv_core.extractChannel(mask, maskChannel, channel);
+			}
 
-		if (inputChannel.rows() != maskChannel.rows()) {
-			throw new IllegalArgumentException("The input and the mask don't have the same number of rows");
-		}
-		if (inputChannel.cols() != maskChannel.cols()) {
-			throw new IllegalArgumentException("The input and the mask don't have the same number of columns");
-		}
+			if (inputChannel.rows() != maskChannel.rows()) {
+				throw new IllegalArgumentException("The input and the mask don't have the same number of rows");
+			}
+			if (inputChannel.cols() != maskChannel.cols()) {
+				throw new IllegalArgumentException("The input and the mask don't have the same number of columns");
+			}
 
-		int height = inputChannel.rows();
-		int width = inputChannel.cols();
-		List<Float> outputList = new ArrayList<>();
+			int height = inputChannel.rows();
+			int width = inputChannel.cols();
+			float[] output = new float[height*width];
 
-		try (var idx = inputChannel.createIndexer()) {
-			try (var idxMask = maskChannel.createIndexer()) {
-				for (int y = 0; y < height; y++) {
-					for (int x = 0; x < width; x++) {
-						if (idxMask.getDouble(y, x) != 0) {
-							outputList.add((float) idx.getDouble(y, x));
+			long[] indices = new long[2];		// this array is used to avoid calling Indexer.getDouble() with varargs arguments (which is slower)
+			int outputIndex = 0;
+			try (var idx = inputChannel.createIndexer()) {
+				try (var idxMask = maskChannel.createIndexer()) {
+					for (int y = 0; y < height; y++) {
+						indices[0] = y;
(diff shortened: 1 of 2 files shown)


qupath-core-processing/src/main/java/qupath/opencv/tools/OpenCVTools.java
```diff
@@ -2651,6 +2651,70 @@ public static List<IndexedPixel> getMaskedPixels(Mat mat, Mat mask) {
...
+		if (inputChannel.rows() != maskChannel.rows()) {
+			throw new IllegalArgumentException("The input and the mask don't have the same number of rows");
+		}
+		if (inputChannel.cols() != maskChannel.cols()) {
+			throw new IllegalArgumentException("The input and the mask don't have the same number of columns");
+		}
+
+		int height = inputChannel.rows();
+		int width = inputChannel.cols();
+		List<Float> outputList = new ArrayList<>();
```
petebankhead, line 2690: This is likely to be a performance bottleneck, because `float` will need to be boxed to `Float`.

I guess the reason to use a list is that you can't know in advance how many pixels you'll need unless you iterate through the mask first, which would also have some performance hit.

My guess is that a good balance of efficiency would be to preallocate a `float[]` array capable of storing all the image pixels, and then populating it with unmasked pixels during iteration - storing an index, and skipping masked-out pixels.

And the end, you can check if the index is equal to the array length - if so, just return the array (i.e. the mask includes everything). If not, a single call to [`Arrays.copyOf(float[], int)`](https://docs.oracle.com/en/java/javase/20/docs/api/java.base/java/util/Arrays.html#copyOf(float[],int)) would return an array of the correct length.

qupath-core-processing/src/main/java/qupath/opencv/tools/OpenCVTools.java
```diff
@@ -2651,6 +2651,70 @@ public static List<IndexedPixel> getMaskedPixels(Mat mat, Mat mask) {
...
+
+		try (var idx = inputChannel.createIndexer()) {
+			try (var idxMask = maskChannel.createIndexer()) {
+				for (int y = 0; y < height; y++) {
+					for (int x = 0; x < width; x++) {
+						if (idxMask.getDouble(y, x) != 0) {
```
petebankhead, line 2696: Assuming this calls [`Indexer.getDouble(long...)`](https://javadoc.io/doc/org.bytedeco/javacpp/latest/index.html) (and not some overridden method) then it could be optimized further by preallocating a `long[]` and updating the `y` and `x` values within the loop.

The reason is that I expect [each call to a method with a `varargs` to be slower](https://stackoverflow.com/questions/2426455/javas-varargs-performance), since an array will need to be created. Applying this twice for each pixel might end up being a long time, whereas if you can pass a single `long[]` created in advance then it should be faster.

petebankhead (changes requested):
Thanks, I've added some comments.

Not important, but if you're sufficiently curious about it to benchmark the proposed optimizations, I'd be interested to know how much difference they make (I expect they'd require quite a large image to be noticeable).

Rylern:
Thanks, the execution time was roughly divided by two with these changes

qupath-core-processing/src/main/java/qupath/opencv/tools/OpenCVTools.java
```diff
@@ -2668,52 +2668,51 @@ public static List<IndexedPixel> getMaskedPixels(Mat mat, Mat mask) {
 	 * @throws IllegalArgumentException when the input and the mask don't have the same dimensions
 	 */
 	public static float[] extractMaskedPixels(Mat input, Mat mask, int channel) {
-		Mat inputChannel = new Mat();
-		opencv_core.extractChannel(input, inputChannel, channel);
+		try (PointerScope scope = new PointerScope()) {
+			Mat inputChannel = new Mat();
```
petebankhead, line 2672: You can avoid creating a `Mat` here, and delay it to the `else` block in case it is not needed.

(The overhead will be small, but may be more than 'normal' for Java since it will involve a trip through the `PointerScope` for allocation/deallocation, and JavaCPP-managed behavior)

qupath-core-processing/src/main/java/qupath/opencv/tools/OpenCVTools.java
```diff
@@ -2668,52 +2668,51 @@ public static List<IndexedPixel> getMaskedPixels(Mat mat, Mat mask) {
...
+				opencv_core.extractChannel(input, inputChannel, channel);
+			}
 
-		Mat maskChannel = new Mat();
-		if (mask.channels() == 1) {
-			maskChannel = mask;
-		} else {
-			opencv_core.extractChannel(mask, maskChannel, channel);
-		}
+			Mat maskChannel = new Mat();
```
petebankhead, line 2679: As above - no need to create the `Mat` here.

qupath-core-processing/src/main/java/qupath/opencv/tools/OpenCVTools.java
```diff
@@ -2668,52 +2668,51 @@ public static List<IndexedPixel> getMaskedPixels(Mat mat, Mat mask) {
...
 				}
 			}
-		}
 
-		if (inputChannel != input) {
-			inputChannel.close();
-		}
-		if (maskChannel != mask) {
-			maskChannel.close();
+			return Arrays.copyOf(output, outputIndex);
```
petebankhead, line 2714: The copy should only be necessary if `outputIndex < output.length`

qupath-core-processing/src/main/java/qupath/opencv/tools/OpenCVTools.java
```diff
@@ -2651,6 +2651,69 @@ public static List<IndexedPixel> getMaskedPixels(Mat mat, Mat mask) {
...
+	 * </ul>
+	 * <p>The input and the mask must have the same dimensions.</p>
+	 *
+	 * @param input  the image from where the pixels should be extracted
+	 * @param mask  the mask to apply to the input
+	 * @param channel  the channel index (zero-based) to consider
+	 * @return a float array containing the values of the input filtered by the mask
+	 * @throws IllegalArgumentException when the input and the mask don't have the same dimensions
+	 */
+	public static float[] extractMaskedPixels(Mat input, Mat mask, int channel) {
```
petebankhead, line 2670: My fault, sorry... I see that it should be `extractMaskedFloats` for similarity to existing methods (since `extractPixels` takes an output array).

petebankhead (changes requested):
Looks good, just small changes and one extra request...

Now that the logic is established, could it extended to include `extractMaskedDoubles`? I've just checked the previous API and it has both `extractFloats` and `extractDoubles`.

`extractFloats` will be better for efficiency most of the time, but `extractDoubles` is useful when
1. the input is 64-bit float (although that's rare)
2. there's a desire to use [`Arrays.stream(double[])`](https://docs.oracle.com/en/java/javase/20/docs/api/java.base/java/util/Arrays.html#stream(double%5B%5D)) (which is more common, and there's no equivalent for `float[]`)

petebankhead:
Yeah, I don't think it's possible without boxing - and better to avoid that.

Although since they aren't identical, could you add a corresponding test for the doubles function?

It would also be good to extend the test to handle different channels...

One a good way to do that is to create an RGB `BufferedImage` and then draw onto it with `Graphics2D`, after setting the foreground color to be `Color.GREEN` (for example). Similar code is at https://github.com/qupath/qupath/blob/67a1ed7ead8e28c54d120b21d08b7d41562eb8c3/qupath-core/src/test/java/qupath/lib/awt/common/TestBufferedImageTools.java#L176 but the type should be `BufferedImage.TYPE_INT_ARGB`.

Sorry to keep adding more, but I think the code is looking good - and it'll be nice to have it more well-tested than many other methods.

petebankhead (approved):

PR #1325, opened 2023-09-11
Use OpenSlide with JNA rather than JNI

alanocallaghan:
This is just proof of concept for the time being, while I check the OpenSlide builds.

Removes the dependency on the OpenSlide Jar, and copies the class files almost verbatim. This removes the dependency on openslide-java, so we just need libopenslide to be on the java library path.

Usage: should be unchanged. Speed: may be marginally slower, though with multiple threads and caching this doesn't seem noticable.


petebankhead:
Thanks! What do you think of shifting the OpenSlide extension to a repository of its own?

Now that I've learned about `includeBuild` and `includeFlat`, it's more feasible to develop extensions in separate repos - even if they end up bundled in the distribution by default.

alanocallaghan:
Looks like I need to also get my head around `includeBuild`, though the gradle documentation is as abstract as ever.

I don't really see a huge difference either way. Would we be pushing updates to the openslide (or other "core" extensions) without making a new QuPath release?

I guess it would make issue tracking (and responsibilities) more granular if we can gradually slim down the qupath repo, though that has the side-effect of making issues harder to find, then it's not really great.

It is straightforward to extract the git history: https://docs.github.com/en/get-started/using-git/splitting-a-subfolder-out-into-a-new-repository

qupath-extension-openslide/src/main/java/qupath/lib/images/servers/openslide/OpenSlide.java
```diff
@@ -0,0 +1,491 @@
...
+        HashMap<String, AssociatedImage> associated = new HashMap<>();
+        for (String s : OpenSlideJNA.INSTANCE
+                .openslide_get_associated_image_names(osr)) {
+            associated.put(s, new AssociatedImage(s, this));
+        }
+
+        associatedImages = Collections.unmodifiableMap(associated);
+
+        // store info for hash and equals
+        canonicalFile = file.getCanonicalFile();
```
petebankhead, line 157: This is where having a file would help... but we can try creating one here if we really need it.

qupath-extension-openslide/src/main/java/qupath/lib/images/servers/openslide/OpenSlide.java
```diff
@@ -0,0 +1,491 @@
...
+        int levelH = (int) Math.round(relativeDS * h);
+
+        // clip to edge of image
+        levelW = (int) Math.min(levelW, getLevelWidth(level) - levelX);
+        levelH = (int) Math.min(levelH, getLevelHeight(level) - levelY);
+        w = (int) Math.round(levelW / relativeDS);
+        h = (int) Math.round(levelH / relativeDS);
+
+        if (debug) {
+            System.out.println("levelW " + levelW + ", levelH " + levelH
```
petebankhead, line 285: Can use SLF4J for any printing

qupath-extension-openslide/src/main/java/qupath/lib/images/servers/openslide/OpenSlide.java
```diff
@@ -0,0 +1,491 @@
...
+        try {
+            checkDisposed();
+            OpenSlideJNA.INSTANCE.openslide_read_region(osr, dest, x, y, level, w, h);
+            checkError();
+        } finally {
+            rl.unlock();
+        }
+    }
+
+    public void paintRegion(Graphics2D g, int dx, int dy, long sx, long sy,
```
petebankhead, line 252: If we're looking for a minimal wrapper around OpenSlide, I don't think we need this - and I don't think QuPath uses it, in favour of https://github.com/qupath/qupath/blob/main/qupath-extension-openslide/src/main/java/qupath/lib/images/servers/openslide/OpenslideImageServer.java#L278

Basically, it has quite a bit of Java-specific logic that potentially reduces efficiency if what you really want at the end is a `BufferedImage` (not to paint to an existing `Graphics`).

qupath-extension-openslide/src/main/java/qupath/lib/images/servers/openslide/OpenSlide.java
```diff
@@ -0,0 +1,491 @@
...
+            }
+            g.fillRect(dx, dy, w, h);
+        }
+    }
+
+    final boolean debug = false;
+
+    private int debugThingy = 0;
+
+    public BufferedImage createThumbnailImage(int x, int y, long w, long h,
```
petebankhead, line 324: Don't think we need this either - it's just requesting a small image.

qupath-extension-openslide/src/main/java/qupath/lib/images/servers/openslide/OpenSlide.java
```diff
@@ -0,0 +1,491 @@
...
+    public BufferedImage createThumbnailImage(int maxSize) throws IOException {
+        return createThumbnailImage(0, 0, getLevel0Width(), getLevel0Height(),
+                maxSize);
+    }
+
+    public double getLevelDownsample(int level) {
+        return levelDownsamples[level];
+    }
+
+    public int getBestLevelForDownsample(double downsample) {
```
petebankhead, line 366: The logic here can be problematic sometimes because of how the levels are calculated (see above).

Also, you can end up having a near perfect level match, but still requesting a higher resolution (and many more tiles).

QuPath uses different logic with a bit more tolerance: https://github.com/qupath/qupath/blob/533ce0cf50ca2a5627b55a8c1fe351cdc72e0694/qupath-core/src/main/java/qupath/lib/images/servers/ServerTools.java#L93

In the end, I don't think the method is needed in a minimal wrapper.

qupath-extension-openslide/src/main/java/qupath/lib/images/servers/openslide/AssociatedImage.java
```diff
@@ -0,0 +1,68 @@
...
+ *  License along with OpenSlide. If not, see
+ *  <http://www.gnu.org/licenses/>.
+ *
+ */
+package qupath.lib.images.servers.openslide;
+
+import java.awt.image.BufferedImage;
+import java.io.IOException;
+
+public class AssociatedImage {
```
petebankhead, line 26: I'm not sure that we need this class at all for accessing associated images.

qupath-extension-openslide/src/main/java/qupath/lib/images/servers/openslide/OpenSlide.java
```diff
@@ -0,0 +1,491 @@
...
+        levelWidths = new long[levelCount];
+        levelHeights = new long[levelCount];
+        levelDownsamples = new double[levelCount];
+
+        for (int i = 0; i < levelCount; i++) {
+            long[] w = new long[1], h = new long[1];
+            OpenSlideJNA.INSTANCE.openslide_get_level_dimensions(osr, i, w, h);
+            levelWidths[i] = w[0];
+            levelHeights[i] = h[0];
+            levelDownsamples[i] = OpenSlideJNA.INSTANCE.openslide_get_level_downsample(
```
petebankhead, line 135: We *could* get the levels, but I'm not sure we should because it can give the wrong idea of what the appropriate level is for a resolution.

It's calculated by [averaging the width and height ratios](https://github.com/openslide/openslide/blame/cba7702258f777bd122432642f50d50aea0ef374/src/openslide.c#L254). However that can cause subtle problems due to rounding having forced the actual pixel dimensions to be slightly off what the (presumably) intended downsample is, and also from the width and height ratios often being different.

QuPath uses different logic that attempts to correct for this [here](https://github.com/qupath/qupath/blob/533ce0cf50ca2a5627b55a8c1fe351cdc72e0694/qupath-core/src/main/java/qupath/lib/images/servers/ImageServerMetadata.java#L1107).

petebankhead (commented):
I know it's just a draft, but I added some comments to the OpenSlide JNA bit (based on the JNI).

I'd rather not take the JNI code directly, mostly because I think it could/would/should end up being changed substantially - since there are different purposes, and we've also got more recent Java features available to us.

*If* the plan is to have a reusable JNA wrapper as a separate dependency then things might be different, but with the coming foreign function / memory API it probably wouldn't have a long lifespan. So I think we should aim for something minimal here: a single, fairly short file that does as little as possible on the Java side (with the Java logic kept inside the `ImageServer` class).

The JNI code also includes a whole slide viewer, so has a log of extra logic to support that - we can just drop pretty much all of it.

If you can then access the ICC Profiles it would be great. Returning a `byte[]` would be sufficient to [create an `ICC_Profile`](https://download.java.net/java/early_access/jdk19/docs/api/java.desktop/java/awt/color/ICC_Profile.html#getInstance(byte%5B%5D)) object.

qupath-extension-openslide/src/main/java/qupath/lib/images/servers/openslide/OpenSlide.java
alanocallaghan: I guess we can just remove any of the openslide logic for calculating downsamples? It doesn't seem to be used.

alanocallaghan:
Cheers, I put literally zero thought into what we need and not, though some of it definitely looked fishy. I figured it wasn't worth thinking about if it wasn't going to work in principle anyways.

The associated image class is useless yeah, it's just a String with a getBufferedImage method. No idea why they use a String -> String HashMap for that.

Also renamed "Openslide" to "OpenSlide" in the server(builder) classes for consistency, which I hope doesn't break anything

qupath-extension-openslide/src/main/java/qupath/lib/images/servers/openslide/OpenSlideServerBuilder.java
```diff
@@ -45,9 +45,9 @@
  * @author Pete Bankhead
  *
  */
-public class OpenslideServerBuilder implements ImageServerBuilder<BufferedImage> {
+public class OpenSlideServerBuilder implements ImageServerBuilder<BufferedImage> {
```
petebankhead, line 48: I see the reasoning, but since the class name is serialized in the project I think there's a strong chance this will break compatibility.

Breaking compatibility is sometimes necessary - especially for scripts / analysis, where a clean break is better than a subtle inconsistency - but if at all possible we'd want projects, images and objects to open properly in new versions. And, if saving the data, not to break any attempt to reopen the same data in an old QuPath version.

Will definitely need to check for that anyway. It might be possible to change the class, but ensure whatever is serialized is still interpretable by OpenSlide (e.g. using `"providerClassName": "openslide"`).

alanocallaghan: Ah, won't bother then, it's just aesthetic

petebankhead: I'm in favor of the aesthetics! Just flagging it's probably more awkward

alanocallaghan: I'm in favour of aesthetics generally, but not really at the risk of breaking changes

petebankhead: This is the offending bit of code: 
https://github.com/qupath/qupath/blob/533ce0cf50ca2a5627b55a8c1fe351cdc72e0694/qupath-core/src/main/java/qupath/lib/images/servers/ImageServerBuilder.java#L348-L350

Could modified it to give a bit more wiggle room to make changes in a future update (since a v0.6 project that doesn't work in v0.4 doesn't seem as bad as one that doesn't work in v0.5)

alanocallaghan:
Yeah this PR is better, the other should be basically equivalent bar the jars and some vague debugging attempts I've made.

Still need to figure out how to handle Windows, since there we need to extract 3 DLLs from the binary distribution

alanocallaghan:
If we don't end up having properly working bundled openslide, I'd suggest that we should be a bit louder about how to set OpenSlide if loading fails - a `Notification` at least, or maybe a `Message` the first time it happens and then log it afterwards.

petebankhead:
@alanocallaghan @finglis Is this working for you? I tried it on Windows this morning, but still no success - not sure if the problem is on my side.

alanocallaghan:
I pushed a new snapshot last night that worked for me, but it's unfortunately using the original 3.4.1 rather than the new 3.4.1, so I'm working on a fix for that now

alanocallaghan:
For what it's worth, I checked the dll dependencies with `ldd` in git bash, and it was all satisfied. I've had to switch to using their windows build because I ended up with unsatisfied links for SSP and some weird minGW libraries that I wasn't going to be able to fix in time, so it really should be a safe bet.

alanocallaghan:
Updated SNAPSHOT is live, @finglis could you try to run this branch and see if OpenSlide loads/works on Windows?

finglis:
Its running with no OpenSlide errors in the logs

alanocallaghan:
Obviously the prayers I've been sending to any deity that will take them have paid off, then. Cheers!

finglis:
Run on the mac and things aren't so positive... 1 moment

petebankhead:
@finglis on tenterhooks here, any updates? I've tried it on two different Macs (Intel & Apple Silicon) and it worked on both. What is not positive, and could it be permissions or OS-related?

alanocallaghan:
False alarm, it was the javafx plugin issue solved in https://github.com/qupath/qupath/pull/1325/commits/7043f57591b708d55fc82dc194dafdb5f2f885c9

finglis:
Apologies yes, false alarm! Was a user issue not a software one

PR #1327, opened 2023-09-12
Extracted pixel parsing from Bio-Formats

Rylern:
I created a first version of `OMEPixelParser` mentionned in #1287.

Is that what you had in mind @petebankhead ?

The class doesn't support arbitrary channels for now.


petebankhead:
I think this looks good & much cleaner than my previous version. It works well for all the images I've tried.

Should the logic be the same, just organized differently - or are there any changes you think I should look at closely?

Rylern:
The logic is the same, I just extracted the databuffer and the sample model creation in two different functions.

petebankhead:
Great, I'll merge then - thanks!

PR #1330, opened 2023-09-13
Openslide jna with jars

alanocallaghan:
Do not merge me I contain jars in a stupid place


petebankhead:
I have some proposed changes on the Java side - it looks like it can work quite nicely with taking a path to an OpenSlide distribution (here, homebrew).

Should I just push them here, or is there some other way I should share them...?

petebankhead:
Well, I pushed my suggestions - possibly to the right PR, possibly not (and at least not to this one).

PR #1339, opened 2023-09-21
Set min logviewer width to avoid clipping

alanocallaghan:
Before:

Screenshot from 2023-09-21 10-10-50

After:

Screenshot from 2023-09-21 10-11-55

### DIFF
--- a/qupath-gui-fx/src/main/java/qupath/lib/gui/QuPathGUI.java
+++ b/qupath-gui-fx/src/main/java/qupath/lib/gui/QuPathGUI.java
@@ -331,6 +331,9 @@ public class QuPathGUI {
 		populateScriptingMenu(getMenu(QuPathResources.getString("Menu.Automate"), false));
 		SystemMenuBar.manageMainMenuBar(menuBar);
 
+		stage.setMinWidth(800);
+		stage.setMinHeight(400);
+
 		logger.debug("{}", timeit.stop());
 	}
 	
--- a/qupath-gui-fx/src/main/java/qupath/lib/gui/commands/LogViewerCommand.java
+++ b/qupath-gui-fx/src/main/java/qupath/lib/gui/commands/LogViewerCommand.java
@@ -26,6 +26,7 @@ package qupath.lib.gui.commands;
 import javafx.application.Platform;
 import javafx.scene.Parent;
 import javafx.scene.Scene;
+import javafx.scene.layout.Region;
 import javafx.stage.Modality;
 import javafx.stage.Stage;
 import javafx.stage.Window;
@@ -52,7 +53,7 @@ public class LogViewerCommand implements Runnable {
 
 	private Window parent;
 
-	private Parent logviewer;
+	private Region logviewer;
 
 	/**
 	 * Constructor.
@@ -91,9 +92,8 @@ public class LogViewerCommand implements Runnable {
 			dialog.setResizable(true);
 			dialog.show();
 			dialog.sizeToScene();
-			// TODO: It would be nice to figure this out automatically
-			dialog.setMinWidth(650);
-			dialog.setMinHeight(440);
+			dialog.setMinWidth(logviewer.getPrefWidth());
+			dialog.setMinHeight(logviewer.getPrefHeight());
 			FXUtils.retainWindowPosition(dialog);
 		} else {
 			dialog.show();
--- a/qupath-gui-fx/src/main/java/qupath/lib/gui/commands/LogViewerCommand.java
+++ b/qupath-gui-fx/src/main/java/qupath/lib/gui/commands/LogViewerCommand.java
@@ -92,7 +92,7 @@ public class LogViewerCommand implements Runnable {
 			dialog.show();
 			dialog.sizeToScene();
 			// TODO: It would be nice to figure this out automatically
-			dialog.setMinWidth(600);
+			dialog.setMinWidth(650);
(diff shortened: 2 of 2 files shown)


Rylern:
> Thanks Alan! While you're there, did you happen to see if it can be set more sensible in an automatic way, as @Rylern suggested at [qupath/log-viewer#56 (comment)](https://github.com/qupath/log-viewer/issues/56#issuecomment-1723196589) ? My hard-coded values don't feel great... and, indeed, were wrong as you've shown.

I actually have to perform some small changes to the LogViewer before using its preferred size. I'm looking at it right now

alanocallaghan:
Using `Region` instead of `Parent` works fine, but yeah up to Leo on how to handle the sizing upstream

Rylern:
I updated my [pull request](https://github.com/qupath/log-viewer/pull/59) with these changes

alanocallaghan:
Seems missing from the FXML now? But maybe I'm missing some logic

alanocallaghan:
Closed the wrong PR :facepalm:

petebankhead:
Hmmm, when using https://github.com/qupath/log-viewer/pull/59 via `includeBuild` this PR isn't working for me - I'm getting an exception as soon as I attempt to open the log viewer, and the pref width & height are still being returned as -1
```
qupath.fx.dialogs.Dialogs - QuPath exception: The width and height must be >= 0. Got: width=-1; height=0
java.lang.IllegalArgumentException: The width and height must be >= 0. Got: width=-1; height=0
```

Is this working for others? If so, I guess something is wrong with my setup and it's really using the older code...

Rylern:
Sorry, yes you should use `logviewer.getWidth()` and `logviewer.getHeight()` instead

petebankhead:
Great, thanks - no need to update here, I'll add that into a new PR I'm making to use the latest changes.

petebankhead:
@Rylern I've just tried it, but it's still problematic... 

Using `getHeight()` doesn't entirely work, but only because I need to add the height of the title bar (so that can be overcome quite easily).

More importantly, the minimum dimensions are very large (862 x 679 px) and I expect that will be a problem for laptop users with small screens.

The main issue seems to be that the table is demanding too much vertical space; if I shrink vertically, I'd expect the divider to move proportionately (this happens if I expand beyond the expected height, but not if I shrink).

Also, I think the horizontal constraint should be relaxed to permit the window it less wide. It's possible to remove columns from the table, and shrink it further; the 'filter 'text field is sufficiently dynamic to support resizing up to a point, but then it reaches a minimum size and the rest of the toolbar becomes clipped.

I think that *this* is the 'true' minimum width that the dialog should support, not the width at the point the window is first shown (which is dominated by the space demanded by the table).

Rylern:
Is this size small enough?
image

petebankhead:
Yes, looks good to me!

alanocallaghan:
While we're at it, could we set a min size on the main qupath window? This seems a bit permissive

Screenshot from 2023-09-22 11-17-53

Rylern:
> Yes, looks good to me!

The pull request was updated

alanocallaghan:
The scene size of the main QuPath instance (1500x800) seemed a bit big for a minimum size, so I went with a mostly arbitrary 800x400. It's the minimum size where the GUI still looks semi-intelligible to me, before everything starts getting replace with ellipses. Also small enough that any computer newer than ~2005 should be fine

petebankhead:
> > Yes, looks good to me!
> The pull request was updated

How should I set the min width / height now on the QuPath side?

petebankhead:
BTW not planning to merge this one as I have a PR to add badges for the log viewer as well... so it's easier to just add the min sizes there rather than try to merge.

alanocallaghan:
okie doke, well up to you how you set the size on qupath but it's worth doing, at the moment you can make it so small it's almost impossible to find, which just seems silly

Rylern:
> > > Yes, looks good to me!
> > > The pull request was updated
> 
> How should I set the min width / height now on the QuPath side?

`logviewer.getWidth()` and `logviewer.getHeight()` should work

petebankhead:
Ok, that works. Still not ideal, since it means the log viewer always opens at a small size...

Is there no way to specify the *minimum* sizes for the log viewer component, rather than relying on the preferred sizes? The preferred size could (and probably should) be larger.

@alanocallaghan thanks, yup - I've added those lines to my code-in-progress.

petebankhead:
I'll merge the PR anyway, because it solves the problem of not being able to access the log message counts - so that bit is great. The minimum size looks good too, so if there's a way to enforce it as the minimum rather than preferred then that would be useful in a future PR.

Rylern:
> Is there no way to specify the _minimum_ sizes for the log viewer component, rather than relying on the preferred sizes? The preferred size could (and probably should) be larger.

I tried to set the minimum size of the log viewer to its computed size, and the preferred size to something arbitrary like 800. However, `logviewer.getMinWidth()` returns -1 in that situation.

A workaround would be (in the quPath side) to set the width/height of the window after setting its min width:
```java
dialog.setMinWidth(logviewer.getPrefWidth());
dialog.setMinHeight(logviewer.getPrefHeight());

dialog.setWidth(800);
dialog.setHeight(600);
```

petebankhead:
Thanks, may try that later. I know min/max/pref size things are very very awkward (I can never get them to behave entirely).

PR #1352, opened 2023-09-29
Simple barchart with selection

alanocallaghan:
- Usage:
  ```groovy
  def viewer = getCurrentViewer()
  def cells = getCellObjects()
  def builder = Charts.barChart()
      .viewer(viewer)
      .title('My barchart')
      .classes(cells)
      .show()
  ```
- Would be good to add the ability to plot a series, with parent classes also
- Handling of PathClass for selection is probably not optimal
- Also renamed Scatterchart to ScatterChart for consistency

### DIFF
--- a/qupath-gui-fx/src/main/java/qupath/lib/gui/charts/Charts.java
+++ b/qupath-gui-fx/src/main/java/qupath/lib/gui/charts/Charts.java
@@ -21,13 +21,7 @@
 
 package qupath.lib.gui.charts;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
@@ -37,12 +31,7 @@ import javafx.collections.FXCollections;
 import javafx.collections.ObservableList;
 import javafx.geometry.Side;
 import javafx.scene.Scene;
-import javafx.scene.chart.Axis;
-import javafx.scene.chart.Chart;
-import javafx.scene.chart.NumberAxis;
-import javafx.scene.chart.PieChart;
-import javafx.scene.chart.ScatterChart;
-import javafx.scene.chart.XYChart;
+import javafx.scene.chart.*;
 import javafx.scene.chart.XYChart.Data;
 import javafx.scene.chart.XYChart.Series;
 import javafx.scene.input.MouseEvent;
@@ -558,14 +547,14 @@ public class Charts {
 	/**
 	 * Builder for creating scatter charts.
 	 */
-	public static class ScatterchartBuilder extends XYNumberChartBuilder<ScatterchartBuilder, ScatterChart<Number, Number>> {
+	public static class ScatterChartBuilder extends XYNumberChartBuilder<ScatterChartBuilder, ScatterChart<Number, Number>> {
 	
 		private ObservableList<Series<Number, Number>> series = FXCollections.observableArrayList();
 		
 		private Integer DEFAULT_MAX_DATAPOINTS = 10_000;
 		private Integer maxDatapoints;
 
-		private ScatterchartBuilder() {}
+		private ScatterChartBuilder() {}
 		
 		/**
 		 * Choose the maximum number of supported datapoints.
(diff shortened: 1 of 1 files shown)


qupath-gui-fx/src/main/java/qupath/lib/gui/charts/Charts.java
```diff
@@ -834,6 +823,268 @@ public static ScatterchartBuilder scatterChart() {
...
+		 * @return this builder
+		 */
+		public <T> BarChartBuilder series(String name, Collection<? extends T> collection, Function<T, PathClass> xFun) {
+			var classAndCount = collection
+					.stream().map(xFun)
+					.collect(
+							Collectors.groupingBy(Function.identity(), Collectors.counting())
+					);
+			return series(name, classAndCount.entrySet().stream()
+					.map(e -> new Data<>(e.getKey().getName(), (Number)e.getValue(), e.getKey()))
```
petebankhead, line 928: Using `PathClass.getName()` just gets one component of the classification - you need `PathClass.toString()` to get all of them.
https://qupath.readthedocs.io/en/stable/docs/concepts/classifications.html
But this will also fail with unclassified objects (unless you replace `null` with [`PathClass.NULL_CLASS`](https://github.com/qupath/qupath/blob/ae4cc88429aea647b09bbd380782bc206237800b/qupath-core/src/main/java/qupath/lib/objects/classes/PathClass.java#L156)).

qupath-gui-fx/src/main/java/qupath/lib/gui/charts/Charts.java
```diff
@@ -834,6 +823,268 @@ public static ScatterchartBuilder scatterChart() {
...
+		 */
+		private void tryToSelect(PathClass pathClass, boolean addToSelection, boolean centerObject) {
+			PathObjectHierarchy hierarchy = null;
+			if (imageData != null)
+				hierarchy = imageData.getHierarchy();
+			else if (viewer != null)
+				hierarchy = viewer.getHierarchy();
+			if (hierarchy == null)
+				return;
+			var objects = pathObjects.stream().filter(p -> p.getPathClass().equals(pathClass)).toList();
```
petebankhead, line 1074: Need to consider that the `PathClass` can be null here.

qupath-gui-fx/src/main/java/qupath/lib/gui/charts/Charts.java
```diff
@@ -21,13 +21,7 @@
...
 package qupath.lib.gui.charts;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
petebankhead, line 24: I remain a bit attached to my import statements...

petebankhead (changes requested):
I like the look of this, and the 'selectability' is very nice indeed.
Currently, it doesn't handle `null` or 'derived' classifications.

Seeing it in action, I wonder if a stacked bar chart would be good for cases where there are two levels to the classification (e.g. `Tumor: Positive`, `Tumor: Negative`...  but then that wouldn't scale easily to [many more classifications](https://qupath.readthedocs.io/en/stable/docs/tutorials/multiplex_analysis.html) so might not be a good idea.

In any case, I think with basic `null` & derived classification support it would be mergeable for v0.5.0. If it doesn't happen already, it would also likely be necessary to sort the bars alphabetically.

qupath-gui-fx/src/main/java/qupath/lib/gui/charts/Charts.java
alanocallaghan: Must be a case of very stubborn IDE settings, or me forgetting to change them on a different computer

qupath-gui-fx/src/main/java/qupath/lib/gui/charts/Charts.java
alanocallaghan: I think this is handled now

qupath-gui-fx/src/main/java/qupath/lib/gui/charts/Charts.java
alanocallaghan: I think I've handled this, but I'm not exactly confident I've done it the right way

petebankhead (approved):

petebankhead:
I added a few changes so that it can handle nulls, and to support passing a `Map` input - which I think is a more Groovy way to create a bar chart than passing collections

```groovy
def viewer = getCurrentViewer()
def cells = getCellObjects()
cells = getDetectionObjects()
def builder = Charts.barChart()
    .viewer(viewer)
    .title('My barchart')
//    .classifications(cells) // Alternative
    .series("Hello", ["There": 1, "Here": 2]) // Using map
    .show()
````

Can reconsider the whole `Charts` class in the future, but it's enjoyable to click on bars to select so I think it's worth merging.

PR #1353, opened 2023-09-29
Set min dims for some stages

alanocallaghan:
Open for discussion on whether we should continue hunting for stages without min sizes set, or should just replace Stage with some thin wrapper that has some basic settings.

For example I'd love a keymapping to close the currently focused window (unless it's the main QuPath window). Usually shortcut+w closes the current tab in script editors, web browsers, etc.

### DIFF
--- a/qupath-gui-fx/src/main/java/qupath/lib/gui/ParameterDialogWrapper.java
+++ b/qupath-gui-fx/src/main/java/qupath/lib/gui/ParameterDialogWrapper.java
@@ -124,6 +124,9 @@ class ParameterDialogWrapper<T> {
 		}, btnRun.disabledProperty()));
 
 		final Stage dialog = new Stage();
+		dialog.setMinWidth(300);
+		dialog.setMinHeight(200);
+
 		QuPathGUI qupath = QuPathGUI.getInstance();
 		if (qupath != null)
 			dialog.initOwner(qupath.getStage());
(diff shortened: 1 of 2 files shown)


petebankhead:
Not sure about the first point; I still struggle with JavaFX's min/max/pref sizes, and other more annoying bugs persist (like the growing dialog when training a pixel classifier).

For the second, the possibly-cleanest way would be to make it a command in the main menubar. If it's fired, then it hunts for the in-focus window using logic somewhat like https://github.com/qupath/qupath-fxtras/blob/9b043792b3b454f1376e57c32e36f8a74aa482ad/src/main/java/qupath/fx/dialogs/Dialogs.java#L486

That way it should work for stages that don't have menubars, for whom the events propagate to the main stage... or at least I think it would. I've failed with menubars and accelerators before, and the behavior across platforms also varies a bit - so it needs careful checking.

petebankhead:
shortcut+w already works in the script editor and log viewer (but only because they explicitly add that as an accelerator).

alanocallaghan:
> For the second, the possibly-cleanest way would be to make it a command in the main menubar. If it's fired, then it hunts for the in-focus window

That makes a lot of sense; would also mean it's documented somewhere.

> Not sure about the first point; I still struggle with JavaFX's min/max/pref sizes, and other more annoying bugs persist (like the growing dialog when training a pixel classifier).

Makes sense

Issue #1027, opened 2022-07-28
Please consider posting hash information for .msi and standalone .exe


KrisJanssen:
Organizations who would like to use your tool might want to verify the binaries you distribute are not being tampered with in transet. One easy way to make that easy (as opposed to have such organizations build your tool from scratch) would be to publish the binaries with MD5/SHA1 hashes on the releases page.

You would make the life of IT employees in such organizations a whole lot easier!

Thanks,

Kris

KrisJanssen:
I recently found resources like this:

Release hash:
https://github.com/MCJack123/ghaction-Generate-Release-Hashes

Release signing:
https://wiki.debian.org/Creating%20signed%20GitHub%20releases (which equally allows one to verify a given download)

mezwick:
Is an sha256 hash available for the current Qupath (0.3.2) release? (specifically the windows distribution?)

petebankhead:
@mezwick Hashes aren't currently generated or made available.

Making a release remains a fairly laborious and manual process. Although the builds themselves are now generated automatically using GitHub Actions, I have to download and check these run on each platform and then upload again. And write all the release notes, tag the version etc. There can also be some extra renaming required, since `jpackage` (used for the build) has some awkwardness connected to artefact naming and 0.x.x versions that affects some platforms but not others (e.g. I think macOS forbids 0.x.x versions, so this needs worked around; also, it needs to be possible for people to have multiple versions installed for reproducibility).

It's already a real pain to do, and I don't want to add any more manual steps if I can avoid it.

It would be strongly preferable to automate the whole process a bit more, using [Upload to Release](https://github.com/marketplace/actions/upload-to-release) to avoid the download/upload requirement, and somehow include hashes (e.g. using the links from @KrisJanssen's last post) at that point.

This seems to me at least a bit awkward to set up though, since the upload action is only triggered when a release is made. The [build workflow](https://github.com/qupath/qupath/blob/v0.3.2/.github/workflows/jpackage.yml) would have to be quite a bit more complex (e.g. to handle cross-platform filenames/content types for the builds, as well as different artefact compression methods). I'd also still need to retain the ability to check the release manually on each platform *before* the release itself is created, because creating the release is what triggers any update notifications... and after the release has been tagged is a bad time to identify some platform-specific breakage. These tend to happen with every release, e.g. because some dialog ends up misbehaving on Ubuntu but looks fine everywhere else.

[...]

PR #1366, opened 2023-10-04
Make both rpm and deb installers for linux builds

alanocallaghan:
Could add something to the PlatformPlugin code, but I think it's simpler just to explicitly build RPMs on Fedora.

If we remove the tar.gz, we should probably provide instructions to make them from source, just in case somebody's workflow relies on it.

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -8,21 +8,20 @@
  jobs:
    build:
 
-     strategy:
-       matrix:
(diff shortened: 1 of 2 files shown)


petebankhead:
Looks good to me. Re. workflows, the question I've seen a couple of times is this one:
* https://github.com/qupath/qupath/issues/1027

Not sure if it's easy to resolve - in any case, happy to merge this any time.

alanocallaghan:
Ah yes! Forgot about the hashes, should be an easy fix

PR #1368, opened 2023-10-05
Add gradle checksum plugin

### LINKED ISSUE #1027: Please consider posting hash information for .msi and standalone .exe

alanocallaghan:
This should create eg `build/dist/QuPath-0.5.0-SNAPSHOT.deb.sha512` with the checksum and the file name, for all platforms.

Hopefully resolve #1027

### DIFF
--- a/qupath-app/build.gradle
+++ b/qupath-app/build.gradle
@@ -19,6 +19,7 @@ buildscript {
 }
 
 plugins {
+    id 'org.gradle.crypto.checksum' version '1.4.0'
     id 'qupath.common-conventions'
     id 'qupath.djl-conventions'
     id 'application'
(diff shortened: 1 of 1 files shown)


petebankhead:
This wasn't working for me, but I've updated the PR and also incorporating the architecture into the package name for macOS (since it should then be easier to have x64 and aarch64 installed side-by-side)

PR #1388, opened 2023-10-13
Use 4.0.0 snapshot for openslide

alanocallaghan:


alanocallaghan:
Need to check this runs nicely across platforms so paging @finglis @Rylern @petebankhead and @alanocallaghan for mac x86-64, fedora, mac m1, and ubuntu respectively

alanocallaghan:
Still has 3.4.1 under the hood on Ubuntu at least, sorry if I tagged mistakenly

alanocallaghan:
No, it just picks up my system libopenslide.so before the Jar version, which I think is not ideal, but hopefully fine once the versions get pushed to homebrew/apt/rpm

petebankhead:
> No, it just picks up my system libopenslide.so before the Jar version, which I think is not ideal

Hmmm, is this something we should address through JNA?

In any case, it seems to be working fine here (just tried running from Gradle on M1, and both the x86_64 and aarch64 builds were working) - so I think that's good news for making the native release, or we can stick with the versioned snapshot for now if you prefer.

The aim is to get rc2 out shortly, but the final v0.5.0 won't be until next week at least.

We'll also need to update the info at https://github.com/qupath/qupath/tree/main/qupath-extension-openslide/src/main/resources/licenses/OpenSlide

My guess is that taking the licenses from the Windows build should be sufficient, or is anything radically different / missing if we do that?

alanocallaghan:
> Hmmm, is this something we should address through JNA?

Not sure honestly, in retrospect maybe respecting people's installed OpenSlide version is good, only falling back on the version we bundle if we have to? But at the moment it's a bit backwards, with the apt version behind ours.

 > In any case, it seems to be working fine here (just tried running from Gradle on M1, and both the x86_64 and aarch64 builds were working) - so I think that's good news for making the native release, or we can stick with the versioned snapshot for now if you prefer.

Great, I want to make sure the main artifact (without qualifiers) includes everything, and to include the sources on maven, will do that before the main release.

> My guess is that taking the licenses from the Windows build should be sufficient, or is anything radically different / missing if we do that?

I *think* the licenses should be the same across versions now for the most part - the extra windows libraries are removed.

petebankhead:
I think that we should probably use the bundled OpenSlide by default, and the system one if the user sets a path (or deletes the bundled one...). But I expect few users would be affected by this.

alanocallaghan:
Last commit should load directly the file from the jar, unless the openslide path is set in prefs

alanocallaghan:
Licenses are up-to-date. For the release of the natives Jar for the final 0.5 release I'd like to have everything semi-cleanly coming from one repo, rather than currently taking their dll directly from openslide.org for windows

Therefore this is good to go, unless you want to clean up the loading code

qupath-extension-openslide/src/main/java/qupath/lib/images/servers/openslide/jna/OpenSlideLoader.java
```diff
@@ -69,6 +69,19 @@ public static synchronized boolean tryToLoad(String... searchPath) throws Unsati
 
     private static synchronized OpenSlideJNA tryToLoadJnaInstance(String... searchPath) {
         String jnaPath = System.getProperty("jna.library.path", null);
+        if (searchPath.length == 0 || (searchPath.length == 1 && searchPath[0].isEmpty())) {
+            logger.info("No search path provided; trying to load OpenSlide from packaged jar");
```
petebankhead, line 73: This will be logged every time under 'normal' conditions - so I think it should be at the `debug` level.

Also, it *might* not be true - since `jna.library.path` could have been supplied by the user (although probably wasn't).

qupath-extension-openslide/src/main/java/qupath/lib/images/servers/openslide/jna/OpenSlideLoader.java
```diff
@@ -69,6 +69,19 @@ public static synchronized boolean tryToLoad(String... searchPath) throws Unsati
...
+            try {
+                File openslideFile = Native.extractFromResourcePath("openslide");
+                return Native.load(openslideFile.getAbsolutePath(), OpenSlideJNA.class);
+            } catch (IOException e) {
+                logger.error("Failed to load OpenSlide from packaged jar");
```
petebankhead, line 78: Could warn here that no search path was specified (so that this message is visible only when there has been trouble).

qupath-extension-openslide/src/main/java/qupath/lib/images/servers/openslide/jna/OpenSlideLoader.java
```diff
@@ -69,6 +69,19 @@ public static synchronized boolean tryToLoad(String... searchPath) throws Unsati
...
+                if (jnaPath == null)
```
petebankhead, line 79: Unless I've missed something, I don't think there's any need for that here, since `jna.library.path` hasn't been modified.

qupath-extension-openslide/src/main/java/qupath/lib/images/servers/openslide/jna/OpenSlideLoader.java
```diff
@@ -69,6 +69,19 @@ public static synchronized boolean tryToLoad(String... searchPath) throws Unsati
 
     private static synchronized OpenSlideJNA tryToLoadJnaInstance(String... searchPath) {
         String jnaPath = System.getProperty("jna.library.path", null);
```
petebankhead, line 71: I struggled with the logic of this method; if the `jna.library.path` isn't used in the event that `searchPath` is empty, then I think better to delay requesting the property until closer to the time it is needed (i.e. after the `if` block).

petebankhead (changes requested):
Great, thanks! Added small comments regarding `tryToLoadJnaInstance` - since the search path can be a thorny problem, I think it needs to be as simple as possible with cautious logging.

petebankhead (approved):

PR #1415, opened 2023-11-06
Fix extension directory handling when it doesn't yet exist

alanocallaghan:
Was unable to figure out the other drag & drop problem.

To reproduce the other issue:

1. Drag & drop any URL or file
2. Drag and drop another URL from the same source.

When drag&drop events come from the same source (eg, a web browser), the first URL/file/etc will persist for all following drag/drop events, on Linux at least. Clearing the dragboard programmatically has no effect (maybe expected, I'd assume it's cleared or a new one is created between events anyway).

### DIFF
--- a/qupath-gui-fx/src/main/java/qupath/lib/gui/ExtensionControlPane.java
+++ b/qupath-gui-fx/src/main/java/qupath/lib/gui/ExtensionControlPane.java
@@ -78,7 +78,6 @@ public class ExtensionControlPane extends VBox {
     @FXML
     private ListView<QuPathExtension> extensionListView;
 
-
     @FXML
     private Button rmBtn;
     @FXML
@@ -87,7 +86,6 @@ public class ExtensionControlPane extends VBox {
     private Button submitBtn;
     @FXML
     private Button openExtensionDirBtn;
-
     @FXML
     private Button updateBtn;
     @FXML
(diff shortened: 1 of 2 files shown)


petebankhead:
I'll merge this because I think it solves the initial problem with exceptions due to the extensions directory not existing.

I couldn't replicate the other issue on my Mac, although I'm not certain I was following the steps correctly. I first dragged https://github.com/qupath/qupath-extension-wsinfer from a browser, and then https://github.com/qupath/qupath-extension-stardist

Both extensions seemed to install 'properly', except WSInfer wasn't compatible so that confused things slightly by throwing exceptions (rightly). Immediately after installation, the extension manager wrongly showed the StarDist extension version as being v0.5.0-SNAPSHOT, but after relaunching QuPath this was corrected to v0.4.0.

alanocallaghan:
I have a very strong suspicion that it's a Linux issue with the system dragboard, but I've not managed to narrow it down. When I said 

> come from the same source

I think I was maybe wrong - I tried dragging the same URL from a text editor, and it still didn't work as expected. I did get a `file://` link to overwrite a `https://` link, but I couldn't do the reverse...

alanocallaghan:
Just coming back to this now, hopefully the gif demonstrates it better than I was explaining it

output

petebankhead:
Aha, tried the first bit (extensions) and couldn't replicate the problem on my Mac and on Windows - the prompt appeared for the correct extension both times. So I presume the weirdness is Linux-specific.

alanocallaghan:
Leo reports a different drag & drop issue (something like this: https://bugs.openjdk.org/browse/JDK-8275033, fixed in JavaFX 21). I tried bumping to javafx 21, but it doesn't fix my issue. Sadly may have to live with it, or disable some d&d abilities on Linux.

petebankhead:
I think JavaFX 21 (or anything above 20) breaks Javadoc viewing support, so we're stuck on 20 for now anyway.

alanocallaghan:
Yep, saw the comment in libs.version.toml, was just confirming that the linked openjdk issue is probably unrelated

PR #1419, opened 2023-11-09
Better handle failed delete extension jar

petebankhead:
Addresses https://forum.image.sc/t/please-help-us-test-the-qupath-v0-5-0-release-candidate/87976/12

If an extension can't be deleted, inform the user and ask if the directory should be opened for manual deletion. Previously, a 'success' message was shown even if the deletion had failed.

I couldn't figure out a way to make the extension 'deletable' - although I vaguely recall @yli-hallila might know more about this.

### DIFF
--- a/qupath-core/src/main/java/qupath/lib/common/GeneralTools.java
+++ b/qupath-core/src/main/java/qupath/lib/common/GeneralTools.java
@@ -754,14 +754,15 @@ public final class GeneralTools {
 	 * 
 	 * @param fileToDelete
 	 * @param preferTrash
+	 * @return true if the file is successfully deleted, false otherwise
 	 */
-	public static void deleteFile(File fileToDelete, boolean preferTrash) {
+	public static boolean deleteFile(File fileToDelete, boolean preferTrash) {
 		if (preferTrash && Desktop.isDesktopSupported()) {
 			var desktop = Desktop.getDesktop();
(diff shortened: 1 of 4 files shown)


yli-hallila:
I had the issue of not being able to delete extensions when QuPath was running. Somewhere in the loading process of the extension the ClassLoader (?) was never closed which left the operating system thinking that the file is still open by some process. I just closed the ClassLoader after initializing the extension, which allowed me to then delete the extension. 

Unsure if closing the ClassLoader has consequences such as not being access the Resource files or such?

petebankhead:
Interesting, thanks! I didn't realize classloaders can be closed... I see they can't generally, but the URLClassLoader can:
https://docs.oracle.com/en/java/javase/20/docs/api/java.base/java/net/URLClassLoader.html#close()

I suspect that would mean no new extensions can be added, but if the user obediently restarts QuPath at that point then it wouldn't matter.

Or there might be a more clever way to manage loading extensions with different classloaders. In any case, that does indeed make it possible to delete extensions on Windows - thanks!

alanocallaghan (approved):
Looks good, didn't test on windows though

petebankhead:
Thanks, I *only* tested it on Windows so far (which isn't much fun with a 13" screen, no battery life, and continually pressing the keys that would correspond to the Mac shortcuts).

PR #1429, opened 2023-11-30
Reintroduce startup scripts

petebankhead:
- New *Startup script path* preference
  - Support running a single Groovy script when QuPath launches (useful for customization)
  - The preference can also be overridden with the system property `qupath.startup.script=/path/to/script.groovy` or blocked with `qupath.startup.script=false` (e.g. in QuPath.cfg)

Intended to help with CUDA setup on Linux.
Likely to please @MichaelSNelson but that is unintentional.

### DIFF
--- a/qupath-gui-fx/src/main/java/qupath/lib/gui/QuPathGUI.java
+++ b/qupath-gui-fx/src/main/java/qupath/lib/gui/QuPathGUI.java
@@ -393,7 +393,8 @@ public class QuPathGUI {
 		}
 		try {
 			logger.info("Running startup script {}", path);
-			Dialogs.showInfoNotification("Startup script", "Running startup script\n" + file.getName());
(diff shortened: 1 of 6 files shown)


qupath-gui-fx/src/main/java/qupath/lib/gui/QuPathGUI.java
```diff
@@ -360,8 +360,46 @@ private QuPathGUI(final Stage stage, final HostServices hostServices, boolean sh
...
+			return;
+		}
+		var file = new File(path);
+		if (!file.exists()) {
+			logger.warn("Startup script does not exist: {}", path);
+			return;
+		}
+		try {
+			logger.info("Running startup script {}", path);
+			Dialogs.showInfoNotification("Startup script", "Running startup script\n" + file.getName());
```
alanocallaghan, line 396: String should be in resources?

Rylern (approved):

PR #1430, opened 2023-11-30
Set system properties from command line

petebankhead:
This provides more opportunities to customize QuPath at launch, rather than afterwards. It may be useful for setting paths or options for DJL or JNA, without relying on persistent preferences.

Example:
```
/path/to/QuPath -DPYTORCH_VERSION=1.13.1 -Doffline=false
```

It's also possible to do
```
/path/to/QuPath -Djava.library.path=/something/
```
but I'm not sure if that works... it should however work to set JNA paths, see https://java-native-access.github.io/jna/4.2.1/com/sun/jna/NativeLibrary.html


petebankhead:
@alanocallaghan @Rylern @finglis can anyone think of a reason not to do this...?

The main goal is to be able to set awkward DJL/CUDA options from a command line (e.g. .bat file or shell script), rather than needing configuration later.

May or may not be successful; if it works, it may make https://github.com/qupath/qupath/pull/1429 unnecessary (but probably worth keeping anyway).

Rylern (approved):

alanocallaghan (approved):

PR #1432, opened 2023-12-01
Script font size

petebankhead:
Add *View → Set font size*

This was an increasingly common request - especially for use in training courses.

Currently, it is possible to put in any css styling code, e.g. `-fx-font-weight: bold;` (although probably isn't a good idea).

### DIFF
--- a/qupath-gui-fx/src/main/java/qupath/lib/gui/scripting/DefaultScriptEditor.java
+++ b/qupath-gui-fx/src/main/java/qupath/lib/gui/scripting/DefaultScriptEditor.java
@@ -156,6 +156,43 @@ public class DefaultScriptEditor implements ScriptEditor {
 	private static final Logger logger = LoggerFactory.getLogger(DefaultScriptEditor.class);
 			
 	private final ScriptEditorDragDropListener dragDropListener;
+
+	/**
+	 * Font sizes available for the script editor.
+	 */
+	public enum FontSize {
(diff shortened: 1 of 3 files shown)


qupath-gui-fx/src/main/java/qupath/lib/gui/scripting/DefaultScriptEditor.java
```diff
@@ -526,7 +538,40 @@ public Stage getStage() {
...
+			else
+				style = "-fx-font-size: " + output;
+			if (!style.endsWith(";"))
+				style = style + ";";
+		}
+		// This only works if no other style is set!
+		fontSize.setValue(style);
+		paneCode.setStyle(style);
+		paneConsole.setStyle(style);
+	}
```
Rylern, line 572: Shouldn't you check that `style != null` before those 3 lines?

petebankhead: I'd actually like style to be `null` - that should reset the style so that it depends upon any existing css/stylesheets. So the default behavior should be a style of null.

alanocallaghan:
> Currently, it is possible to put in any css styling code, e.g. -fx-font-weight: bold; (although probably isn't a good idea).

I'd be inclined to only support integers in a sensible range, but I guess the flexibility could be nice for some niche cases.

Otherwise lgtm

alanocallaghan (approved):

finglis (approved):

PR #1447, opened 2024-01-17
Change openslide loading behaviour

alanocallaghan:
When running a script from the command line, extensions aren't loaded until much later, so we throw an error if you try to open an image with openslide.

See https://forum.image.sc/t/running-qupath-through-command-line-unable-to-access-image-data/90746 for an example

### DIFF
--- a/qupath-extension-openslide/src/main/java/qupath/lib/images/servers/openslide/OpenslideServerBuilder.java
+++ b/qupath-extension-openslide/src/main/java/qupath/lib/images/servers/openslide/OpenslideServerBuilder.java
@@ -23,20 +23,19 @@
 
 package qupath.lib.images.servers.openslide;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import qupath.lib.common.LogTools;
+import qupath.lib.images.servers.FileFormatInfo;
+import qupath.lib.images.servers.FileFormatInfo.ImageCheckType;
+import qupath.lib.images.servers.ImageServer;
+import qupath.lib.images.servers.ImageServerBuilder;
+import qupath.lib.images.servers.openslide.jna.OpenSlideLoader;
+
 import java.awt.image.BufferedImage;
 import java.io.File;
 import java.net.URI;
 import java.nio.file.Paths;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
-import qupath.lib.common.GeneralTools;
-import qupath.lib.common.LogTools;
-import qupath.lib.images.servers.FileFormatInfo;
-import qupath.lib.images.servers.ImageServer;
-import qupath.lib.images.servers.ImageServerBuilder;
-import qupath.lib.images.servers.FileFormatInfo.ImageCheckType;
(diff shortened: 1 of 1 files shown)


alanocallaghan:
Original approach wasn't right at all, I forgot how the loader method was being used in the extension class.


Small note I don't like about our openslide loading behaviour: if the preference isn't set, we always load from the jar. If it is set, and there's no openslide found there, we may end up using the system openslide, which seems unintuitive to me, but unfortunately is hard to get around because of the JNA search behaviour

alanocallaghan:
I actually dislike the implementation I used here, because the fact that the extension will load the library when checking is only handled with a side effect, because the static block is called when the class is first loaded.

A better solution would be for `OpenSlideExtension::isOpenSlideAvailable()` to check the value of `OpenSlideLoader.isOpenSlideAvailable()`, and if not, try to load it using the preference.

Rylern:
> A better solution would be for `OpenSlideExtension::isOpenSlideAvailable()` to check the value of `OpenSlideLoader.isOpenSlideAvailable()`, and if not, try to load it using the preference.

I agree, but then you will need to call `OpenSlideExtension::isOpenSlideAvailable()` instead of `OpenSlideLoader.isOpenSlideAvailable()` everywhere else in the code. In the current version, `OpenSlideServerBuilder` is using `OpenSlideLoader.isOpenSlideAvailable()`, which won't load

alanocallaghan:
Yeah, the idea would be to funnel all calls through the Extension class method, via JavaDocs at least. There's probably also some way to ensure it by reflection or some such

Or by re-arranging the classes, the loader method could be package-private and the other public, but that's not really ideal either

petebankhead:
> the idea either way is to funnel all calls through the Extension class method

The extension relies upon `QuPathGUI`, whereas an `OpenSlideImageServer` generally doesn't. It feels questionable to me to allow the UI to get involved in loading OpenSlide. Potentially, we could have a cleaner separation of the UI and OpenSlide/image reading, in which case I think loading OpenSlide should very much belong in the latter.

(I realize having preferences in the UI to control how OpenSlide loads complicates things a little... but only a little, I think)

alanocallaghan:
> It feels questionable to me to allow the UI to get involved in loading OpenSlide.

I agree, as I alluded to, the "sensible" way would mean restructuring the code so that the loader and extension are in the same package, but the current design means that doesn't make sense.

> Potentially, we could have a cleaner separation of the UI and OpenSlide/image reading, in which case I think loading OpenSlide should very much belong in the latter.
(I realize having preferences in the UI to control how OpenSlide loads complicates things a little... but only a little, I think)

I suppose the pertinent question right now is: "should a preference set in the UI be used when running with the command line?" I'd say yes, but I can see the argument for no.

If yes, something like this solution makes sense for now, until a redesign.

If no, then just default to the JNA if the extension class isn't involved

alanocallaghan:
Changed to simply try to load, ignoring the preference, in the `buildServer` method, if OpenSlide isn't already loaded, which seems reasonable to me.

Rylern (approved):

petebankhead (approved):
I haven't run it, but I think the logic makes sense.

Issue #1451, opened 2024-01-22
convert-ome returns 0 upon failure


johnyaku:
## IMPORTANT!

This section is **only** for bug reports, which **must** follow the template below.

Please post any other questions to the official QuPath forum at https://forum.image.sc/tags/qupath

If in doubt, use [image.sc](https://forum.image.sc/tags/qupath). This helps us keep things organized.

## Before we begin...

Before submitting your bug report, please check the following:

* [x] I've definitely found a bug (it you're not sure, please use [image.sc](https://forum.image.sc/tags/qupath) instead)
* [x] I've checked https://qupath.github.io for a new release that might already have fixed the issue
* [x] I've checked the [Changelog](https://github.com/qupath/qupath/blob/master/CHANGELOG.md) to see if the bug has already been fixed in the next release
* [x] I've checked for existing GitHub issues describing the same problem

## Bug report

**Describe the bug**
`QuPath convert-ome` returns exit code 0 despite failure.

**To Reproduce**
```
scale=2
tile=1024
comp=ZLIB
input=broken_image.tif
output=broke_omage.ome.tif
QuPath convert-ome $input $output \
    --pyramid-scale=$scale \
    --tile-size=$tile \
    --compression=$comp \
    --parallelize
```
Output (truncated)
```
20:39:42.582 [main] [INFO ] q.l.i.s.b.BioFormatsServerOptions - Setting max Bio-Formats readers to 2
20:39:43.272 [main] [ERROR] loci.formats.tiff.TiffParser - Error reading IFD type at: 14
20:39:43.274 [main] [ERROR] loci.formats.tiff.TiffParser - Error reading IFD type at: 14
20:39:43.275 [main] [ERROR] loci.formats.tiff.TiffParser - Error reading IFD type at: 14
...
20:39:43.604 [main] [ERROR] q.l.i.writers.ome.ConvertCommand - Unable to open file:broken_image.tif
java.io.IOException: Unable to open file:broken_image.tif
        at qupath.lib.images.servers.ImageServers.buildServer(ImageServers.java:306)
        at qupath.lib.images.writers.ome.ConvertCommand.run(ConvertCommand.java:147)
        at picocli.CommandLine.executeUserObject(CommandLine.java:2026)
        at picocli.CommandLine.access$1500(CommandLine.java:148)
        at picocli.CommandLine$RunLast.executeUserObjectOfLastSubcommandWithSameParent(CommandLine.java:2461)
[...]

PR #1455, opened 2024-01-25
Added non-zero return codes

Rylern:
See #1451.

Should I also refactor the class to follow the new style guide?


petebankhead:
Looks good! I'll merge this as it looks small enough to be a good candidate for v0.5.1 (if we make a release).

Feel free to refactor according to the new style as a separate PR, but it's not urgent if it's currently working.

PR #1457, opened 2024-01-30
Added non-empty heterogeneous collection check

Rylern:
In the `calculateLeftRightSharedBoundaryIntersectionScore(Geometry left, Geometry right, double overlapTolerance)` function (see [here](https://github.com/qupath/qupath/blob/13bdeed047b4d05f35f47308b36b48c0f2bb3a24/qupath-core/src/main/java/qupath/lib/objects/utils/ObjectMerger.java#L478)), the `Geometry.intersection(Geometry)` is called without checking that the parameter isn't a non-empty heterogeneous collection (which is not supported by the `intersection` function).

For example:

```java
calculateLeftRightSharedBoundaryIntersectionScore(
	POLYGON ((26629 11368, 26628 11369, 26627 11369, 26626 11370, 26624 11370, 26623 11371, 26622 11371, 26621 11372, 26621 11373, 26619 11375, 26618 11375, 26617 11376, 26617 11382, 26616 11383, 26616 11386, 26615 11387, 26615 11393, 26614 11394, 26614 11395, 26616 11397, 26616 11399, 26617 11400, 26617 11401, 26618 11402, 26618 11404, 26619 11404, 26621 11406, 26621 11407, 26622 11407, 26623 11406, 26626 11406, 26627 11407, 26629 11407, 26630 11406, 26631 11406, 26632 11407, 26637 11407, 26638 11406, 26639 11406, 26640 11405, 26644 11405, 26646 11403, 26646 11402, 26647 11401, 26648 11401, 26648 11400, 26649 11399, 26649 11397, 26650 11396, 26650 11393, 26651 11392, 26650 11391, 26651 11390, 26651 11388, 26651 11384, 26651 11382, 26650 11381, 26650 11380, 26649 11380, 26643 11374, 26643 11373, 26641 11371, 26641 11370, 26640 11369, 26633 11369, 26632 11368, 26629 11368)),
	POLYGON ((26651 11384, 26651 11388, 26652 11387, 26652 11385, 26651 11384)),
	0.125
);
```

throws an `IllegalArgumentException` on the line

```java
var sharedIntersection = rightIntersection.intersection(leftIntersection);
```

because `leftIntersection` is a non-empty heterogeneous collection. 

This PR checks that `leftIntersection` isn't a non-empty heterogeneous collection before calling the `intersection()` function.

### DIFF
--- a/qupath-core/src/main/java/qupath/lib/objects/utils/ObjectMerger.java
+++ b/qupath-core/src/main/java/qupath/lib/objects/utils/ObjectMerger.java
@@ -25,6 +25,7 @@ import org.locationtech.jts.geom.Coordinate;
 import org.locationtech.jts.geom.CoordinateSequence;
 import org.locationtech.jts.geom.CoordinateSequenceFilter;
 import org.locationtech.jts.geom.Geometry;
+import org.locationtech.jts.geom.GeometryCollection;
 import org.locationtech.jts.geom.GeometryFactory;
 import org.locationtech.jts.geom.LineString;
 import org.locationtech.jts.index.SpatialIndex;
@@ -51,6 +52,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.BiPredicate;
 import java.util.function.Function;
 import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 
 /**
  * Helper class for merging objects using different criteria.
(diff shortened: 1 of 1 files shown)


petebankhead:
Thanks, it makes sense to handle this - but I'm not sure if returning 0.0 is desirable.

I can't remember exactly whether that code should only handle lines or areas (I think lines?) but shouldn't we extract the relevant kind of geometry rather than assuming it's 0.0?

And can you specify when this issue occurs?

From what I can remember, our boundary-handling criterion was based on line intersections - because the processed tiles were assumed not to overlap. We probably need boundary-criteria that can operate using area intersections for cases where tiles *do* overlap. I'm not sure if we want to handle both areas and lines cases in this function, or if we want to restrict it to one or the other.

(Quick thoughts as I won't be able to look at this properly for a few days)

Rylern:
> I can't remember exactly whether that code should only handle lines or areas (I think lines?) but shouldn't we extract the relevant kind of geometry rather than assuming it's 0.0?

> And can you specify when this issue occurs?

Based on what I get, `leftIntersection` and `rightIntersection` are either LINESTRING, MULTILINESTRING, or GEOMETRYCOLLECTION with LINESTRINGs and a POINT (which is when the issue occurs).

Instead of returning 0.0, I can keep only the LINESTRING and MULTILINESTRING from the Geometry. Would you prefer that?

petebankhead:
That sounds better to me. Rather than explicitly limit it to lines, you could keep the shape with the highest dimension that you can.

There's a function here that should help: https://github.com/qupath/qupath/blob/13bdeed047b4d05f35f47308b36b48c0f2bb3a24/qupath-core/src/main/java/qupath/lib/roi/GeometryTools.java#L412

Rylern:
OK, I updated the PR

Rylern:
The Javadoc was only requiring the argument to be homogeneous, but the right/upper can also be homogenized

Issue #1458, opened 2024-01-31
M1 github actions runner now available


kaczmarj:
hello friendly qupath team, i wanted to let you all know that github actions now has an apple m1 runner. this might be useful to package qupath on arm macs -- though i see that qupath already has arm releases for apple.

https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/

petebankhead:
Thanks @kaczmarj ! Closing this now as @alanocallaghan has taken care of it in https://github.com/qupath/qupath/pull/1464

PR #1460, opened 2024-02-07
Update jpackage.yml for mac aarch64

petebankhead:
See https://github.com/qupath/qupath/issues/1458

### DIFF
--- a/.github/workflows/jpackage.yml
+++ b/.github/workflows/jpackage.yml
@@ -17,7 +17,9 @@
           - platform: ubuntu-latest
             container: fedora:latest
             type: rpm
-          - platform: macos-latest
+          - platform: macos-12 # x64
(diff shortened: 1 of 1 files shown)


alanocallaghan:
Is it worth revisiting the idea of a draft release pipeline using actions? Manually triggered

petebankhead:
Sounds good, as long as it doesn't enforce any tag until we *actually* release.

Also, if you want to take it on... I think this PR fails because the two mac builds become 'pkg' and only one survives.

alanocallaghan:
Question: why do we upload javadocs for each? I'd have thought these would be the same across platforms/architectures?

petebankhead:
Closed in favor of https://github.com/qupath/qupath/pull/1464

PR #1462, opened 2024-02-09
Histogram not updating fix

### LINKED ISSUE #1459: Brightness/Contrast histogram doesn't always update when switching between similar images
## Bug report

**Describe the bug**
When switching between images with similar channels with the Brightness/Contrast window open, the histogram doesn't automatically update as it should. Rather, the user sometimes need to select a different channel and then return to the one they wanted to look at.

**To Reproduce**
1. Create a project with some similar images (same channels, names)
2. Open Brightness/Contrast
3. Open the first image, see the histogram displayed
4. Open the second image; the histogram from the first probably still lingers

**Expected behavior**
The histogram updates when the image updates.

 - OS: macOS (but presumably all)
 - QuPath Version: 0.5.0

**Additional context**
The 'Apply to similar images' setting doesn't seem to make any difference to this behavior.

The issue isn't evident if the images have different channels.

Rylern:
This PR attempts to fix #1459.

I **replicated** the issue by:
* Creating and opening a project containing `LuCa-7color_[13860,52919]_1x1component_data.tif` (refered as `A`) and `LuCa-7color_[17572,60173]_3x3component_data.tif` (refered as `B`), two different images but with the same channels.
* Opening `A`.
* Opening the brightness & contrast dialog. The histogram of `A` is displayed.
* Opening `B` while keeping the brightness & contrast dialog open. The histogram of `A` is still displayed.

**The issue** comes from the fact that in [ImageDisplay](https://github.com/qupath/qupath/blob/main/qupath-gui-fx/src/main/java/qupath/lib/display/ImageDisplay.java), the function [setImageData()](https://github.com/qupath/qupath/blob/13bdeed047b4d05f35f47308b36b48c0f2bb3a24/qupath-gui-fx/src/main/java/qupath/lib/display/ImageDisplay.java#L202) is called when an image is open in QuPath. This function calls the `updateChannelOptions()` function and then the `updateHistogramMap()` function (see [here](https://github.com/qupath/qupath/blob/13bdeed047b4d05f35f47308b36b48c0f2bb3a24/qupath-gui-fx/src/main/java/qupath/lib/display/ImageDisplay.java#L228C4-L228C24)):
* `updateChannelOptions()` updates the [channelOptions](https://github.com/qupath/qupath/blob/13bdeed047b4d05f35f47308b36b48c0f2bb3a24/qupath-gui-fx/src/main/java/qupath/lib/display/ImageDisplay.java#L504) variable. It's an ObservableValue and, because of listeners, the [BrightnessContrastHistogramPane.updateHistogram()](https://github.com/qupath/qupath/blob/13bdeed047b4d05f35f47308b36b48c0f2bb3a24/qupath-gui-fx/src/main/java/qupath/lib/gui/commands/display/BrightnessContrastHistogramPane.java#L162C17-L162C32) function is called. This is the function responsible for displaying the histogram.
* `updateHistogramMap()` updates the `histogramManager` variable. This variable is responsible for providing the histogram values.

So the problem is that the function displaying the histogram can be called before the function updating the histogram values. It is not possible to simply switch the calls of the `updateChannelOptions()` and `updateHistogramMap()` functions.

**The solution** I found was to make `histogramManager` an ObservableValue. In this case, `BrightnessContrastHistogramPane.updateHistogram()` can be called each time the `histogramManager` value changes.

I don't like this solution as it exposes the `histogramManager` variable (which is an implementation detail). But this was the solution requiring the less amount of refactoring I found.

### DIFF
--- a/qupath-gui-fx/src/main/java/qupath/lib/display/ImageDisplay.java
+++ b/qupath-gui-fx/src/main/java/qupath/lib/display/ImageDisplay.java
@@ -55,6 +55,7 @@ import javafx.beans.property.BooleanProperty;
 import javafx.beans.property.LongProperty;
 import javafx.beans.property.SimpleBooleanProperty;
 import javafx.beans.property.SimpleLongProperty;
+import javafx.beans.property.ReadOnlyObjectProperty;
 import javafx.collections.FXCollections;
 import javafx.collections.ObservableList;
 import javafx.collections.ListChangeListener;
@@ -119,7 +120,7 @@ public class ImageDisplay extends AbstractImageRenderer {
 			useGrayscaleLutProperty(), useInvertedBackgroundProperty());
 
 	private static Map<String, HistogramManager> cachedHistograms = Collections.synchronizedMap(new HashMap<>());
-	private HistogramManager histogramManager = null;
+	private final ObjectProperty<HistogramManager> histogramManager = new SimpleObjectProperty<>(null);
 	private List<BufferedImage> imagesForHistograms = new ArrayList<>(); // Cache images needed to recompute histograms
 
 	private static BooleanProperty showAllRGBTransforms = PathPrefs.createPersistentPreference("showAllRGBTransforms", true);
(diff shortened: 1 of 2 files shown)


petebankhead:
Thanks, this is some excellent detective work!

Closing in favor of https://github.com/qupath/qupath/pull/1476 for v0.5.1 because it is simpler (and written based on the info you posted here), but we may revisit it as this is more listener-friendly.

PR #1463, opened 2024-02-09
Added keyboard shorcuts to tooltips of toolbar

Rylern:


petebankhead (changes requested):
Thanks, I think there's an encoding issue with the properties file.

Would it be possible to avoid duplicating the shortcuts in the properties files, and instead extract them directly from the `Action` accelerator property?

The actions should be converted to buttons etc. via [`ActionTools`](https://github.com/qupath/qupath/blob/13bdeed047b4d05f35f47308b36b48c0f2bb3a24/qupath-gui-fx/src/main/java/qupath/lib/gui/actions/ActionTools.java#L91) - so perhaps some logic could be added there?

A complication I see is that making the tooltip text auto-update if the accelerator or action 'long description' is changed - but this is not an essential feature (since it should be rare).

What do you think?

qupath-gui-fx/src/main/resources/qupath/lib/gui/localization/qupath-gui-strings.properties
```diff
@@ -873,13 +873,13 @@ Prefs.Viewer.scalebarLineWidth.description = Adjust line thickness for the scale
 Prefs.Viewer.locationFontSize = Location text font size
 Prefs.Viewer.locationFontSize.description = Adjust font size for location text
 Prefs.Viewer.calibratedLocationString = Use calibrated location text
-Prefs.Viewer.calibratedLocationString.description = Show pixel locations on the viewer in �m where possible
```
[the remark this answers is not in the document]
Rylern: Yes, Intellij keeps setting the encoding to ISO-8859-1

Rylern:
> Would it be possible to avoid duplicating the shortcuts in the properties files, and instead extract them directly from the Action accelerator property?

Yes, it's now the case.

> A complication I see is that making the tooltip text auto-update if the accelerator or action 'long description' is changed - but this is not an essential feature (since it should be rare).

Right now I have the impression that the tooltip text will auto-update if the language changes. Is that enough?

qupath-gui-fx/src/main/java/qupath/lib/gui/actions/ActionTools.java
```diff
@@ -454,12 +451,27 @@ private static void parseConfig(Action action, ActionConfig annotation, Localize
...
-				if (annotation.bindLocale())
-					localizedResourceManager.registerProperty(action.longTextProperty(), bundle, descriptionKey);
-				else
-					action.setLongText(QuPathResources.getString(bundle, descriptionKey));
+				if (annotation.bindLocale()) {
+					StringProperty resourceLongTextProperty = new SimpleStringProperty();
+					resourceLongTextProperty.addListener((p, o, n) -> {
+						String longText = n;
+						if (action.getAccelerator() != null) {
+							longText = "(" + action.getAccelerator() + ") " + longText;
```
petebankhead, line 462: Should we use the `.toString()` value or the `.getDisplayText()`?

See the difference with the following Groovy script:
```groovy
def combo = javafx.scene.input.KeyCombination.keyCombination("shortcut+c")
println combo
println combo.getDisplayText()
```
which prints
```
Shortcut+C
⌘C
```
On Windows this may be different, with the display text `Ctrl+C` but I haven't checked.

qupath-gui-fx/src/main/resources/qupath/lib/gui/localization/qupath-gui-strings.properties
```diff
@@ -873,13 +873,13 @@ Prefs.Viewer.scalebarLineWidth.description = Adjust line thickness for the scale
...
+Prefs.Viewer.calibratedLocationString.description = Show pixel locations on the viewer in µm where possible
```
petebankhead, line 876: This seems to display ok, but somehow git is picking up that it's different with a weird character.

qupath-gui-fx/src/main/java/qupath/lib/gui/actions/ActionTools.java
```diff
@@ -454,12 +451,27 @@ private static void parseConfig(Action action, ActionConfig annotation, Localize
...
+						}
+						action.setLongText(longText);
+					});
+					localizedResourceManager.registerProperty(resourceLongTextProperty, bundle, descriptionKey);
+				}
+				else {
+					String longText = QuPathResources.getString(bundle, descriptionKey);
+					if (action.getAccelerator() != null) {
```
petebankhead, line 470: Would it be 'cleaner' to get the long text for an `Action` in a method, to avoid the duplication?

petebankhead (commented):
This looks good to me - the main remaining question is to agree on whether to use the `toString()` representation of the `.getDisplayText()` one; I've shown an example above.

@Rylern @alanocallaghan @finglis 

One other non-crucial limitation: changes to accelerators aren't reflected in the tooltips. For example, I can run a Groovy script
```groovy
Platform.runLater {
    def combo = javafx.scene.input.KeyCombination.keyCombination("shortcut+i")
    getQuPath().getCommonActions().PREFERENCES.setAccelerator(combo)
}
```
and on macOS I'll see the updated accelerator in the system menubar, but not in the tooltip.

This is sufficiently obscure that I think we can ignore it, unless you see a simple way to overcome it.

qupath-gui-fx/src/main/java/qupath/lib/gui/actions/ActionTools.java
Rylern: `.getDisplayText()` seems better to me

qupath-gui-fx/src/main/resources/qupath/lib/gui/localization/qupath-gui-strings.properties
Rylern: Yeah I don't know why it's happening

qupath-gui-fx/src/main/java/qupath/lib/gui/actions/ActionTools.java
Rylern: Yes

Rylern:
I addressed the 3 issues

petebankhead (approved):

petebankhead:
Looks great, thanks!

PR #1464, opened 2024-02-09
Update jpackage.yml to create all artifacts and draft a release

alanocallaghan:
- Adds mac arm64 build
- Only uploads javadoc from mac arm64 
- Removes the linux stuff as RPMs weren't usable for snapshot/rc builds.
- Bumps action versions to remove some warnings

See #1460

So workflow would basically be, run the jpackage action, add info about release draft (changelog/etc), people download + test the artifacts from the draft release, when happy make a tag and publish the release


petebankhead:
Looks great - is there any way to run it before merging? I think it's possible if it's on a branch of this repo, but not a fork.

Is it also possible to separate out the draft of the release - or check what happens if running twice without a change of version? I imagine sometimes we might want to make builds without creating a release - but having the option to make a draft release as well is a fantastic improvement.

alanocallaghan:
> Looks great - is there any way to run it before merging? I think it's possible if it's on a branch of this repo, but not a fork.

Here's what it makes on my fork: https://github.com/alanocallaghan/qupath/releases/tag/v0.5.1-SNAPSHOT

If I dispatch it again without making any changes, it just makes another release. There's no issue with tag collision until you try to publish both. I did update the release code so it points at the specific commit SHA that the builds correspond to, though, rather than a moving target (eg `main`).

> Is it also possible to separate out the draft of the release - or check what happens if running twice without a change of version?

Yes, I rearranged the workflows so that jpackage and draft release are separate workflows (though ofc draft release does jpackage first). So you can just build binaries, or build and draft a release.

Any other wishlist ideas for CI while I'm at it?

petebankhead (approved):

Issue #1443, opened 2023-12-20
Full image annotation for Sparse training image throws errors for detections


Ross-AC:
## Bug report

**Describe the bug**
When optimising perameters for a project by creating a sparse composite image from different image entries. When running cell detection using script editor, errors are generated for the final tiles **IF** the entire shape is not a perfect rectangle. If using stardist script for cell detection, errors are generated before any detections.

**Can avoid both by manual annotation** with a polygon tool to follow image boundaries -> so not sure if a crucial bug. Carried over from last release to v5.

**To Reproduce**
1. Ctl+Shift+A
2. Run cell detection as a script:
```
setImageType('FLUORESCENCE');
selectAnnotations();
runPlugin('qupath.imagej.detect.cells.WatershedCellDetection', '{"detectionImage":"DAPI","requestedPixelSizeMicrons":0.1,"backgroundRadiusMicrons":4.0,"backgroundByReconstruction":true,"medianRadiusMicrons":1.0,"sigmaMicrons":2.5,"minAreaMicrons":11.0,"maxAreaMicrons":400.0,"threshold":1000.0,"watershedPostProcess":true,"cellExpansionMicrons":5.0,"includeNuclei":true,"smoothBoundaries":true,"makeMeasurements":true}')
```
4. Error:
```
ERROR: Error running plugin: java.lang.NullPointerException: Cannot invoke "java.awt.image.BufferedImage.getSampleModel()" because "img" is null
java.util.concurrent.ExecutionException: java.lang.NullPointerException: Cannot invoke "java.awt.image.BufferedImage.getSampleModel()" because "img" is null
    at java.base/java.util.concurrent.FutureTask.report(Unknown Source)
    at java.base/java.util.concurrent.FutureTask.get(Unknown Source)
    at qupath.lib.plugins.AbstractTaskRunner.awaitCompletion(AbstractTaskRunner.java:147)
    at qupath.lib.plugins.AbstractTaskRunner.runTasks(AbstractTaskRunner.java:117)
    at qupath.lib.gui.TaskRunnerFX.runTasks(TaskRunnerFX.java:106)
    at qupath.lib.plugins.AbstractPlugin.runPlugin(AbstractPlugin.java:147)
    at qupath.lib.gui.QuPathGUI.runPlugin(QuPathGUI.java:2245)
    at qupath.lib.gui.scripting.QPEx.runPlugin(QPEx.java:248)
[...]

PR #1465, opened 2024-02-12
Make AbstractTileableImageServer.readRegion return an empty tile when AbstractTileableImageServer.readTile returns an empty tile

Rylern:
When reading an **empty** region (I'm not sure of the term, see the first screenshot of #1443) from a [SparseImageServer](https://github.com/qupath/qupath/blob/main/qupath-core/src/main/java/qupath/lib/images/servers/SparseImageServer.java):
* the [SparseImageServer.readTile()](https://github.com/qupath/qupath/blob/13bdeed047b4d05f35f47308b36b48c0f2bb3a24/qupath-core/src/main/java/qupath/lib/images/servers/SparseImageServer.java#L221) function returns an empty `BufferedImage`.
* but the [AbstractTileableImageServer.readRegion()](https://github.com/qupath/qupath/blob/13bdeed047b4d05f35f47308b36b48c0f2bb3a24/qupath-core/src/main/java/qupath/lib/images/servers/AbstractTileableImageServer.java#L266) (which is extended by SparseImageServer) function returns `null`.

This means that calling `SparseImageServer.readRegion()` on empty regions will return `null`, which can triggers errors (see #1443).

This PR makes `AbstractTileableImageServer.readRegion()` return an empty tile (instead of `null`) when `AbstractTileableImageServer.readTile()` returns an empty tile. This should fix the first issue of #1443.


petebankhead:
I think this is a good & elegant way to fix the problem. I haven't merged it yet as I'm trying to think if it could have any other implications that mean we should target v0.6.0 rather than v0.5.1.

Were you able to reproduce the original error? I've tried creating some training images with empty regions, but haven't managed to cause the problem yet.

Rylern:
Yes, in a project, start from LuCa-7color_[17572,60173]_3x3component_data.tif, define two annotations with the Region class like this:
Screenshot from 2024-02-29 15-14-35

Then Classify, Training Images, Create Training images, default parameters, and it creates this image:
image

Then this script throws the error:

```groovy
setImageType('FLUORESCENCE');
selectAnnotations();
runPlugin('qupath.imagej.detect.cells.WatershedCellDetection', '{"detectionImage":"DAPI","requestedPixelSizeMicrons":0.1,"backgroundRadiusMicrons":4.0,"backgroundByReconstruction":true,"medianRadiusMicrons":1.0,"sigmaMicrons":2.5,"minAreaMicrons":11.0,"maxAreaMicrons":400.0,"threshold":1000.0,"watershedPostProcess":true,"cellExpansionMicrons":5.0,"includeNuclei":true,"smoothBoundaries":true,"makeMeasurements":true}')
```

petebankhead:
Great, thanks - can replicate it now. I was trying something similar but the key difference was that I was using an RGB image, and that seems to always work.

Well, the current `ImageServer` interface seems to have evolved into a bit of a mess. A note in the javadocs seems to suggest we shouldn't return `null` anyway: https://github.com/qupath/qupath/blob/474c08c13381c87969b7874ea3e5b4850757c7ba/qupath-core/src/main/java/qupath/lib/images/servers/ImageServer.java#L202

But another note suggests that `isEmptyRegion` can be used to determine when `readBufferedImage` (now `readRegion`) would return `null`:
https://github.com/qupath/qupath/blob/474c08c13381c87969b7874ea3e5b4850757c7ba/qupath-core/src/main/java/qupath/lib/images/servers/ImageServer.java#L314

So... definitely need to fix this for v0.6.0, maybe should already tackle it in v0.5.1.

petebankhead:
I've updated one of my old incorrect javadocs - assuming it still builds fine I think this is good to merge, thanks!

Issue #1397, opened 2023-10-20
Folder based Dicom WSI open as independent dcm images when added to a project


donn-deciphex:
## Bug report

**Describe the bug**
DICOM WSI is a multifile format comprised of a few DCM files linked by metadata within each file to define a complete WSI.  However it is not being treated as one image when in project mode in QuPath.

**To Reproduce**
1. Create a project in QuPath
2. Hit Add images and select the 1 Dicom WSI (all files .dcm from the containing folder need to be added)
3. Observe the one WSI it displays as N .dcm images (each pyramid level and embedded image is displayed separately).

**Expected behavior**
It should display as 1 WSI for every unique Dicom WSI you add, just like SVS or NDPI etc.  

**Screenshots**
None

**Desktop (please complete the following information):**
Version: 0.4.4
Build time: 2023-08-31, 05:46
Latest commit tag: '647f7ac'
Java version:		17.0.8
Operating system:		Windows 11  -  10.0
Architecture:    		amd64

**Additional context**
None

petebankhead:
Are there public example images to replicate this issue?

QuPath v0.5.0 will include OpenSlide 4.0.0, and this provides dicom WSI support. So the behavior may change then - but we have no examples to test with.

donn-deciphex:
You can try these. Just download and unzip.  https://openslide.cs.cmu.edu/download/openslide-testdata/DICOM/

petebankhead:
Thanks, can you try the QuPath v0.5.0 release candidate - https://github.com/qupath/qupath/releases/tag/v0.5.0-rc1 ?

This behaves different from v0.4.4, using OpenSlide to read the image.

To me, it appears to behave properly - although has the limitation that if you add more than one of the dicom files, then the same image is effectively duplicated in the project. I'm not sure if that's expected or acceptable, but addressing it is probably not straightforward. Dicom is the only format I know that behaves in this way, and QuPath tries to avoid having to treat formats as a special case (relying upon OpenSlide or Bio-Formats to untangle the meaning of the image file paths provided).

donn-deciphex:
Thanks for your help, yes that (QuPath v0.5.0 release candidate) seems to be opening the WSI as one image rather than several separate dcm images. However like you said it still displays the list of dcm files in the Image List which may not be desired.  

Not sure what's next in terms of this ticket - the main concern is addressed but the open UX issue may want to be handled separately.

petebankhead:
[...]

PR #1474, opened 2024-02-23
Zarr reading and writing support

Rylern:
This PR adds the possibility to open Zarr images.

This can be done by clicking on File->Open URI..., and writing the path to the Zarr directory. Dragging the directory to QuPath or using File->Open doesn't work. Should I improve this?


alanocallaghan:
Let me know when this is tested on different platforms so I can do a non-snapshot blosc release

Rylern:
It works on Linux, Windows, and Mac with Apple Silicon. So the only remaining OS is Mac on Intel

petebankhead:
This is huge! I should be able to try it on an Intel Mac - is there a link to any good test image?

Rylern:
Yes, here: https://oc.embl.de/index.php/s/0TPPw4CUt0G6nng

petebankhead:
Works for me on an Intel Mac!

> This can be done by clicking on File->Open URI..., and writing the path to the Zarr directory. Dragging the directory to QuPath or using File->Open doesn't work. Should I improve this?

Yeah, we'll need to provide a nicer way to open the images or add to a project - ideally including a drag & drop option.

Will take a bit of thought to figure out how to do it. Maybe we need an 'Open image directory' option? But then people might assume that it opens all the images in a directory (which, I suppose, it something it could do...).

DICOM is also a more recently-relevant format that our current 'open / add' mechanism doesn't support well: https://github.com/qupath/qupath/issues/1397

So we should probably think about both together.

Rylern:
About commit [a7b184f](https://github.com/qupath/qupath/pull/1474/commits/a7b184f2bf0134432c52080e394ab51fea02dfe3):

In Zarr files (or at least in the Zarr files I have), the z values are scaled with the resolution. This means that if there is an image of shape (t: 10, c: 20, z: 30, y:40, x: 50) at full resolution, it will have a shape of (t: 10, c: 20, z: **15**, y:20, x: 25) with a downsample of 2 (and not (t: 10, c: 20, z: **30**, y:20, x: 25)).

This means that in the BioFormats `IFormatReader.getIndex(z, c, t)` function, `z` can only take values between 0 and **15** (and not **30**) when called for the image of downsample 2.

I'm not satisfied with this commit because it's ugly and assumes every Zarr files scale the z values with the resolution (which I was not able to find if it's the case in the [OME-ZARR specs](https://ngff.openmicroscopy.org/latest/)).

petebankhead:
Ah, yes that does seem to be a significant complication... have you seen this? https://github.com/ome/bioformats/pull/3794

If we were able to support 2D pyramidal Zarr in QuPath v0.6.0 that would already be a major step forward (especially if we can write as well) - and we could delay worrying about multiresolution z until we switch to ImgLib2 and (most likely) an entirely new viewer.

alanocallaghan:
Great, shall I push out a proper release of the blosc jar, then?

> Will take a bit of thought to figure out how to do it. Maybe we need an 'Open image directory' option? But then people might assume that it opens all the images in a directory (which, I suppose, it something it could do...).

Both useful, but should be distinguished semantically in some way

Rylern:
> If we were able to support 2D pyramidal Zarr in QuPath v0.6.0 that would already be a major step forward (especially if we can write as well) - and we could delay worrying about multiresolution z until we switch to ImgLib2 and (most likely) an entirely new viewer.

OK, so should I remove the last commit?

> Both useful, but should be distinguished semantically in some way

I may be missing something obvious, but can't we just allow the File->Open window to open files AND directories?

alanocallaghan:
> I may be missing something obvious, but can't we just allow the File->Open window to open files AND directories?

My feeling is that mixing the two behaviours could lead to weird or unexpected behaviour if eg I open a directory that contains zarr(s)

petebankhead:
I think in JavaFX we can only have a file *or* a directory chooser... and you need to select the right chooser since they can't be mixed (at least on macOS - it might be platform-dependent).

Haven't checked out the last commit yet...

gradle/libs.versions.toml
```diff
@@ -2,6 +2,8 @@
 
 bioformats      = "7.0.1"
 bioimageIoSpec  = "0.1.0"
+omeZarrReader   = "0.4.1-SNAPSHOT"
+blosc           = "1.21.5-SNAPSHOT"
```
alanocallaghan, line 6:
```suggestion
blosc           = "1.21.5"
```

Rylern:
I improved the z-stack handling as we discussed yesterday @petebankhead

qupath-extension-bioformats/src/main/java/qupath/lib/images/servers/bioformats/BioFormatsImageServer.java
```diff
@@ -1376,16 +1376,24 @@ BufferedImage openImage(TileRequest tileRequest, int series, int nChannels, bool
...
-					z /= tileRequest.getDownsample();
-				}
 		
 				synchronized(ipReader) {
 					ipReader.setSeries(series);
 					ipReader.setResolution(level);
+
+					// Some files provide z scaling (the number of z stacks decreases when the resolution becomes
+					// lower, like the width and height), so z needs to be updated for levels > 0
+					if (level > 0) {
```
petebankhead, line 1386: I don't know if `ipReader.setResolution` is expensive. In case it is, we should try to avoid calling it.
* Here, I think you can put this block *before* the initial `ipReader.setResolution(level)` and include that in an `else` (since this block set the resolution to the level at the end anyway).
* You can also use `if (level > 0 && z > 0)` since otherwise it shouldn't matter

qupath-extension-bioformats/src/main/java/qupath/lib/images/servers/bioformats/BioFormatsImageServer.java
```diff
@@ -1376,16 +1376,24 @@ BufferedImage openImage(TileRequest tileRequest, int series, int nChannels, bool
...
+						ipReader.setResolution(0);
+						int zStacksFullResolution = ipReader.getSizeZ();
+						ipReader.setResolution(level);
+						int zStacksCurrentResolution = ipReader.getSizeZ();
+
+						if (zStacksFullResolution != zStacksCurrentResolution) {
```
petebankhead, line 1392: Should this be `round` or `floor`? My intuition says `round` but my intuition isn't certain...

Also, it would be useful to look at the number of `z` slices for different levels in the sample images you have access to. It might be possible to compute the ratio between the `z` slices for a level and the zero level, to determine how much to rescale the `z` index (which might be more robust than using the `downsample` value, in case it's possible that the downsampling in `z` uses a different factor).

Rylern: I tried to use `round` but I got an error when `z / tileRequest.getDownsample()` was giving something like 133.5, which turned into 134 after applying `round`, when the max possible value was 133.

Rylern:
I addressed the two issues

PR #1476, opened 2024-03-01
Alternative Histogram not updating fix

petebankhead:
Alternative simple approach (hack) to address https://github.com/qupath/qupath/issues/1459 based on @Rylern's detailed investigations at https://github.com/qupath/qupath/pull/1462


Rylern (approved):
Better than #1462

petebankhead:
Not sure it's better, but quicker for now without changing the API - so if we both think it should solve the issue I'll merge it.

Your insights from exploring this ugly bit of the code are welcome any time for refactoring in a major release :)

Issue #1472, opened 2024-02-23
Grid views don't show objects by default if no measurements are available


finglis:
## Feature request

**Is your feature request related to a problem? Please describe.**
After using the TMA dearrayer, viewing the cores within the TMA grid view wont show cores unless the "show measurements" is toggled on or an order is selected. 

**Describe the solution you'd like**
Would be good if this additional step is removed and the cores are visible within the TMA grid view on opening.

**Describe alternatives you've considered**

**Additional context**

PR #1478, opened 2024-03-01
Fix grid views to show without measurements

petebankhead:
Fixes https://github.com/qupath/qupath/issues/1472 Also speeds up the animations a bit (as they were quite sluggish) and sets a 'No measurements!' placeholder in the combo box when there are no measurements available.


Rylern:
The TMA core grid view doesn't update when the TMA grid is deleted / re-created / modified. Is that intended?

PR #1485, opened 2024-03-04
Linux tar should not be nested

alanocallaghan:
My fault for [trying to write a valid tar from memory](https://xkcd.com/1168/) but the current linux build ends up as a tar.xz that contains the build/dist/QuPath directory, rather than the top level being QuPath


alanocallaghan:
To explain the flags:

-c says create an archive
-C says cd to this directory first
`QuPath` specifies that after cd-ing, archive the QuPath dir
then we pipe into xz and redirect stdout to the tar.xz file

PR #1486, opened 2024-03-19
Improve handling of extension assets

alanocallaghan:
Previously, we just assumed qupath extension releases would follow the convention `qupath-extension-[name]-[version].jar` (or similar, I can't really remember). This version uses the github API to poll for the latest release. If there's only one release, it goes with that. If there's more than one, it presents them as options in a table. Once the user has selected one, it then asks the github API about the assets of that release, and downloads the first matching one, assuming it's not a javadoc or sources jar. If there's more than one, it presents them as options to the user.


petebankhead:
I think this looks good and could be merged any time.

I have a few wishlist requests, which would be added here or in a later PR:
* The popover doesn't match the css, which looks especially strange in dark mode.
   * Using [`WebViews`](https://github.com/qupath/qupath/blob/f12b38a32906e0257a34ea380e9a66d67e58ae75/qupath-gui-fx/src/main/java/qupath/lib/gui/tools/WebViews.java#L38) to create the `WebView` and bind the stylesheet may help, (although I think `PopOver` is generally hard to style in a well-behaved way)
* The `i` info button has different styling when highlighted in the sub-popup window (it basically just ignores whether the row is selected - whichever logic is in the main table might be useful here too)
* The `Owner` and `Repository` fields shouldn't be automatically reset whenever a download is requested or fails with an error
* It feels convoluted to me that the user *must* fill in two fields, whenever they might just want to copy and paste one GitHub url. Could/should the `Owner` and `Repository` be able to identify when a full URL has been pasted, and then automatically split it?
   * e.g. Pasting `https://github.com/qupath/qupath-extension-py4j` would result in `Owner=qupath` and `Repository=qupath-extension-py4j`

Pinging @Rylern and @finglis for their opinions as well.

extension-manager

alanocallaghan:
Re: 

> It feels convoluted to me that the user must fill in two fields

Surely the obvious solution here is to have a single `repo` field that takes eg `qupath/qupath-extension-doesstuff` or the full URL, (git or https?) maybe something else

petebankhead:
Sounds good to me 👍

alanocallaghan:
Re: 

> The i info button has different styling when highlighted in the sub-popup window (it basically just ignores whether the row is selected - whichever logic is in the main table might be useful here too)

I'm not sure I follow? The buttons seem to follow the same display logic in the main extension manager window and the sub-popup. The colours are different here but that's a weird issue when I take screenshots. Buttons in both are dark with white icon/glyph. As far as I can tell the buttons should always be dark background, light glyph.

Screenshot from 2024-07-30 12-27-34
Screenshot from 2024-07-30 12-27-23

petebankhead:
Ah you're right, sorry. I was mixing it up with the appearance whenever the buttons are inactive (because the extension can't be updated). When that is the case, the selection color impacts the button color... but it doesn't if the buttons are active. So the displays are consistent after all.

alanocallaghan:
Think this is sorted now, handles the following cases:

1. `qupath-extension-blah` (we assume it's `qupath/qupath-extension-blah`)
2. `owner/qupath-extension-blah`
3. `(https://)?(www.)?github.com/owner/qupath-extension-blah`
4. `git@github.com:/owner/qupath-extension-blah`

petebankhead:
I'm afraid this isn't working for me - the externalized strings seem to be missing. If I enter
```
qupath/qupath-extension-jpen
```
and agree to the prompts for a download, I end up with an exception. The important part seems to be:
```
Caused by: java.util.MissingResourceException: Can't find resource for bundle java.util.PropertyResourceBundle, key ExtensionControlPane.tag
	at java.base/java.util.ResourceBundle.getObject(ResourceBundle.java:567)
	at java.base/java.util.ResourceBundle.getObject(ResourceBundle.java:561)
	at java.base/java.util.ResourceBundle.getString(ResourceBundle.java:523)
```

I also have the feeling the text field is too small, and it should expand (filling all available space would be ok):

short

If being really picky, I'd also ask for a tiny bit of horizontal space between the label and text field (possibly 5 or so pixels of padding).

Lastly, the tooltip is also misleading - it implies that `qupath-extension-stardist` is a valid repository on its own (although maybe just because I have the old strings).

alanocallaghan:
Sorry about that, seems intellij was having me edit the strings file in the build directory rather than the one in the source directory

petebankhead:
Ah, I've had that experience too.... looks good, thanks!

PR #1492, opened 2024-04-10
fix a bug where RectangleROI would only contain() points sitting on two out of four edges

carlocastoldi:
Essentially the title

the following test fails on the last three assumptions

```java
	@Test
	public void rectangleContains() {
		ROI rectangle = ROIs.createRectangleROI(0,0,1,1, ImagePlane.getDefaultPlane());
		assertTrue(rectangle.contains(.5, .5));
		assertTrue(rectangle.contains(.9999, .9999));
		assertFalse(rectangle.contains(1.0001, 1.0001));
		assertTrue(rectangle.contains(0, 0));
		assertTrue(rectangle.contains(.9999, 0));
		assertTrue(rectangle.contains(0, .9999));
		assertTrue(rectangle.contains(1, 0)); // FAILS
		assertTrue(rectangle.contains(0, 1)); // FAILS
		assertTrue(rectangle.contains(1, 1)); // FAILS
	}
```

This PR fixes this and adds a small test for it

### DIFF
--- a/qupath-core/src/main/java/qupath/lib/roi/RectangleROI.java
+++ b/qupath-core/src/main/java/qupath/lib/roi/RectangleROI.java
@@ -59,7 +59,7 @@ public class RectangleROI extends AbstractPathBoundedROI implements Serializable
 	
 	@Override
 	public boolean contains(double x, double y) {
-		return x >= this.x && x < x2 && y >= this.y && y < y2;
+		return x >= this.x && x <= x2 && y >= this.y && y <= y2;
 	}
 	
 	@Override
--- a/qupath-core/src/test/java/qupath/lib/roi/TestROIs.java
+++ b/qupath-core/src/test/java/qupath/lib/roi/TestROIs.java
@@ -127,6 +127,20 @@ public class TestROIs {
 		}
 		
 	}
+
+	@Test
+	public void rectangleContains() {
+		ROI rectangle = ROIs.createRectangleROI(0,0,1,1, ImagePlane.getDefaultPlane());
+		assertTrue(rectangle.contains(.5, .5));
+		assertTrue(rectangle.contains(.9999, .9999));
+		assertFalse(rectangle.contains(1.0001, 1.0001));
+		assertTrue(rectangle.contains(0, 0));
+		assertTrue(rectangle.contains(.9999, 0));
+		assertTrue(rectangle.contains(0, .9999));
+		assertTrue(rectangle.contains(1, 0));
+		assertTrue(rectangle.contains(0, 1));
+		assertTrue(rectangle.contains(1, 1));
+	}
 	
 	static void checkROIMeasurements(ROI roi, double pixelWidth, double pixelHeight, double delta) {
 		if (roi.isPoint()) {


carlocastoldi:
Oh, i guess this choice was done to uniquely assert in which of two adjacent tiles a point is contained?

carlocastoldi:
It follows this definition of "insideness": https://docs.oracle.com/en/java/javase/17/docs/api/java.desktop/java/awt/Shape.html#def_insideness

petebankhead:
Yes, QuPath originally used Java `Shape` for some of the more complex operations (especially Java's `Area` class).

Now much of that logic has been replaced by Java Topology Suite, which potentially has a different interpretation of insideness - so there could be an inconsistency if you have a `RectangleROI` vs. a `PolygonROI` or `GeometryROI` that happens to also represent a rectangle. Although from a quick check, it looks like the behavior does seem to generally be consistent:
```groovy
def roi = ROIs.createRectangleROI(0,0,1,1, ImagePlane.getDefaultPlane())
println roi.contains(1, 1)

def polygon = ROIs.createPolygonROI(roi.getAllPoints(), roi.getImagePlane())
println polygon.contains(1, 1)

def geometry = GeometryTools.geometryToROI(roi.getGeometry(), roi.getImagePlane())
println geometry.contains(1, 1) // Because it actually returns a RectangleROI, not a GeometryROI
```
I see `false` in all cases there, which fits with my attempts to preserve the same behavior with updating the ROIs to use JTS.

But I see `true` with
```groovy
def geometry2 = new qupath.lib.roi.GeometryROI(roi.getGeometry(), roi.getImagePlane())
println geometry2.contains(1, 1)
```
because it's using JTS logic entirely. However, that is actually calling a package private method - shouldn't happen in normal use, because usually you'd rely upon `GeometryTools.geometryToROI()` and this returns a `RectangleROI` where possible..

I'm a bit apprehensive about changing the behavior, since it has been preserved across versions and is consistent with at least *some* definition of 'contains'.

If you meed something more rigorously defined, you should use `roi.getGeometry()` and then you can use the methods of [`Geometry`](https://locationtech.github.io/jts/javadoc/) and the full [DE-9IM Intersection Matrix](https://en.wikipedia.org/wiki/DE-9IM).

carlocastoldi:
That is a very interesting read. I am indeed reading about the DE-9IM interesection matrix, thank you for the detailed reply.
Initially I was rather afraid of an inconsistent behaviour of which I really didn't understand much. I wouldn't worry about the private method, tho. I'm closing this PR.

However, now it's a lot more clear but I guess the `ROI` documentation could be a little more explicit on its behaviour. Would you accept a PR that simply states that the what the code assumes as "inside"? I know you said that _potentially_ it  has a different interpretation of insideness, but perhaps having a general rule is better than nothing


Just so that you know: I was interested in testing if a point outside of a ROI's bounding box is **always** outside of the ROI itself. Where the bounding box is defined as:
```java
new Rectangle2D.Double(roi.getBoundsX(), roi.getBoundsY(), roi.getBoundsWidth(), roi.getBoundsHeight())
```

petebankhead:
I suspect that JTS's approach is preferable, and the tests here *should* pass - it does seem weird that they don't.

We have two QuPath grants - one to maintain the software (leading to v1.0), and one to rewrite large parts (leading to v2.0 within the next few years). My instinct is to keep the current behavior for v1.0, but look to change it in the longer term for v2.0 - along with a more thorough revision of ROIs generally.

carlocastoldi:
That's awesome. Thank you!

I will try to implement it with JTS, then.
However the implication is that "Polygons do not contain their boundary", correct? Does that mean that if I'm testing if a point `p` is inside the boundary or **on** the boundary, i should check that `JVT.Geometry.covers(p) || JVT.Geometry.intersects(p)`, right?

I'm saying this because of what the DE-9IM article says about the `contains`/`withIn` predicates
> This issue is caused by the final clause of the Contains definition above: "at least one point of the interior of B lies in the interior of A". For this case, the predicate Covers has more intuitive semantics (see definition), avoiding boundary considerations.

carlocastoldi:
Thank you a lot for your time. I just learnt I still have lots to learn.
Looking forward to seeing how v2.0 implementation ^^

PR #1502, opened 2024-04-24
Add IoU merger

alanocallaghan:
Simple merger that checks if IoU is higher than the given threshold.

I also implemented an OutputHandler that does something like this:

1. create objects within the regionRequest
2. remove any objects within N pixels of the boundary of the regionRequest
3. mask objects based on the parent annotation (not the proxy/tile, as is usually done).

Since these are probably useful together, should I add this OutputHandler to this PR?

### DIFF
--- a/qupath-core/src/main/java/qupath/lib/objects/utils/ObjectMerger.java
+++ b/qupath-core/src/main/java/qupath/lib/objects/utils/ObjectMerger.java
@@ -382,11 +382,13 @@ public class ObjectMerger {
 
     private static BiPredicate<Geometry, Geometry> createIoUMergeTest(double iouThreshold) {
         return (geom, geomOverlap) -> {
-            double union = geomOverlap.union(geomOverlap).getArea();
+            var i = geom.intersection(geomOverlap);
+            var intersection = i.getArea();
+            double union = geom.getArea() + geomOverlap.getArea() - intersection;
             if (union == 0) {
                 return false;
             }
-            return geom.intersection(geomOverlap).getArea() / union > iouThreshold;
+            return intersection / union > iouThreshold;
         };
     }
 
--- a/qupath-core/src/main/java/qupath/lib/objects/utils/ObjectMerger.java
+++ b/qupath-core/src/main/java/qupath/lib/objects/utils/ObjectMerger.java
@@ -371,7 +371,7 @@ public class ObjectMerger {
      * tile/region request.
      * You should probably also remove any objects that touch the regionRequest boundaries, as these will probably be
      * clipped, and merging them will result in weirdly shaped detections.
-     * @param iouThreshold Intersection over union threshold; any pairs with values above this are merged.
+     * @param iouThreshold Intersection over union threshold; any pairs with values greater than or equal to this are merged.
      * @return an object merger that can merge together any objects with sufficiently high IoU and the same classification
      */
     public static ObjectMerger createIoUMerger(double iouThreshold) {
--- a/qupath-core/src/main/java/qupath/lib/objects/utils/ObjectMerger.java
+++ b/qupath-core/src/main/java/qupath/lib/objects/utils/ObjectMerger.java
@@ -359,6 +359,36 @@ public class ObjectMerger {
                 0);
     }
 
+    /**
+     * Create an object merger that can merge together any objects with sufficiently large intersection over union.
+     * <p>
+     * Objects must also have the same classification and be on the same image plane to be mergeable.
+     * <p>
+     * IoU is calculated using Java Topology Suite intersection, union, and getArea calls.
+     * <p>
+     * This merger assumes that you are using an OutputHandler that doesn't clip to tile boundaries (only to region
+     * requests) and that you are using sufficient padding to ensure that objects are being detected in more than on
+     * tile/region request.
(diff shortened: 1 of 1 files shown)


qupath-core/src/main/java/qupath/lib/objects/utils/ObjectMerger.java
```diff
@@ -359,6 +359,38 @@ public static ObjectMerger createTouchingMerger() {
...
+     * tile/region request.
+     * You should probably also remove any objects that touch the regionRequest boundaries, as these will probably be
+     * clipped, and merging them will result in weirdly shaped detections.
+     * @return an object merger that can merge together any objects with sufficiently high IoU and the same classification
+     */
+    public static ObjectMerger createIoUMerger(double iouThreshold) {
+        return new ObjectMerger(
+                ObjectMerger::sameClassTypePlaneTest,
+                createIoUMergeTest(iouThreshold),
+                0.0625);
```
petebankhead, line 380: Do you have any idea what I was thinking when I introduced the use of this value?

qupath-core/src/main/java/qupath/lib/objects/utils/ObjectMerger.java
```diff
@@ -359,6 +359,38 @@ public static ObjectMerger createTouchingMerger() {
...
+
+    private static BiPredicate<Geometry, Geometry> createIoUMergeTest(double iouThreshold) {
+        return (geom, geomOverlap) -> {
+            var i = geom.intersection(geomOverlap);
+            var intersection = i.getArea();
+            double union = geom.getArea() + geomOverlap.getArea() - intersection;
+            if (union == 0) {
+                return false;
+            }
+            return intersection / union > iouThreshold;
```
petebankhead, line 391: Probably `>=` is more intuitive?

qupath-core/src/main/java/qupath/lib/objects/utils/ObjectMerger.java
```diff
@@ -359,6 +359,38 @@ public static ObjectMerger createTouchingMerger() {
...
+    public static ObjectMerger createIoUMerger(double iouThreshold) {
+        return new ObjectMerger(
+                ObjectMerger::sameClassTypePlaneTest,
+                createIoUMergeTest(iouThreshold),
+                0.0625);
+    }
+
+    private static BiPredicate<Geometry, Geometry> createIoUMergeTest(double iouThreshold) {
+        return (geom, geomOverlap) -> {
+            var i = geom.intersection(geomOverlap);
```
petebankhead, line 385: I think `intersection` can be slow - could we benchmark this (ideally through adding tests)?
Since we don't *really* need to get the intersection, I imagine we could use `difference` or `union` instead, and compute the areas from that. But I don't know which of the methods is fastest.

qupath-core/src/main/java/qupath/lib/objects/utils/ObjectMerger.java
alanocallaghan: No, but it seemed like a perfectly reasonable magic number so I stuck with it

qupath-core/src/main/java/qupath/lib/objects/utils/ObjectMerger.java
alanocallaghan: I started with `a.intersection(b).getArea() / a.union(b).getArea()` and the profiler was claiming that union was way more expensive than intersection, but I can profile the different options

petebankhead: My exceedingly fast and non-systematic test using two QuPath annotations suggests that `intersection` is indeed the fastest - less than half the time of the other two.

```groovy
def geoms = getAnnotationObjects().collect {it.ROI.geometry}
def g1 = geoms[0]
def g2 = geoms[1]

var timeit = new Timeit()
    .microseconds()
    .checkpointAndRun("Checking", () -> {
        g1.intersection(g2)
//        g1.difference(g2)
//        g1.union(g2)
    }, 1000)
    .summarizeCheckpoints()
    .stop();


println(timeit)
```

So.... as you were.

qupath-core/src/main/java/qupath/lib/objects/utils/ObjectMerger.java
petebankhead: Any thoughts?
Is `>` preferable to ensure only one possible match if the threshold is 0.5...?
Either way, I think the parameter needs documented so it's clear which rule is applied.

alanocallaghan: Probably >= is more intuitive yeah, and should be consistent with the other thresholds

qupath-core/src/main/java/qupath/lib/objects/utils/ObjectMerger.java
```diff
@@ -371,6 +371,7 @@ public static ObjectMerger createTouchingMerger() {
      * tile/region request.
      * You should probably also remove any objects that touch the regionRequest boundaries, as these will probably be
      * clipped, and merging them will result in weirdly shaped detections.
+     * @param iouThreshold Intersection over union threshold; any pairs with values above this are merged.
```
petebankhead, line 374: To me this implies `>` :)

PR #1504, opened 2024-04-29
Use parallel merger

alanocallaghan:
When merge tests are relatively expensive and there's a lot of objects, this provides a substantial (50% for me) improvement. Uses the ForkJoinPool.commonPool by default so won't spawn new threads (in theory).

### DIFF
--- a/qupath-core/src/main/java/qupath/lib/objects/utils/ObjectMerger.java
+++ b/qupath-core/src/main/java/qupath/lib/objects/utils/ObjectMerger.java
@@ -53,6 +53,7 @@ import java.util.concurrent.ConcurrentLinkedDeque;
 import java.util.function.BiPredicate;
 import java.util.function.Function;
 import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 
 /**
  * Helper class for merging objects using different criteria.
@@ -190,13 +191,14 @@ public class ObjectMerger {
                     else
                         allPotentialNeighbors = allObjects;
                     var neighbors = filterCompatibleNeighbors(current, allPotentialNeighbors);
+                    // alreadyVisited is not concurrent, so do this serially
                     var addable = neighbors.stream()
-                            .filter(neighbor -> !alreadyVisited.contains(neighbor)) // alreadyVisited (set) is not thread-safe
-                            .parallel()
+                            .filter(neighbor -> !alreadyVisited.contains(neighbor)).toList();
+                    addable = addable.parallelStream()
                             .filter(neighbor -> mergeTest.test(
                                     currentGeometry,
-                                    getGeometry(neighbor, geometryMap)));
-                    pending.addAll(addable.toList());
+                                    getGeometry(neighbor, geometryMap))).toList();
+                    pending.addAll(addable);
                 }
             }
             if (cluster.isEmpty()) {
@@ -207,7 +209,6 @@ public class ObjectMerger {
         }
         return clusters;
     }
-
 
     /**
      * Recursively build a cluster of objects that can be merged.
--- a/qupath-core/src/main/java/qupath/lib/objects/utils/ObjectMerger.java
+++ b/qupath-core/src/main/java/qupath/lib/objects/utils/ObjectMerger.java
@@ -49,6 +49,7 @@ import java.util.Objects;
 import java.util.Queue;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentLinkedDeque;
 import java.util.function.BiPredicate;
 import java.util.function.Function;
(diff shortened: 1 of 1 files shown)


petebankhead:
I think the parallelisation makes sense. When moving to a stream, shouldn't we use `filter` *at least* once, rather than rely on `forEach`?

(If looks like `filter` could be used twice and the results added to the `pending` collection - avoiding the need to make it concurrent - but not sure if that's cleaner or not).

qupath-core/src/main/java/qupath/lib/objects/utils/ObjectMerger.java
```diff
@@ -189,15 +190,15 @@ private List<List<PathObject>> computeClustersIterative(Collection<? extends Pat
                     else
                         allPotentialNeighbors = allObjects;
                     var neighbors = filterCompatibleNeighbors(current, allPotentialNeighbors);
-                    for (var neighbor : neighbors) {
+                    neighbors.parallelStream().forEach(neighbor -> {
                         if (!alreadyVisited.contains(neighbor)) {
```
Rylern, line 194: `alreadyVisited` is not thread-safe, so I would move this line outside of the concurrent block (which shouldn't have any real performance issue?)

alanocallaghan:
Both good points, really the only necessary change is a parallel filter op

qupath-core/src/main/java/qupath/lib/objects/utils/ObjectMerger.java
```diff
@@ -189,15 +190,13 @@ private List<List<PathObject>> computeClustersIterative(Collection<? extends Pat
...
+                            .filter(neighbor -> !alreadyVisited.contains(neighbor)) // alreadyVisited (set) is not thread-safe
+                            .parallel()
+                            .filter(neighbor -> mergeTest.test(
                                     currentGeometry,
-                                    getGeometry(neighbor, geometryMap))) {
-                                pending.add(neighbor);
-                            }
-                        }
-                    }
+                                    getGeometry(neighbor, geometryMap)));
```
Rylern, line 198: Here, the line `.filter(neighbor -> !alreadyVisited.contains(neighbor))` is still executed concurrently. For example, the code:

```java
var integers = IntStream.range(0, 10)
        .boxed()
        .peek(i -> System.err.println(Thread.currentThread()))
        .parallel()
        .toList();
System.err.println(integers);
```

prints:

```
Thread[main,5,main]
Thread[ForkJoinPool.commonPool-worker-2,5,main]
Thread[ForkJoinPool.commonPool-worker-5,5,main]
Thread[ForkJoinPool.commonPool-worker-1,5,main]
Thread[ForkJoinPool.commonPool-worker-7,5,main]
Thread[main,5,main]
Thread[ForkJoinPool.commonPool-worker-3,5,main]
Thread[ForkJoinPool.commonPool-worker-1,5,main]
Thread[ForkJoinPool.commonPool-worker-5,5,main]
Thread[ForkJoinPool.commonPool-worker-4,5,main]
[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
```

This is because `parallel()` is an [intermediary operation](https://docs.oracle.com/javase/8/docs/api/java/util/stream/package-summary.html#StreamOps), so it makes the whole stream parallel (not only what happens after its call).

petebankhead: Interesting! I guess this monstrosity overcomes that...?
```java
var integers = IntStream.range(0, 10)
        .boxed()
        .filter(i -> true)
        .toList()
        .stream()
        .parallel()
        .toList();
System.err.println(integers);
```

Rylern: Yes, because `toList()` is a terminal operation.

alanocallaghan: That's unintuitive, though the justifications make sense. Should be fixed now

Rylern: The reason is explained in the [Javadocs](https://docs.oracle.com/javase/8/docs/api/java/util/stream/package-summary.html#StreamOps):

>  Processing streams lazily allows for significant efficiencies; in a pipeline such as the filter-map-sum example above, filtering, mapping, and summing can be fused into a single pass on the data, with minimal intermediate state. Laziness also allows avoiding examining all the data when it is not necessary; for operations such as "find the first string longer than 1000 characters", it is only necessary to examine just enough strings to find one that has the desired characteristics without examining all of the strings available from the source. (This behavior becomes even more important when the input stream is infinite and not merely large.)

Rylern (approved):

PR #1509, opened 2024-04-30
Use logviewer and qupath-fxtras snapshots

petebankhead:
Temporarily during development, until we make the final releases.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -36,12 +36,12 @@ jts             = "1.19.0"
 junit           = "5.9.2"
 
 logback         = "1.3.11"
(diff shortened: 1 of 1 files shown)


petebankhead:
@alanocallaghan @Rylern @finglis shall we merge this for now, then remember to make non-SNAPSHOT releases before v0.6.0?

Otherwise we need to test things by using `includeBuild` for `qupath-fxtras` and `log-viewer`, or else commit to new releases for them now (but there may be more changes to come).

alanocallaghan:
Probs makes sense to allow co-development around releases, so snapshot makes sense to me

PR #1510, opened 2024-05-01
UX - Export measurements resizable UI 

finglis:
Previously the export measurements window was a fixed size resulting in a small controlsfx ListSelectionView component. The interface is now resizable with the list boxes flexing to fill the space.


petebankhead:
Looks much better, thanks!

PR #1517, opened 2024-05-03
Make non-collapsible titledpanes not highlight on mouseover 

alanocallaghan:


finglis:
In the process of adding the class to other titled panes, please don't merge yet :)

alanocallaghan:
That version of `simplifyTitledPane` was only removed/deprecated a few months ago so I'd be inclined to let it linger until 1.0 at least https://github.com/qupath/qupath/commit/a728924ca0b54f0070801dd3f2d6bbd700bf60a9

petebankhead:
If I understand correctly, this PR is to make the top title change on hover but *not* the bottom one - because the top one can be expanded/collapsed, but the bottom one can't.

https://github.com/qupath/qupath/assets/4690904/d115720b-3ace-41b2-b7e7-017cb21cc4c7

tbh I'd never noticed this or seen it as problematic.

If you think it needs a fix, then should it not go into `qupath-fxtras`? This is the new home to [`simplifyTitledPane`](https://github.com/qupath/qupath-fxtras/blob/4a88b9b427ae6c9a17c33702c06f0c536cad6915/src/main/java/qupath/fx/utils/FXUtils.java#L550)... but then you'd need to load an external .css (like with the simplify method), e.g.
```java
public void makeNonCollapsible(TitledPane pane) {
    pane.setCollapsible(false);
    // Whatever other styling is needed here
}
```

Hard-coding a reference to the CSS class feels a bit brittle to me, and is tied very much to QuPath in a way that is unusable elsewhere. And I imagine we'll end up with inconsistencies as we'll forget / extension writers won't know to add this style class... which to me seems potentially worse.

You could also change `main.css` to avoid any change on hover at all, but personally I think it looks quite nice and helps titles stand out.

Was there any particular user complaint or confusion caused by the default JavaFX approach of slightly changing the behavior on hover, regardless of the 'collapsible' status?

alanocallaghan:
> tbh I'd never noticed this or seen it as problematic.

Fair yeah, it's been bugging me, but I am happy to allow it to continue to bug me if it's not consequential or shared by others

> You could also change main.css to avoid any change on hover at all, but personally I think it looks quite nice and helps titles stand out.

It certainly helps for collapsible panes

alanocallaghan:
I'm going to close this unless somebody complains, sorry for getting you onboard in my crusade

PR #1518, opened 2024-05-06
Remove run task dependency on getJavadocs

Rylern:
See https://github.com/qupath/qupath/pull/1513#issuecomment-2096277248

### DIFF
--- a/qupath-app/build.gradle
+++ b/qupath-app/build.gradle
@@ -108,11 +108,11 @@ dependencies {
 
 
 /**
- * Retrieve Javadocs of dependencies and sub-projects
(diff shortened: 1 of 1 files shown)


petebankhead:
Excellent, thanks - can confirm that
```
./gradlew clean run
```
doesn't include the javadocs, but
```
./gradlew clean getJavadocs run
```
does. And they are also included with
```
./gradlew clean jpackage
```
Only minor comment is that `getJavadocs` doesn't seem a very gradle-ish task name and isn't so descriptive (use `gradlew tasks` to see all those that are available).  What do you think of
```
assembleJavadocs Copies the Javadoc jars to a directory for access within QuPath
```
?

PR #1519, opened 2024-05-07
Switched to new javadoc viewer

Rylern:
When using the new Javadoc viewer within QuPath, the redirection issue described on https://github.com/qupath/qupath/pull/1513#issuecomment-2095553840 still happens. I'm going to see how to fix this.


Rylern:
> When using the new Javadoc viewer within QuPath, the redirection issue described on https://github.com/qupath/qupath/pull/1513#issuecomment-2095553840 still happens. I'm going to see how to fix this.

Fixed in the last commit

petebankhead:
What's the correct way to run this?

I tried
```
./gradlew clean assembleJavadocs run
```
but I'm seeing 
```
No javadocs could be found - sorry!
```
(in a serif font that really ought to be sans-serif)

Rylern:
I only checked that the Javadocs were found when creating a jpackage. The logic to find Javadocs around the executable is present in the javadoc-viewer library, so when running QuPath with Intellij, it searches around the javadoc-viewer library, not around the QuPath executable. I will move this logic from the javadoc-viewer library to QuPath.

> (in a serif font that really ought to be sans-serif)

I forgot to link the QuPath stylesheet to the `WebView`, I'll also add this.

Rylern:
I fixed the two things. I updated the snapshot on maven.scijava.org so Intellij might not directly see it.

qupath-gui-fx/src/main/java/qupath/lib/gui/JavadocViewer.java
```diff
@@ -1,695 +1,110 @@
-/*-
```
petebankhead, line 1: We should keep the copyright statement in every file that's part of core QuPath. If it has been completely rewritten, then we can set the year to be the current year. Otherwise we can use 2022-2024.

petebankhead (changes requested):
Just one small comment on the missing copyright at the top of a file

qupath-gui-fx/src/main/java/qupath/lib/gui/JavadocViewer.java
Rylern: Added in last commit

petebankhead (approved):

PR #1520, opened 2024-05-09
Improve ContourTracing performance

petebankhead:
This makes `ContourTracing` much faster in some circumstances - especially when labeled images were involved.

There are two main differences:
1. Previously, the code would tend to loop through all pixels to generate the contour for each label; now it generates bounding boxes in a single loop, and then iterates only through the relevant bounding box for each label (when using `createROIs`)
2. The tracing algorithm has been replaced by a reliance on JTS's `Polygonizer` class; this was already used before, but now we rely on it more heavily.

The second of these makes this *slightly* risky - I encountered exceptions and infinite loops along the way, so this needs to be tested very carefully.

### DIFF
--- a/qupath-core/src/main/java/qupath/lib/analysis/images/ContourTracing.java
+++ b/qupath-core/src/main/java/qupath/lib/analysis/images/ContourTracing.java
@@ -36,6 +36,7 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.SequencedCollection;
 import java.util.TreeMap;
 import java.util.TreeSet;
 import java.util.concurrent.ExecutionException;
@@ -52,11 +53,14 @@ import java.util.stream.Stream;
 import javax.imageio.ImageIO;
 
 import org.locationtech.jts.geom.Coordinate;
+import org.locationtech.jts.geom.CoordinateSequence;
+import org.locationtech.jts.geom.Coordinates;
 import org.locationtech.jts.geom.Envelope;
 import org.locationtech.jts.geom.Geometry;
 import org.locationtech.jts.geom.GeometryFactory;
 import org.locationtech.jts.geom.LineString;
 import org.locationtech.jts.geom.Polygon;
+import org.locationtech.jts.geom.impl.CoordinateArraySequence;
 import org.locationtech.jts.geom.util.AffineTransformation;
 import org.locationtech.jts.geom.util.PolygonExtracter;
 import org.locationtech.jts.index.quadtree.Quadtree;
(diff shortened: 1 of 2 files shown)


qupath-core/src/main/java/qupath/lib/analysis/images/ContourTracing.java
```diff
@@ -1254,395 +1277,100 @@ private static boolean selected(double v, double min, double max) {
...
+		if (envelope != null) {
+			xStart = Math.max(xStart, (int)Math.floor(envelope.getMinX()-1));
+			yStart = Math.max(yStart, (int)Math.floor(envelope.getMinY()-1));
+			xEnd = Math.min(xEnd, (int)Math.ceil(envelope.getMaxX())+1);
+			yEnd = Math.min(yEnd, (int)Math.ceil(envelope.getMaxY())+1);
+		}
+
+		List<LineString> lines = new ArrayList<>();
+		Coordinate lastHorizontalEdgeCoord = null;
+		Coordinate[] lastVerticalEdgeCoords = new Coordinate[xEnd-xStart+1];
```
alanocallaghan, line 1311: Is there a reason why the horizontal and vertical coords are treated differently here (one sequence stored, one not)?

alanocallaghan:
Speed improvement is really noticable! 

There is a non-zero chance that I'm misunderstanding this, but I strongly dislike this if/else block. First, I'd prefer to get the short block out of the way first by an early return. 

https://github.com/qupath/qupath/blob/a9e4c5c05ac9dae267162705b465b50e307a3ba4/qupath-core/src/main/java/qupath/lib/analysis/images/ContourTracing.java#L608-L647

Second, is this for loop doing anything? 

https://github.com/qupath/qupath/blob/a9e4c5c05ac9dae267162705b465b50e307a3ba4/qupath-core/src/main/java/qupath/lib/analysis/images/ContourTracing.java#L642-L646

If `maxLabel <= minLabel`, then presumably `maxLabel == minLabel`, and either way 
`for (int i = minLabel; i <= maxLabel; i++) {` is at most going to do one iteration.

petebankhead:
Thanks! I've just pushed some changes that I was working on while you were reviewing... since it basically rewrites `createROIs` to make it simpler and less a hack-y adjustment of the original, I think it addresses your last comments.

qupath-core/src/main/java/qupath/lib/analysis/images/ContourTracing.java
petebankhead: Yes, we need to know if the last pixel was an edge so that we can continue/end the edge. That's easy when traversing horizontally, but we need to store for an entire row when traversing vertically.

My original implementation looped through all the pixels twice (horizontally then vertically) so then no array was required, but it seemed to be doing unnecessary work - and to therefore be a bit slower.

petebankhead:
The `for (var entry : envelopes.entrySet())` loop in `createROIs` should be easily parallelizable - I'll check if that helps as well.

I was having trouble with https://github.com/locationtech/jts/issues/874 but these problems seem to have disappeared... slightly concerningly as I'm not entirely sure what I changed to make them disappear. That's my main worry about this PR.

qupath-core/src/main/java/qupath/lib/analysis/images/ContourTracing.java
```diff
@@ -588,38 +591,63 @@ public static Map<Number, ROI> createROIs(Raster raster, int band, RegionRequest
...
 	public static Map<Number, ROI> createROIs(SimpleImage image, RegionRequest region, int minLabel, int maxLabel) {
-		// Check how many labels are needed
-		float[] pixels = SimpleImages.getPixels(image, true);
-		if (maxLabel < minLabel) {
-			float maxValue = minLabel;
-			for (float p : pixels) {
-				if (p > maxValue)
-					maxValue = p;
+		var envelopes = new HashMap<Number, Envelope>();
+		if (minLabel != maxLabel) {
```
alanocallaghan, line 595: Can we invert this if/else? Keeps the start of the two branches closer to the relevant condition

qupath-core/src/main/java/qupath/lib/analysis/images/ContourTracing.java
alanocallaghan: Ah okay, yeah that makes sense

alanocallaghan:
Yeah, I didn't come across that myself yet trying a few images, are we using a version of JTS that at least has the exception code to avoid infinite loops?

Aside from style it all looks good and correct to me, but I wouldn't necessarily use that as strong evidence that it is :)

qupath-core/src/main/java/qupath/lib/analysis/images/ContourTracing.java
petebankhead: I don't mind. I thought it was easier to read to show the use of envelopes before showing their non-use in a special case.

petebankhead:
As far as I can tell, there isn't a version of JTS that has the exception bit merged - just a commit on a fork. An infinite loop was what I was seeing, unfortunately.

Was wondering if we should reinstate the old code and activate it via a system properly, but since the old code *also* uses `Polygonizer` it might not be immune from the problem anyway.

Issue #1522, opened 2024-05-10
ImageJ macro runner can't transfer both cell nuclei & boundaries


petebankhead:
## Bug report

**Describe the bug**
See https://forum.image.sc/t/unexpected-behavior-difference-in-imagej-macro-runner-vs-send-region-to-imagej/96089

**Expected behavior**
It's possible to transfer the cell nucleus and cell boundary to ImageJ for use with the macro runner.

*Possibly* this should also pay attention to the `OverlayOptions` in the viewer by default... but this is questionable as it introduces a dependency on the viewer settings.

**Desktop (please complete the following information):**
 - OS: All
 - QuPath Version: v0.5.1 (and earlier)

**Additional context**
The issue relates to `OverlayOptions` not being passed through, as shown at https://github.com/qupath/qupath/blob/5066159c6e4fa197c28d003662798dbf08804f08/qupath-extension-processing/src/main/java/qupath/imagej/gui/ImageJMacroRunner.java#L238

petebankhead:
Closing this in favor of using the new ImageJ script runner.

PR #1523, opened 2024-05-10
Send cells to ImageJ macro runner

petebankhead:
A step towards fixing https://github.com/qupath/qupath/issues/1522

Incomplete, because it behaves inconsistently depending upon whether the macro is being run on a selected object or across all annotations / TMA cores. In the first case, the `ImageDisplay` and `OverlayOptions` may be used, but in the second case they are unavailable.


petebankhead:
Closing this in favor of using the new ImageJ script runner instead.

PR #1524, opened 2024-05-17
Add "prompt to close" method to script editor interface

alanocallaghan:
This means that when a user attempts to close QuPath, we can check if there are any open scripts with unsaved changes and handle these as with other types of unsaved changes.

Unsure if changing the editor interface is a big deal, presumably there aren't too many implementations out there

Raised by @finglis

### DIFF
--- a/qupath-gui-fx/src/main/java/qupath/lib/gui/QuPathGUI.java
+++ b/qupath-gui-fx/src/main/java/qupath/lib/gui/QuPathGUI.java
@@ -1015,13 +1015,13 @@ public class QuPathGUI {
 		}
 
 		// should prompt users to save changes if desired.
-		if (!scriptEditor.promptToClose()) {
+		if (!scriptEditor.requestClose()) {
 			e.consume();
 			return;
 		}
 		
 		// Warn if there is a script running
-		if (scriptRunning.get()) {
+        if (scriptRunning.get()) {
 			if (!Dialogs.showYesNoDialog("Quit QuPath", "A script is currently running! Quit anyway?")) {
 				logger.trace("Pressed no to quit window with script running!");
 				e.consume();
--- a/qupath-gui-fx/src/main/java/qupath/lib/gui/QuPathGUI.java
+++ b/qupath-gui-fx/src/main/java/qupath/lib/gui/QuPathGUI.java
@@ -1015,7 +1015,10 @@ public class QuPathGUI {
 		}
 
 		// should prompt users to save changes if desired.
-		scriptEditor.promptToClose();
+		if (!scriptEditor.promptToClose()) {
+			e.consume();
+			return;
+		}
 		
 		// Warn if there is a script running
 		if (scriptRunning.get()) {
--- a/qupath-gui-fx/src/main/java/qupath/lib/gui/QuPathGUI.java
+++ b/qupath-gui-fx/src/main/java/qupath/lib/gui/QuPathGUI.java
@@ -1013,6 +1013,9 @@ public class QuPathGUI {
 				return;
 			}
 		}
+
+		// should prompt users to save changes if desired.
+		scriptEditor.promptToClose();
 		
 		// Warn if there is a script running
 		if (scriptRunning.get()) {
--- a/qupath-gui-fx/src/main/java/qupath/lib/gui/scripting/DefaultScriptEditor.java
+++ b/qupath-gui-fx/src/main/java/qupath/lib/gui/scripting/DefaultScriptEditor.java
@@ -166,7 +166,19 @@ public class DefaultScriptEditor implements ScriptEditor {
 	private SplitPane splitMain;
 	private ToggleGroup toggleLanguages = new ToggleGroup();
 	private Font fontMain = Font.font("Courier");
-	
+
+	/**
+	 * Flag that quit was cancelled.
+	 * This is a workaround for the fact that (on macOS at least) Cmd+Q triggers a close request on all windows,
+	 * which can't be cancelled.
+	 * If the user cancels quitting because of unsaved changes, the window close requests continue anyway - leading to
+	 * the 'Save changes?' dialog being shown again.
+	 * This flag is set to true if the quit was cancelled recently, and then reset by Platform.runLater().
+	 * The purpose is to give an indication that the user has already requested to cancel quitting, and shouldn't be
+	 * immediately prompted again.
+	 */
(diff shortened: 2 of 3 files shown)


finglis (approved):
Working well, many thanks!

Issue was raised by a user

alanocallaghan:
Pinging @petebankhead about the change in the interface as it'll possibly break stuff, if not merge away

petebankhead:
If you add a `default` implementation of the new method then nothing should break, but I don't think it really matters here since I'm not aware of anyone being affected by the change.

More importantly, the behavior seems a bit unexpected to me. I opened QuPath (no images), opened a script editor, and added text to two editors.

Then attempting to quit QuPath results in

Screenshot 2024-05-21 at 17 29 33

That's fine, but if I choose `Cancel` then the dialog disappears while QuPath remains open.

Then if I attempt to quit again, I see

Screenshot 2024-05-21 at 17 30 21

with no indication that 'Untitled 2' is a script - and no script editor visible. So I think this has the potential to be confusing.

I think that:
* Pressing cancel shouldn't result in the script editor window being closed
* The dialog message should somehow distinguish when the quit attempt is being blocked by a script rather than something else

alanocallaghan:
Hmm yes, I thought it would not close on cancel, and obviously if it has been closed we shouldn't check any more as it should be self contained. Odd

alanocallaghan:
I can't fully reproduce this - if I do the following:

- Open QuPath
- Edit two scripts
- Attempt to close QuPath
- Hit "Cancel" on the script editor save on exit prompt

Then the cancel dialog disappears with nothing changed, ie the script editor remains open.

However, I can close the script editor by using its window close button, then close QuPath, and at that point I get the same "Untitled 2" close prompt that you show in the second screenshot, which I agree is very confusing.

I think that clarifying the text on the script editor save prompt is a good improvement either way. However, should we prompt users to save scripts when closing the editor, rather than when closing QuPath? Seems more friendly in a way, unless we also add some caching/autosave to scripts to ensure work isn't lost, because eg if a users closes the script editor, they may think it's saved, then QuPath crashes, and work is lost.

finglis:
I'm also not able to reproduce the script editor closing on cancel currently (on mac). 
I do prefer the friendliness of the save prompt being when the script editor is closed since there is no visual indicator there is an unsaved script. I guess the notification badge could be used here as an alternative but think I am in favour of save on script editor close.

petebankhead:
I can still reproduce the issue on macOS following the steps Alan mentions (where 'edit 2 scripts' means new ones, not existing scripts).

> However, should we prompt users to save scripts when closing the editor, rather than when closing QuPath?

Probably. I think I'll find this annoying sometimes, because I close the script editor as a way of hiding the window - because otherwise it isn't possible to hide a (child) window. So we might need a new hide option, although then we're almost back to the original problem.

Another option would be to show the script editor when the user attempts to quit, and at that point clearly notify them that they have unsaved *scripts* (before moving in to whatever other unsaved stuff they might have).

petebankhead:
I also can't close QuPath *without* interacting with the script editor, because I get an exception
```
15:17:30.307	[JavaFX Application Thread]	ERROR	qupath.lib.gui.QuPathUncaughtExceptionHandler	Cannot invoke "javafx.stage.Stage.close()" because "this.dialog" is null	java.lang.NullPointerException: Cannot invoke "javafx.stage.Stage.close()" because "this.dialog" is null
	at qupath.lib.gui.scripting.DefaultScriptEditor.promptToClose(DefaultScriptEditor.java:518)
	at qupath.lib.gui.QuPathGUI.handleCloseMainStageRequest(QuPathGUI.java:1018)
	at javafx.base@22.0.1/com.sun.javafx.event.CompositeEventHandler.dispatchBubblingEvent(CompositeEventHandler.java:86)
	at javafx.base@22.0.1/com.sun.javafx.event.EventHandlerManager.dispatchBubblingEvent(EventHandlerManager.java:232)
	at javafx.base@22.0.1/com.sun.javafx.event.EventHandlerManager.dispatchBubblingEvent(EventHandlerManager.java:189)
	at javafx.base@22.0.1/com.sun.javafx.event.CompositeEventDispatcher.dispatchBubblingEvent(CompositeEventDispatcher.java:59)
	at javafx.base@22.0.1/com.sun.javafx.event.BasicEventDispatcher.dispatchEvent(BasicEventDispatcher.java:58)
	at javafx.base@22.0.1/com.sun.javafx.event.EventDispatchChainImpl.dispatchEvent(EventDispatchChainImpl.java:114)
	at javafx.base@22.0.1/com.sun.javafx.event.EventUtil.fireEventImpl(EventUtil.java:74)
	at javafx.base@22.0.1/com.sun.javafx.event.EventUtil.fireEvent(EventUtil.java:54)
	at javafx.base@22.0.1/javafx.event.Event.fireEvent(Event.java:198)
	at javafx.graphics@22.0.1/com.sun.javafx.stage.WindowPeerListener.closing(WindowPeerListener.java:100)
	at javafx.graphics@22.0.1/com.sun.javafx.tk.quantum.GlassStage.lambda$requestClosingAllWindows$3(GlassStage.java:204)
	at java.base/java.security.AccessController.doPrivileged(AccessController.java:400)
	at javafx.graphics@22.0.1/com.sun.javafx.tk.quantum.GlassStage.requestClosingAllWindows(GlassStage.java:203)
	at javafx.graphics@22.0.1/com.sun.javafx.tk.quantum.QuantumToolkit$2.handleQuitAction(QuantumToolkit.java:370)
	at javafx.graphics@22.0.1/com.sun.glass.ui.mac.MacApplication$4.action(MacApplication.java:226)
```

alanocallaghan:
> Probably. I think I'll find this annoying sometimes, because I close the script editor as a way of hiding the window - because otherwise it isn't possible to hide a (child) window. So we might need a new hide option, although then we're almost back to the original problem.

Yeah, I prefer the behavior when it's not a child except for the fact that it shows up as a different icon on win/Mac because it lets me put the editor behind the main qupath window. I think I looked for a middle ground but didn't find one.

alanocallaghan:
> I also can't close QuPath without interacting with the script editor, because I get an exception

This should be fixed.

> I can't fully reproduce this - if I do the following: [...]

I did manage to reproduce it just now, it should be fixed as well.

I've added an exit method to the editor's menu, with ctrl/cmd+q, and added handling to general close requests (these were previously commented out).

PR #1528, opened 2024-05-22
Decreased size of downsampled image

### LINKED ISSUE #1527: Black column on downsampled images
## Bug report

*Describe the bug*
At some downsample levels, images can display a 1 pixel-width black column on the right edge.

*To Reproduce*
1. Open the [GTEX-12584-1526 image](https://brd.nci.nih.gov/brd/imagedownload/GTEX-12584-1526
). This image has levels with downsamples 1, 4, and 16.
2. Check the right edge of the image. No black column, everything is normal.
3. Click on File, Export images...,  Original pixels, JPEG (or PNG) and downsample of 32.
4. Check the right edge of the produced image. There is a black column of width 1.

*Expected behavior*
No black column on any downsample.

*Desktop:*
 - OS: macOS and Linux, so probably not OS dependant.
 - QuPath Version: 0.6.0-SNAPSHOT

*Additional context*
The problem occurs in the [AbstractTileableImageServer.readRegion()](https://github.com/qupath/qupath/blob/76415104d0653c674716d69a1ecb51e59dae0e83/qupath-core/src/main/java/qupath/lib/images/servers/AbstractTileableImageServer.java#L266) function. When requesting an image with downsample 32:
* A black image of width 1432 (=`round(imageWidth/32)`) is created ([here](https://github.com/qupath/qupath/blob/76415104d0653c674716d69a1ecb51e59dae0e83/qupath-core/src/main/java/qupath/lib/images/servers/AbstractTileableImageServer.java#L302C4-L302C17)).
* All tiles with downsample 16 (because it's the lowest resolution image) are retrieved ([here](https://github.com/qupath/qupath/blob/76415104d0653c674716d69a1ecb51e59dae0e83/qupath-core/src/main/java/qupath/lib/images/servers/AbstractTileableImageServer.java#L275C3-L275C13)).
* Images corresponding to the above tiles are written to the above black image ([here](https://github.com/qupath/qupath/blob/76415104d0653c674716d69a1ecb51e59dae0e83/qupath-core/src/main/java/qupath/lib/images/servers/AbstractTileableImageServer.java#L328)).

The problem is that the resulting image of downsample 32 is of width 1432, which corresponds to pixels 0 to **45824** (=`1432*32`) of the full resolution image, while the width of the image of downsample 16 is 2863, which corresponds to pixels 0 to **45808** (=`2863*16`) of the full resolution image.

45808 < 45824, so the last pixel of each row of the resulting image of downsample 32 is not painted.

Rylern:
A proposal to fix #1527. It simply consists in reducing the size of the downsampled image by rounding down its width and height, instead of rounding it.

### DIFF
--- a/qupath-core/src/main/java/qupath/lib/images/servers/AbstractTileableImageServer.java
+++ b/qupath-core/src/main/java/qupath/lib/images/servers/AbstractTileableImageServer.java
@@ -296,8 +296,8 @@ public abstract class AbstractTileableImageServer extends AbstractImageServer<Bu
 		
 		long startTime = System.currentTimeMillis();
 		// Handle the general case for RGB
-		int width = (int)Math.max(1, Math.round(request.getWidth() / request.getDownsample()));
-		int height = (int)Math.max(1, Math.round(request.getHeight() / request.getDownsample()));
+		int width = Math.max(1, (int) (request.getWidth() / request.getDownsample()));
+		int height = Math.max(1, (int) (request.getHeight() / request.getDownsample()));
 		if (isRGB()) {
 			BufferedImage imgResult = createDefaultRGBImage(width, height);
 			Graphics2D g2d = imgResult.createGraphics();


petebankhead:
It's tempting, but it feels risky. The choice of `round` and `floor` when converting pixel coordinates has been problematic in the past, e.g. see https://github.com/qupath/qupath/pull/1072#issuecomment-1278540089

I'm not 100% sure, but I think switching this behavior could break or confuse other things whenever we want to try to ensure we obtain a consistent image size.

Rylern:
I will close this PR then. I don't see any other easy fix

PR #1531, opened 2024-05-24
Fix black columns with downsampled images

petebankhead:
Attempts to fix https://github.com/qupath/qupath/issues/1527

This tries to take a conservative approach that should fix the most common occurrence of this: basically, if a request is made for to get the 'last' pixels on the right or bottom of the image, then a check is made to see if this can be fulfilled using the available lower resolution tiles.

If the tiles don't quite make it (due to the lower pyramid levels being truncated), then a slightly smaller image can be returned - typically missing the last row and/or column.

The unfortunate consequence would be that we can't compute the returned image size based upon the  `RegionRequest` (width, height and downsample) - because this might have been internally adjusted. To handle that, we resize then to fit expectations.

It feels messy and suboptimal, but failing to do this can result in black rows or columns appearing - which can thwart thresholds and image statistics and certain resolutions, and this is more problematic.

### DIFF
--- a/qupath-core/src/main/java/qupath/lib/images/servers/AbstractTileableImageServer.java
+++ b/qupath-core/src/main/java/qupath/lib/images/servers/AbstractTileableImageServer.java
@@ -298,33 +298,10 @@ public abstract class AbstractTileableImageServer extends AbstractImageServer<Bu
 		int width = (int)Math.max(1, Math.round(request.getWidth() / request.getDownsample()));
 		int height = (int)Math.max(1, Math.round(request.getHeight() / request.getDownsample()));
 
-		// Fix output size to handle right/bottom edge issue caused by rounding/flooring within image pyramid
-		// See https://github.com/qupath/qupath/issues/1527
-		// The problem is that a black border can be created when there aren't quite enough pixels to fill the region,
-		// which happens if the lower-resolution levels have been truncated (even by a fraction of a pixel)
-		if (request.getDownsample() > 1 && nResolutions() > 1
-				&& request.getMaxX() == getWidth() || request.getMaxY() ==  getHeight()) {
-			int maxX = -Integer.MAX_VALUE;
-			int maxY = -Integer.MAX_VALUE;
-			for (var tile : tiles) {
-				maxX = Math.max(maxX, tile.getRegionRequest().getMaxX());
-				maxY = Math.max(maxY, tile.getRegionRequest().getMaxY());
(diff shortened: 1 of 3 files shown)


qupath-core/src/main/java/qupath/lib/images/servers/AbstractTileableImageServer.java
```diff
@@ -426,26 +407,89 @@ else if (xEnd - xStart <= 0 || yEnd - yStart <= 0)
...
+				&& (request.getMaxX() == getWidth() || request.getMaxY() == getHeight())
+				&& (request.getMinX() >= 0 && request.getMinY() >= 0)) {
+			int maxX = -Integer.MAX_VALUE;
+			int maxY = -Integer.MAX_VALUE;
+			for (var tile : tiles) {
+				maxX = Math.max(maxX, tile.getRegionRequest().getMaxX());
+				maxY = Math.max(maxY, tile.getRegionRequest().getMaxY());
+			}
+			if (maxX < request.getMaxX() || maxY < request.getMaxY()) {
+				int width2 = (int)Math.max(1, Math.round((maxX - request.getMinX()) / request.getDownsample() - 1e-9));
```
Rylern, line 479: Maybe a small comment explaining the `- 1e-9` would be useful (it's not immediately obvious)

Rylern (commented):
Running the Liverquant tissue detection works well with this PR. However, when I click on File, Export images..., Original pixels, JPEG (or PNG) and downsample of 32 with the [GTEX-12584-1526 image](https://brd.nci.nih.gov/brd/imagedownload/GTEX-12584-1526), I still get the black column, do you know why?

qupath-core/src/main/java/qupath/lib/images/servers/AbstractTileableImageServer.java
petebankhead: Good point, I'll add one. It's because it appears x.5 needs to be rounded down, not up (at least for the troublesome image).

petebankhead:
Interesting, I have no idea why... I'll investigate

petebankhead:
I've added comments. The difference with export is because of the use of `ImageServers.pyramidalize`. There is a workaround possible by creating a rectangular ROI around the entire image first (because of this PR).

It's not optimal, but I think we can leave it at the comment - we might move away from this image export before it becomes a significant enough problem to warrant more time.

PR #1532, opened 2024-05-24
Improve display of annotation names

petebankhead:
Improve the display of annotation names.

## QuPath v0.5.1
image

## With PR
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ This is a *work in progress* for the next major release.
 
 ### Enhancements
 * Read and write OME-Zarr images (https://github.com/qupath/qupath/pull/1474)
+* Improved display of annotation names
 
 ### Bugs fixed
 * Tile export to .ome.tif can convert to 8-bit unnecessarily (https://github.com/qupath/qupath/issues/1494)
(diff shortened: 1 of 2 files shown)


petebankhead:
This should always find a point on the ROI and connect the name to that. This should make it easier to see which object is named, especially if the object has disconnected parts and holes.

The last commit uses the selected object color to further reduce ambiguity.

Rylern:
Looks good to me, just one thing: is `paintOverlay` supposed to be called from multiple threads? Some fields are read/modified without synchronization

petebankhead:
@Rylern Very good question... no, but maybe?

Its primary use is for showing overlays in the viewer, where it should really be called on the application thread (or a background thread dedicated to rendering).

It *could* also be called when writing out a rendered image + overlays... and *maybe* that would involve writing tiles in parallel. If so, then I'd expect important variables never to be changed: we'd create an overlay per image we want to export.

Therefore I can't think of a time when this will be problematic in practice, but it's definitely something to be cautious about. 

If we don't find any specific case where the current behavior is problematic, then I'm not sure it's worth making it properly thread-safe - and instead devote the time that would require to an entirely new viewer with overlays that are designed much better than this from the start.

PR #1540, opened 2024-06-11
Support log histograms with measurement tables

petebankhead:
Also use log10 rather than log (thanks Alan!)


alanocallaghan:
I was just about to mention that we currently call normalised counts (ie density) "Counts", but this also solves that nitpick

PR #1544, opened 2024-06-12
Implement scatter plot in measurement table

alanocallaghan:
Implement scatter plots in measurement viewer alongside histograms.

Users can select X and Y variables, and select points in the plot to view them in the UI. Options currently are:

- Number of points shown, above this number we do subsampling
- Random seed (for subsampling)
- Show grid
- Show axis labels
- Point size
- Point opacity

Screenshot:  Screenshot from 2025-01-24 17-15-15


alanocallaghan:
Also tbc whether the scatter plots here are replaced with a more lightweight version that doesn't create a billion JavaFX nodes, or whether we just subsample the points for display

alanocallaghan:
Closing this as performance is horrible with large numbers of points, but will likely revisit after implementing a binning and/or subsampling strategy

alanocallaghan:
Reopened just so I don't delete the branch and forget about it.

I think the existing number of point limits plus a smarter subsampling strategy (density-based) should be fine... maybe?

Something like a binned 2D KDE followed by probabilistic subsampling per bin

alanocallaghan:
Suggest as a first pass for merging this to remove the 2 measurement classifier, and to add a parameterised maximum number of points.

This may also be redrawing from scratch more than is necessary.

Rylern (approved):
Looks good. Only two things:

* When I give a negative number to "Max number of elements", an uncaught exception is thrown.
* The `ScatterPlotChart` and `ScatterPlotDisplay` files could have some Javadocs on their public functions.

qupath-gui-fx/src/main/java/qupath/lib/gui/charts/ScatterPlotChart.java
```diff
@@ -0,0 +1,337 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.Random;
+import java.util.function.Function;
+
+/**
+ * A scatter plot wrapping the JavaFX chart {@link ScatterChart} to add functionality like
+ */
+public class ScatterPlotChart extends ScatterChart<Number, Number> {
+    private final IntegerProperty rngSeed = new SimpleIntegerProperty(42);
```
petebankhead, line 36: Please add a blank line here, for consistency with all our other classes & most Java code.

Most classes should also have a logger, since there is probably something, somewhere worth logging.

qupath-gui-fx/src/main/java/qupath/lib/gui/charts/ScatterPlotDisplay.java
```diff
@@ -0,0 +1,139 @@
...
+    private final PathTableData<PathObject> model;
+    private final ComboBox<String> comboNameX = new ComboBox<>();
+    private final ComboBox<String> comboNameY = new ComboBox<>();
+    private final BorderPane pane = new BorderPane();
+    private final ParameterList paramsScatter = new ParameterList()
+            .addIntParameter("nPoints", "Max number of points", 10000, null,  "Maximum number of points to be drawn (>=1)")
+            .addIntParameter("randomSeed", "Random seed", 42, null, "Random seed for point subsampling")
+            .addBooleanParameter("drawGrid", "Draw grid", true, "Whether to draw gridlines on the plot")
+            .addBooleanParameter("drawAxes", "Draw axes", true, "Whether to draw axis ticks on the plot")
+            .addDoubleParameter("pointOpacity", "Point opacity", 1, null, "The opacity of points displayed on the plot.")
```
petebankhead, line 31: These should have lower and upper bounds, which would result in them being sliders, e.g.
```java
            .addDoubleParameter("pointOpacity", "Point opacity", 1, null, 0, 1, "Set the opacity for individual points (0-1)")
            .addIntParameter("pointSize", "Point size", 4, "px", 1, 10, "Set the size for each individual point");
```
where I set `pointSize` to be an int only because the alternative would give a slider that supports lots of decimal places, which doesn't look great.

qupath-gui-fx/src/main/java/qupath/lib/gui/charts/ScatterPlotChart.java
```diff
@@ -0,0 +1,337 @@
...
+    }
+
+    private void resampleAndUpdate() {
+        Collections.shuffle(data, new Random(rngSeed.get()));
+        series.getFirst().getData().clear();
+        series.getFirst().getData().setAll(data.subList(0, Math.min(maxPoints.get(), data.size())));
+        updateChart();
+    }
+
+    void setDataFromMeasurements(Collection<PathObject> pathObjects, String xMeas, String yMeas) {
```
petebankhead, line 163: Use `<? extends PathObject>` whenever possible, to avoid unnecessary restriction.

petebankhead: Would also prefer full words instead of `xMeas` and `yMeas` - not sure the extra cognitive autocompletion it requires is worth it.

qupath-gui-fx/src/main/java/qupath/lib/gui/charts/ScatterPlotChart.java
```diff
@@ -0,0 +1,337 @@
...
+    }
+
+    /**
+     *
+     * Try to select an object if possible (e.g. because a user clicked on it).
+     * @param pathObject the object to select
+     * @param addToSelection if true, add to an existing selection; if false, reset any current selection
+     * @param centerObject if true, try to center it in a viewer (if possible)
+     */
+    private static void tryToSelect(PathObject pathObject, ImageData<?> imageData, QuPathViewer viewer, boolean addToSelection, boolean centerObject) {
```
petebankhead, line 284: Isn't this code duplication from `Charts`? Is it necessary?

qupath-gui-fx/src/main/java/qupath/lib/gui/charts/ScatterPlotChart.java
```diff
@@ -0,0 +1,337 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.Random;
+import java.util.function.Function;
+
+/**
+ * A scatter plot wrapping the JavaFX chart {@link ScatterChart} to add functionality like
```
petebankhead, line 33: Either mention explicitly what it is like, or add a comma first if intending a regional, conversational style

petebankhead (commented):
Added some comments - the main one is there there seems to be a lot of code duplication between this and `Charts`, is it needed?

I'm also not entirely sure that `ParameterList` is the way to go; it was used a lot in early QuPath, but we now tend to code the JavaFX explicitly to have more control over how it looks. This could be changed later, but for now at least I don't think opacity and point size should be text boxes.

I've reservations about the use of two split panes for the histogram and scatterchart (rather than one, and then tabs to move between charts) since it uses a lot of horizontal space. Although I think `SummaryMeasurementTableCommand` will need a bigger revision - which I'd prefer to leave until https://github.com/qupath/qupath/pull/1747 is merged.

qupath-gui-fx/src/main/java/qupath/lib/gui/charts/ScatterPlotDisplay.java
alanocallaghan: I didn't particularly like the behaviour of the plot using sliders, but fair enough

qupath-gui-fx/src/main/java/qupath/lib/gui/charts/ScatterPlotChart.java
alanocallaghan: Short answer: no
Long answer: I found it difficult to find a good point to refactor the existing code in Charts, as it performs subsampling in a spot that I didn't expect.* But that's not a good reason to not refactor, as the charts class would also benefit.

*possibly resulting in confusing code and/or unexpected outputs

qupath-gui-fx/src/main/java/qupath/lib/gui/charts/ScatterPlotDisplay.java
petebankhead: Sliders, spinners... something bounded. If using `ParameterList` there aren't many options.

qupath-gui-fx/src/main/java/qupath/lib/gui/charts/ScatterPlotChart.java
alanocallaghan: I have made a first pass at a refactor, although I don't think I did a good job.

This means that now the following two calls will produce plots with the same number of points, where they didn't before:

```groovy
import qupath.lib.gui.charts.Charts.ScatterChartBuilder

new ScatterChartBuilder()
    .measurements(getDetectionObjects(), "Nucleus: Area", "Nucleus: Perimeter")
    .limitDatapoints(10)
    .show()
    
    
new ScatterChartBuilder()
    .limitDatapoints(10)
    .measurements(getDetectionObjects(), "Nucleus: Area", "Nucleus: Perimeter")
    .show()
```

qupath-gui-fx/src/main/java/qupath/lib/gui/charts/ScatterPlotChart.java
alanocallaghan: Apologies, evidently I tabbed into the other class to see which options I added and then forgot to complete this

alanocallaghan:
Using tabs rather than two panels isn't much work so I implemented it also

qupath-gui-fx/src/main/java/qupath/lib/gui/charts/ScatterPlotChart.java
petebankhead: Don't think changing behavior is a big problem, `Charts` was just there for scripting and not very well known. Should mention in the changelog though.

I'm not entirely sure if subsampling should be at the series or chart level. Subsampling all datapoints makes sense, although for rare classes I imagine we could often lose all the instances so they aren't represented at all.

alanocallaghan: I think subsampling the series is the easiest option that is unlikely to have consequences; I'll update the docs to make that clear

petebankhead: Yeah, although if there are 10x of one class than another, they might end up having the same number. Which could be ok, I guess... or if we have a limit of 10,000 points but 100 classes then we still end up with a whole lot of points. No option seems without problems somewhere.

alanocallaghan: I guess the nicest way to do it would be to proportionally subsample from all of the series, but I doubt many people will notice which way we do it

alanocallaghan:
I think I addressed your comments. I would like to further clean the code in Charts at some point, but I suspect this can wait for a broader overhaul in a future release

alanocallaghan:
Assuming folks are happy with this as-is; happy to address further issues in another PR if not

PR #1554, opened 2024-07-02
Core code for stain normalization

petebankhead:
## Purpose

Introduce stain normalization as a *possibility* via core support (i.e. there aren't any commands to apply this easily or interactively yet).

This focusses firstly on color deconvolution and reconvolution as a method of stain normalization.

> **This is not just for visualization!**
> This creates a JSON-friendly `ImageServer`, so can be used to add normalized images to projects and thereby impact analysis. It may be an important step towards developing scripts and classifiers that work across larger datasets exhibiting more variation.

## Script
The following Groovy script shows it in action:

```groovy
/**
 * Script to demonstrate color standardization using color deconvolution.
 * To use it, you should first open an RGB image in QuPath and set the stain vectors (including background),
 * e.g. using 'Estimate stain vectors'.
 */

import javafx.application.Platform
import qupath.lib.color.ColorDeconvolutionStains
import qupath.lib.gui.viewer.overlays.PixelClassificationOverlay
import qupath.lib.images.ImageData
import qupath.lib.images.servers.ImageServer
import qupath.lib.images.servers.TransformedServerBuilder

import java.awt.image.BufferedImage

import static qupath.lib.gui.scripting.QPEx.*

def viewer = getCurrentViewer()
def imageData = getCurrentImageData()

// Show as an overlay, instead of adding a new image to the project
boolean showAsOverlay = true

// Check we have a valid image
if (imageData == null || imageData.getColorDeconvolutionStains() == null) {
    throw new IllegalArgumentException("This script requires an image with color deconvolution stains")
}

// Define target stains, or use defaults for H&E or H-DAB
ColorDeconvolutionStains targetStains = getDefaultTargetStains(imageData, true)
if (targetStains == null) {
    println "No color deconvolution stains defined for this image"
    return
}

// We would optionally adjust this to rescale the stains
def scales = [1, 1, 1] as double[]

Platform.runLater {
    if (showAsOverlay) {
        // Show normalized image as an overlay
        def overlay = PixelClassificationOverlay.create(
                viewer.getOverlayOptions(),
                imgData -> createServer(imgData, targetStains, scales),
                viewer.getImageDisplay()
        )
        overlay.setLivePrediction(true)
        viewer.setCustomPixelLayerOverlay(overlay)
    } else {
        // Open normalized image in the viewer (this should also add it to the project)
        def newData = new ImageData<BufferedImage>(
                createServer(imageData, targetStains, scales),
                imageData.getImageType())
        newData.setColorDeconvolutionStains(targetStains)
        // Uncomment these lines to replace the current image, instead of opening a new one
//        getProjectEntry().saveImageData(newData)
//        newData = getProjectEntry().readImageData()
        viewer.setImageData(newData)
        viewer.resetCustomPixelLayerOverlay()
    }
}

/**
 * Create an image server that normalizes the colors using color deconvolution.
 * @param imageData the original image
 * @param targetStains the target stain vectors for the output image
 * @param scales optionally rescale each stain by the specified factor
 * @return
 */
ImageServer<BufferedImage> createServer(ImageData<BufferedImage> imageData, ColorDeconvolutionStains targetStains, double[] scales = null) {
    return new TransformedServerBuilder(imageData.getServer())
        .stainNormalize(imageData.getColorDeconvolutionStains(), targetStains, scales)
        .build()
}

/**
 * Get the default stains for an image type, or null if no default stains are defined
 * @param imageType
 * @return
 */
ColorDeconvolutionStains getDefaultTargetStains(ImageData<?> imageData, boolean preferDefaultStains) {
    if (preferDefaultStains) {
        switch (imageData.getImageType()) {
            case ImageData.ImageType.BRIGHTFIELD_H_E:
                return ColorDeconvolutionStains.makeDefaultColorDeconvolutionStains(
                        ColorDeconvolutionStains.DefaultColorDeconvolutionStains.H_E)
            case ImageData.ImageType.BRIGHTFIELD_H_DAB:
                return ColorDeconvolutionStains.makeDefaultColorDeconvolutionStains(
                        ColorDeconvolutionStains.DefaultColorDeconvolutionStains.H_DAB)
            default:
                println "WARN: No default stains defined for image type ${imageData.getImageType()}"
        }
    }
    // The most we can do is 'whiten' the background
    return imageData.getColorDeconvolutionStains().changeMaxValues(255, 255, 255)
}
```

## Example
Here, I've applied the script to [CMU-1.ndpi (CC0, from OpenSlide test data)](https://openslide.cs.cmu.edu/download/openslide-testdata/Hamamatsu/CMU-1.ndpi).

### Original image
cmu-1-ndpi-orig

### After running the script
cmu-1-ndpi-normalized-full

### With stain intensity adjustment

The script above effectively shifts the stain colors and sets the background to white.

Updating the `scales` variable to
```groovy
def scales = [1, 0.5, 0] as double[]
```
effectively suppresses the residual value and weakens the eosin - demonstrating how intensity variations can also be compensated for.

cmu-1-ndpi-normalized-less-eosin


Rylern:
It looks good to me (however I'm not very familiar with the theory behind it).

Just 2 things:
* Is there a reason for defining the `BufferedImageNormalizer` interface, instead of just using [`Function<BufferedImage, BufferedImage>`](https://docs.oracle.com/javase/8/docs/api/java/util/function/Function.html) or [`UnaryOperator<BufferedImage>`](https://docs.oracle.com/javase/8/docs/api/java/util/function/UnaryOperator.html)?
* Shouldn't there be unit tests for at least for the `ColorTransformer.colorDeconvolveReconvolveRGBArray()` function and the `SubtractOffsetAndScaleNormalizer` class?

alanocallaghan:
+1 for both of Leo's comments, otherwise looks good to me

petebankhead:
Thanks @Rylern and @alanocallaghan - good call making me right the tests, that helped identify and fix some bugs, and address some suboptimal behavior concerning rounding and out-of-range values.

> Is there a reason for defining the BufferedImageNormalizer interface, instead of just using [Function<BufferedImage, BufferedImage>](https://docs.oracle.com/javase/8/docs/api/java/util/function/Function.html) or [UnaryOperator<BufferedImage>](https://docs.oracle.com/javase/8/docs/api/java/util/function/UnaryOperator.html)?

I've updated this to use [`BufferedImageOp`](https://docs.oracle.com/en/java/javase/21/docs/api/java.desktop/java/awt/image/BufferedImageOp.html), which then
1. allows the classes to be used directly with [`Graphics2D.drawImage`](https://docs.oracle.com/en/java/javase/21/docs/api/java.desktop/java/awt/Graphics2D.html#drawImage(java.awt.image.BufferedImage,java.awt.image.BufferedImageOp,int,int)).
2. makes it possible to control whether in-place image conversion is used

Potentially one of the other standard Java interfaces could be used, but I want to avoid giving the impression that it's a good idea to use arbitrary methods or lambda expressions. These can be convenient for testing, but will fail when images are added to a project - because we need the methods to be JSON-serializable.

The JSON serialization is taken care of through the `ImageServers` class, and we have to explicitly register every normalization class so that it can be used within a project.

PR #1556, opened 2024-07-03
Project browser improvements

petebankhead:
* Support hiding project thumbnails
* Supporting hiding / resizing thumbnails from a context menu (not just the preferences)
  * Hiding thumbnails may improve performance for large projects if thumbnails haven't been generated yet
  * It also addresses a request for hide thumbnails when masking image names
* Improve code by separating out some methods
* Fix a bug that meant metadata couldn't be edited when using `gradlew run`
* Retain sorting after adding metadata (stolen from @alanocallaghan's code at #1538)
* Move styles to `main.css` (rather than calling `setStyle(String)`)


petebankhead:
The latest commit adds a warning indicator if image files are missing:

Screenshot 2024-07-03 at 12 59 41

This can be helpful if you want to remove images from a project, and need to know which have missing files.

In case the check is a performance issue, the results are cached until a new project is opened.
It's also possible to turn off the checks with the *'Skip file checks in projects'* preference.

Rylern:
LGTM.

I think it would be beneficial to the `ProjectBrowser` class to put the UI in a separate FXML file.

alanocallaghan (approved):
Looks good, although difficult to review the diff properly. Soft agree on FXML, but there's a lot of code that needs to be FXMLed and I think we should have more clearly defined standards before refactoring old code

PR #1558, opened 2024-07-12
Add Collection<ShapeFeatures> method for adding measurements

alanocallaghan:
See: https://github.com/qupath/qupath-extension-instanseg/pull/38

### DIFF
--- a/qupath-core-processing/src/main/java/qupath/lib/analysis/features/ObjectMeasurements.java
+++ b/qupath-core-processing/src/main/java/qupath/lib/analysis/features/ObjectMeasurements.java
@@ -21,17 +21,10 @@
 
 package qupath.lib.analysis.features;
 
-import java.awt.image.BufferedImage;
-import java.io.IOException;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.LinkedHashMap;
-import java.util.LinkedHashSet;
-import java.util.Map;
-import java.util.Set;
-
+import ij.process.ByteProcessor;
+import ij.process.ColorProcessor;
+import ij.process.FloatProcessor;
+import ij.process.ImageProcessor;
 import org.apache.commons.math3.stat.descriptive.DescriptiveStatistics;
 import org.apache.commons.math3.stat.descriptive.StatisticalSummary;
 import org.locationtech.jts.algorithm.Area;
@@ -45,11 +38,6 @@ import org.locationtech.jts.geom.Polygonal;
 import org.locationtech.jts.geom.util.AffineTransformation;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
-import ij.process.ByteProcessor;
-import ij.process.ColorProcessor;
-import ij.process.FloatProcessor;
-import ij.process.ImageProcessor;
 import qupath.imagej.tools.IJTools;
 import qupath.imagej.tools.PixelImageIJ;
 import qupath.lib.analysis.images.SimpleImage;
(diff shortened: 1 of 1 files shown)


petebankhead:
I feel this draws more attention to my dubious design decision: specify nothing and get everything.

I suggest deprecating `addShapeMeasurements` and adding
```java
public static void addAllShapeMeasurements(Collection<? extends PathObject> pathObjects, PixelCalibration cal) {
   return addShapeMeasurements(pathObjects, cal, Arrays.asList(ShapeFeatures.values());
}

// Mark as deprecated from v0.6.0
public static void addShapeMeasurements(Collection<? extends PathObject> pathObjects, PixelCalibration cal, ShapeFeatures... features) {
   if (features.length == 0)
      return addAllShapeMeasurements(pathObjects, cal);
   else
      return addShapeMeasurements(pathObjects, cal, Arrays.asList(features);
}

public static void addShapeMeasurements(Collection<? extends PathObject> pathObjects, PixelCalibration cal, Collection<ShapeFeature> shapeFeatures) {
   // As you have it, except don't add anything if the collection is empty
}
```
What do you think?

petebankhead:
This might cause some trouble elsewhere when `addShapeMeasurements` is used with the assumption nothing-means-everything (e.g. in `QP`), so would complicate things a bit. But less now than in the future, if we agree that an empty array or collection should add no measurements instead of all measurements.

petebankhead:
I've finally remembered my (kind of) justification for defaulting to adding all shape measurements if none are specified.

[`Files.newOutputStream(Path path, OpenOptions... options)`](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/nio/file/Files.html#newOutputStream(java.nio.file.Path,java.nio.file.OpenOption...)) does something slightly similar:

> If no options are present then this method works as if the [CREATE](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/nio/file/StandardOpenOption.html#CREATE), [TRUNCATE_EXISTING](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/nio/file/StandardOpenOption.html#TRUNCATE_EXISTING), and [WRITE](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/nio/file/StandardOpenOption.html#WRITE) options are present.

So the existing behavior is:
```java
// Adds specified shape measurements
addShapeMeasurements(pathObjects, call, ShapeFeatures.AREA, ShapeFeatures.CIRCULARITY)

// Adds all shape measurements
addShapeMeasurements(pathObjects, call)
```

With this PR, we have
```java
// Adds specified shape measurements
addShapeMeasurements(pathObjects, call, List.of(ShapeFeatures.AREA, ShapeFeatures.CIRCULARITY))

// Adds all shape measurements?
addShapeMeasurements(pathObjects, call, Collections.emptyList())
```
This last one feels wrong to me - it feels that no measurements should be added, because we have had to explicitly pass something - and that was an empty list. It does match the behavior using varargs, so making them inconsistent feels strange too, but with varargs the empty array is created implicitly.

So I think our options are:
1. Merge this PR to support passing args as a collection; an empty collection adds all shape measurements
2. Don't merge this PR; just keep the previous behavior and the calling code needs to figure out what to do
3. Add support for specifying features in a collection, but in a different way (possibly changing the old behavior or deprecating the old method, as described above)

@alanocallaghan @Rylern @finglis what do you think?

alanocallaghan:
4. Keep the current varargs behaviour and add a collection method that behaves (more or less) as expected when given an empty collection?

petebankhead:
Should it have a different name?
I think it sounds ok, but if the names are unchanged then these would give different results:
```java
addShapeMeasurements(pathObjects, call, new ShapeFeatures[0])
addShapeMeasurements(pathObjects, call, Collections.emptyList())
```

Backing up a bit, why is it necessary to have a method that takes a collection rather than using varargs all the time anyway?

I saw the linked code, but I don't see why it can't just generate an array there and pass that as the varargs (since it can accept an array as a parameter without trouble)

alanocallaghan:
Oh fair enough, didn't realise you can pass an array to varargs. Then sure, will leave as is for now

PR #1560, opened 2024-07-17
Make plugin panes only appear once

alanocallaghan:
This means that any interactive plugin windows are effectively singletons

See https://github.com/qupath/planner/issues/2

### DIFF
--- a/qupath-gui-fx/src/main/java/qupath/lib/gui/QuPathGUI.java
+++ b/qupath-gui-fx/src/main/java/qupath/lib/gui/QuPathGUI.java
@@ -246,8 +246,9 @@ public class QuPathGUI {
 	 * can result in exceptions (or even segfaults...).
 	 */
 	private BooleanProperty menusInitializing = new SimpleBooleanProperty(false);
-	
-	
+	private final Map<Class<?>, ParameterDialogWrapper<BufferedImage>> pluginDialogs = new HashMap<>();
+
+
 	/**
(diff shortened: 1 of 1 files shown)


petebankhead:
Hmmm, I'm not sure what we gain from this?

We lose the ability to have multiple dialogs open to test the use of different parameters, e.g. for cell detection.

And also, we lose the ability to double-click on a past entry under the 'Workflow' tab to open cell detection with the correct parameters. I think this is a really handy and important feature.

So I'd tend towards rejecting and closing this, or have I missed a benefit?

alanocallaghan:
These are fair points; for me at least within a project being able to open a plugin with the same settings as I used previously seem really handy, though unsure if it outweighs the above

petebankhead:
Currently, you can do that via the workflow tab - assuming you've run it once before on that image.

Or you can leave the window open when switching images - assuming they are the same type... bad things may happen if not, e.g. fluorescence to brightfield.

Although that led me to see that bad things happen with this PR when the image type changes: QuPath isn't able to adjust and update the available options. So if I run cell detection for brightfield, then I can't use it for fluorescence within that QuPath session - because it will be stuck with the same available channels the first time it was run.

alanocallaghan:
Ah, fair enough. To me, it makes sense for there to be a single cell detection window, and I'd put sub-panes in that window if needed, but that sounds like a total pain and pretty useless in terms of overall UX gain

PR #1563, opened 2024-07-29
Improved methods to access objects from PathObjectHierarchy

petebankhead:
## Background
A goal of this PR is to further reduce the requirement to refer to specific Java subclasses of `PathObject`, e.g. `PathDetectionObject`, `PathAnnotationObject`.

This process started several releases back, e.g.
* `pathObject.isAnnotation()` should be used in preference of `pathObject instanceof PathAnnotationObject`
* `PathObjects.createAnnotationObject(ROI)` should be used instead of `new PathAnnotationObject(ROI)`.

Eventually the specific subclasses may be removed entirely, in favor of using a single `PathObject` class.

Along the way, the API should become easier to understand and to use.

## What has changed

One of the main times when the specific Java class remains relevant is when querying objects from `PathObjectHierarchy` using one of the following methods:
```java
public Collection<PathObject> getObjectsForROI(Class<? extends PathObject> cls, ROI roi)

public Collection<PathObject> getObjectsForRegion(Class<? extends PathObject> cls, ImageRegion region, Collection<PathObject> pathObjects)

public boolean hasObjectsForRegion(Class<? extends PathObject> cls, ImageRegion region)

public synchronized Collection<PathObject> getPointObjects(Class<? extends PathObject> cls)
```

This PR deprecates these methods, and provides alternatives that do not require Java classes to be specified.

### Accessing objects for a `ROI`
This provides a way to access objects within (in some sense!) a specified ROI:
```java
// Old method, discouraged (but not yet deprecated)
public Collection<PathObject> getObjectsForROI(Class<? extends PathObject> cls, ROI roi)

// New methods
public Collection<PathObject> getAllObjectsForROI(ROI roi)
public Collection<PathObject> getAnnotationsForROI(ROI roi)
public Collection<PathObject> getTilesForROI(ROI roi)
public Collection<PathObject> getCellsForROI(ROI roi)
public Collection<PathObject> getAllDetectionsForROI(ROI roi)
```

**There is an important (and possibly-unexpected) subtlety here!**

The new methods follow the original in *returning objects using QuPath's 'hierarchy' rules*:
* An annotation is 'within' a ROI if it is *completely contained* by the ROI
* A detection is 'within' a ROI if *its centroid* is within the ROI

Note that these methods take a `ROI` as input and not a `PathObject`: they aren't using the parent/child relationships between objects when deciding what to return.

However they do correspond with how QuPath determines dynamic measurements (e.g. Positive % of cells), or what happens when `resolveHierarchy()` is called.

If all this seems too confusing and you just want to get objects for a defined area, you may instead want to try...

### Accessing objects for an `ImageRegion`
This provides a way to get all objects that intersect (or even just *may* intersect) with a specified 2D image region.

```java
// Old method, deprecated
public Collection<PathObject> getObjectsForRegion(Class<? extends PathObject> cls, ImageRegion region, Collection<PathObject> pathObjects)

// New methods (the collection parameter is optional & can be null)
public Collection<PathObject> getAllObjectsForRegion(ImageRegion region, Collection<PathObject> pathObjects)
public Collection<PathObject> getAnnotationsForRegion(ImageRegion region, Collection<PathObject> pathObjects)
public Collection<PathObject> getAllDetectionsForRegion(ImageRegion region, Collection<PathObject> pathObjects)

public Collection<PathObject> getAllObjectsForRegion(ImageRegion region)
public Collection<PathObject> getAnnotationsForRegion(ImageRegion region)
public Collection<PathObject> getAllDetectionsForRegion(ImageRegion region)
```

As the javadocs mention, this is permitted to return *too many objects*.
Its purpose is to act as a fast spatial cache, where returned objects are likely to be filtered further.

To help with this, there are new methods in `PathObjectTools`:
```java
public static <T extends PathObject> Collection<T> filterByROICovers(ROI roi, Collection<T> pathObjects)
public static <T extends PathObject> Collection<T> filterByROIIntersects(ROI roi, Collection<T> pathObjects)
public static <T extends PathObject> Collection<T> filterByROIContainsCentroid(ROI roi, Collection<T> pathObjects)
```
These can apply filters that normally require some `ROI.getGeometry()` gymnastics, and may be faster through the use of prepared geometries and parallel streams.

When cells are involved, one might want to perform the tests using the *nucleus* ROI rather than the main (outer) ROI. This is also what happens internally with the hierarchy checks (i.e. the nucleus centroid is used where possible to check if a cell is inside a ROI). The following methods support that:
```java
public static <T extends PathObject> Collection<T> filterByROICoversNucleus(ROI roi, Collection<T> pathObjects)
public static <T extends PathObject> Collection<T> filterByROIIntersectsNucleus(ROI roi, Collection<T> pathObjects)
public static <T extends PathObject> Collection<T> filterByROIContainsNucleusCentroid(ROI roi, Collection<T> pathObjects)
```

In case you just want to check if objects are present quickly - but don't necessarily need the objects themselves - you can use:

```java
// Old method, deprecated
public boolean hasObjectsForRegion(Class<? extends PathObject> cls, ImageRegion region)

// New methods
public boolean hasObjectsForRegion(ImageRegion region)
public boolean hasAnnotationsForRegion(ImageRegion region)
public boolean hasDetectionsForRegion(ImageRegion region)
```

These should effectively report whether `getXXXForRegion` would return an empty collection or not, without needing to generate that collection.

### Accessing objects with point ROIs
This simply accesses objects then filters by ROI type.

```java
// Old method, deprecated
public synchronized Collection<PathObject> getPointObjects(Class<? extends PathObject> cls)

// New methods
public Collection<PathObject> getAllPointObjects()
public Collection<PathObject> getAllPointAnnotations() 
```
This should be sufficiently obscure that there is no need to have separate methods to request point objects of more classes. If you *really* need point detections, for example. filtering the resulting collection should be straightforward, e.g.
```java
var pointAnnotations = hierarchy.getAllPointObjects().stream().filter(PathObject::isDetection).toList();
```

---

To support these changes, `GeometryROI.contains(x, y)` was also updated to make use of an [`IndexedPointInAreaLocator`](https://locationtech.github.io/jts/javadoc/org/locationtech/jts/algorithm/locate/IndexedPointInAreaLocator.html) for complex geometries - enabling centroid tests to benefit from performance improvements with no special logic required outside the `ROI` class itself.

Also, some deprecated methods - and deprecated classes that used them - were removed.


petebankhead:
The following image and script show the differences in how objects might be requested from the object hierarchy.

The tests that each cell object passes contribute to the classification assigned, so that
* if two cells have the same color you can assume they were returned by the same methods
* if two cells have different colors, then at least one cell was returned by at least one method that didn't return the other cell (e.g. it intersects the selected ROI, but isn't completely covered by it)

image


```groovy
// Get hierarchy & selected object/ROI
def hierarchy = getCurrentHierarchy()
def selectedObject = getSelectedObject()
def roi = selectedObject.getROI()

// Total number of detections everywhere
def allDetections = getDetectionObjects()
println "Num detections (all): \t${allDetections.size()}"

// Direct children of the selected object
def childObjects = selectedObject.getChildObjects()
println "Num child objects: \t${childObjects.size()}"

// Get all detections for the region (rectangular bounding box, quick test)
def region = ImageRegion.createInstance(roi)
def regionObjects = hierarchy.getAllDetectionsForRegion(region, null)
println "Num in region bounds: \t${regionObjects.size()}"

// Detections within selected object, using hierarchy rules
def hierarchyWithin = hierarchy.getAllDetectionsForROI(roi)
println "Num 'within' ROI: \t${hierarchyWithin.size()}"

// Detections with nucleus (or main ROI) centroids within the selected object
def nucleusCentroidWithin = PathObjectTools.filterByRoiContainsNucleusCentroid(roi, allDetections)
println "Num nucleus centroid in ROI: \t${nucleusCentroidWithin.size()}"

// Detections with centroids within the selected object
def centroidWithin = PathObjectTools.filterByRoiContainsCentroid(roi, allDetections)
println "Num centroid in ROI: \t${centroidWithin.size()}"

// Detections with ROIs intersecting the selected object
def intersecting = PathObjectTools.filterByRoiIntersects(roi, allDetections)
println "Num intersecting ROI: \t${intersecting.size()}"

// Detections with ROIs intersecting the selected object - accessed from region
// This should contain the same elements as intersecting (possibly in a different order)
def intersecting2 = PathObjectTools.filterByRoiIntersects(roi, regionObjects)
assert intersecting.size() == intersecting2.size()
assert (intersecting as Set) == (intersecting2 as Set)

// Detections with ROIs completely within the selected object
def completelyCovered = PathObjectTools.filterByRoiCovers(roi, allDetections)
println "Num completely covered: \t${completelyCovered.size()}"

// Set classifications for visualization
allDetections.each {it.classifications = []}
childObjects.each{it.classifications += ['child']}
regionObjects.each{it.classifications += ['region']}
hierarchyWithin.each{it.classifications += ['within']}
nucleusCentroidWithin.each{it.classifications += ['nucleus-centroid']}
centroidWithin.each{it.classifications += ['roi-centroid']}
intersecting.each{it.classifications += ['intersecting']}
completelyCovered.each{it.classifications += ['covered']}
```

Rylern:
It looks good. I just have a few comments regarding tests.

Currently, a lot of tests are made within just one function. If we want to follow unit tests guidelines, each test function must only tests one thing. This means that only one function should be tested at the same time, and this function should only be called once. Also, to write the test function, the Arrange / Act / Assert / (cleanup) pattern should be used. All of this is detailed [here](https://github.com/qupath/Java-style-guide/tree/main?tab=readme-ov-file#writing-a-test-function).

Also, I think there are a few untested functions, like `PathObjectHierarchy.getAllObjectsForRegion()` or `PathObjectHierarchy.hasObjectsForRegion()`.

petebankhead:
Thanks, would you be able to provide any example tests for these functions?

I understand it's not great to test a lot in one function, but I wasn't sure how to do it without making the testing code enormous and much more complex - since an appropriate hierarchy needs to be built with objects + ROIs etc.

alanocallaghan:
One option would be to have multiple test classes with different `@BeforeEach` sections?

Related question: is the style guide for one assert, or just to test one function?

Rylern:
> since an appropriate hierarchy needs to be built with objects + ROIs etc

If you have objects that are used in several tests, you can define them as private fields of the test class, or (if their initialization is more complex) you can use `@BeforeAll` as Alan mentionned. If these objects are not immutable, make sure not to modify them inside the test functions.

Rylern:
> Related question: is the style guide for one assert, or just to test one function?

The important thing is to test one function with one set of parameters. To check the result, there is usually one `assert`, but in some situations (for example if the resulting object has no custom `equals` implementation) you can define multiple `assert`.

Rylern:
> Thanks, would you be able to provide any example tests for these functions?

Here is for `PathObjectsTools.filterByRoiCovers()`:

```java
public class TestPathObjectTools {
        private ROI roiRect = ROIs.createRectangleROI(1, 2, 100, 200, ImagePlane.getDefaultPlane());
	private ROI roiEllipse = ROIs.createEllipseROI(1, 2, 100, 200, ImagePlane.getDefaultPlane());
	private ROI roiDiamond = ROIs.createPolygonROI(List.of(
			new Point2(roiRect.getBoundsX(), roiRect.getCentroidY()),
			new Point2(roiRect.getCentroidX(), roiRect.getBoundsY()),
			new Point2(roiRect.getBoundsX()+roiRect.getBoundsWidth(), roiRect.getCentroidY()),
			new Point2(roiRect.getBoundsX(), roiRect.getBoundsY()+roiRect.getBoundsHeight())
	), ImagePlane.getDefaultPlane());
	private ROI roiOverlaps = roiRect.translate(10, 0);
	private ROI roiOverlaps2 = roiRect.translate(0, 10);
	private ROI roiSeparate = roiRect.translate(1000, 1000);
	private List<ROI> allRois = List.of(roiRect, roiEllipse, roiDiamond, roiOverlaps, roiOverlaps2, roiSeparate);
	private List<PathObject> pathObjects = allRois
			.stream()
			.map(PathObjects::createDetectionObject)
			.toList();

	@Test
	void Check_ROIs_Covered_By_Rectangle() {
		List<ROI> expectedRois = List.of(roiRect, roiEllipse, roiDiamond);

		List<ROI> coveredROIs = PathObjectTools.filterByRoiCovers(roiRect, pathObjects).stream().map(PathObject::getROI).toList();

		assertEquals(expectedRois, coveredROIs);
	}

	@Test
	void Check_ROIs_Covered_By_Ellipse() {
		List<ROI> expectedRois = List.of(roiEllipse);

		List<ROI> coveredROIs = PathObjectTools.filterByRoiCovers(roiEllipse, pathObjects).stream().map(PathObject::getROI).toList();

		assertEquals(expectedRois, coveredROIs);
	}

        // other ROIs...
}
```

PR #1575, opened 2024-08-06
Use light scrollbars with javadocs

petebankhead:
Alternative approach to https://github.com/qupath/qupath/pull/1574

#1575 creates a better separation of the stylesheet, but the advantages of the approach proposed here are:
* It requires very little code (and, in particular, doesn't involve replicating anything from modena)
* It's possible for other QuPath-wide themes to override the styles if they need to

### DIFF
--- a/qupath-gui-fx/src/main/java/qupath/lib/gui/JavadocViewerRunner.java
+++ b/qupath-gui-fx/src/main/java/qupath/lib/gui/JavadocViewerRunner.java
@@ -62,6 +62,8 @@ public class JavadocViewerRunner implements Runnable {
     private static final StringProperty javadocPath = PathPrefs.createPersistentPreference(JAVADOC_PATH_PREFERENCE, null);
     private final JavadocViewerCommand command;
 
+    private static final String JAVADOCS_STYLE_CLASS = "javadoc-viewer";
+
     /**
      * Create the command. This will not create the viewer yet.
      *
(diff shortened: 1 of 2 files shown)


qupath-gui-fx/src/main/java/qupath/lib/gui/JavadocViewerRunner.java
```diff
@@ -62,6 +62,8 @@ public class JavadocViewerRunner implements Runnable {
     private static final StringProperty javadocPath = PathPrefs.createPersistentPreference(JAVADOC_PATH_PREFERENCE, null);
     private final JavadocViewerCommand command;
 
+    private static final String JAVADOCS_STYLE_CLASS = "javadoc-viewer";
```
Rylern, line 65: Could you move that line above `private final JavadocViewerCommand command;`? I prefer defining constants above everything else

qupath-gui-fx/src/main/java/qupath/lib/gui/JavadocViewerRunner.java
```diff
@@ -89,6 +91,8 @@ public JavadocViewerRunner(Stage owner) {
                         .filter(Objects::nonNull)
                         .toArray(URI[]::new)
         );
+        if (!command.getJavadocViewer().getStyleClass().contains(JAVADOCS_STYLE_CLASS))
+            command.getJavadocViewer().getStyleClass().add("javadoc-viewer");
```
Rylern, line 95: I think it would be better to put that part in the `run()` function.

When you instantiate a `JavadocViewerCommand`, the underlying `JavadocViewer` is not created yet. However, calling `command.getJavadocViewer()` will create the `JavadocViewer` (and proceed to search for Javadocs).

As the `JavadocViewerRunner` is instantiated during QuPath startup, I think it's better if we delay the `JavadocViewer` creation to the moment when we actually open the `JavadocViewer` window for the first time. This may increase QuPath startup time.

petebankhead:
Closed in favor aof addressing this in https://github.com/qupath/javadoc-viewer

PR #1578, opened 2024-08-08
Changed fireHierarchyChangedEvent to fireObjectAddedEvent

### LINKED ISSUE #1487: Undo/Redo unexpected behavior after importing from file
Before submitting your bug report, please check the following:

* [x] I've definitely found a bug (it you're not sure, please use [image.sc](https://forum.image.sc/tags/qupath) instead)
* [x] I've checked https://qupath.github.io for a new release that might already have fixed the issue
* [x] I've checked the [Changelog](https://github.com/qupath/qupath/blob/master/CHANGELOG.md) to see if the bug has already been fixed in the next release
* [x] I've checked for existing GitHub issues describing the same problem

## Bug report

**Describe the bug**
If you import objects from a file, then select an object like a rectangle and change/move it and then finally using
undo it weirdly going to delete all added objects instead of undoing the change/move.

**To Reproduce**
1. Loading image using `Open...`
2. Loading objects using `Import objects from file`
3. Double click (select) any object
4. Move/Change the selected object, like transforming a rectangle or whatever
5. Undoing the action

**Expected behavior**
Undoing the move/change of the object.
Instead it just deletes all objects.

 - OS: Windows
 - QuPath Version: 0.5.1

Rylern:
A proposal to fix #1487.

The [`UndoRedoManager.hierarchyChanged()`](https://github.com/qupath/qupath/blob/3544e613b40fd123236936d76e2cb5ee08d855f7/qupath-gui-fx/src/main/java/qupath/lib/gui/UndoRedoManager.java#L472) function is called each time the hierarchy changes and is supposed to update the `UndoRedoManager` state. However, when loading objects using `Import objects from file`, the `PathObjectHierarchyEvent` parameter of this function contains no changed objects (`PathObjectHierarchyEvent.getChangedObjects()` function). So, [this condition](https://github.com/qupath/qupath/blob/3544e613b40fd123236936d76e2cb5ee08d855f7/qupath-gui-fx/src/main/java/qupath/lib/gui/UndoRedoManager.java#L474) is `true`, the event is ignored, and the `UndoRedoManager` doesn't update its state. This creates the issues described in #1487.

Therefore, the issue is not coming from the `UndoRedoManager`, but from the function that created the event.

The event is created in [`PathObjectHierarchy.addObjects()`](https://github.com/qupath/qupath/blob/3544e613b40fd123236936d76e2cb5ee08d855f7/qupath-core/src/main/java/qupath/lib/objects/hierarchy/PathObjectHierarchy.java#L626). I'm sure there is reason behind it, but here a hierarchy changed event is emitted, instead of an object added event. This PR simply emits an object added event instead of a hierarchy changed event.

The `PathObjectHierarchy.addObjects()` function add several objects, but in the PR only one object event is emitted. It seems to be enough still.

### DIFF
--- a/qupath-core/src/main/java/qupath/lib/objects/hierarchy/PathObjectHierarchy.java
+++ b/qupath-core/src/main/java/qupath/lib/objects/hierarchy/PathObjectHierarchy.java
@@ -623,7 +623,7 @@ public final class PathObjectHierarchy implements Serializable {
 			counter++;
 		}
 		if (changes)
-			fireObjectAddedEvent(getRootObject(), pathObjects.iterator().next());
+			fireHierarchyChangedEvent(getRootObject());
 //			fireChangeEvent(getRootObject());
 		return changes;
 	}
--- a/qupath-core/src/main/java/qupath/lib/objects/hierarchy/PathObjectHierarchy.java
+++ b/qupath-core/src/main/java/qupath/lib/objects/hierarchy/PathObjectHierarchy.java
@@ -623,7 +623,7 @@ public final class PathObjectHierarchy implements Serializable {
 			counter++;
 		}
 		if (changes)
-			fireHierarchyChangedEvent(getRootObject());
+			fireObjectAddedEvent(getRootObject(), pathObjects.iterator().next());
 //			fireChangeEvent(getRootObject());
 		return changes;
 	}
--- a/qupath-gui-fx/src/main/java/qupath/lib/gui/UndoRedoManager.java
+++ b/qupath-gui-fx/src/main/java/qupath/lib/gui/UndoRedoManager.java
@@ -471,7 +471,12 @@ public class UndoRedoManager implements ChangeListener<QuPathViewer>, QuPathView
 	@Override
 	public void hierarchyChanged(PathObjectHierarchyEvent event) {
 		// Try to avoid calling too often
-		if (undoingOrRedoing || event.isChanging() || maxUndoHierarchySize.get() <= 0 || event.getChangedObjects().stream().allMatch(p -> p instanceof ParallelTileObject))
+		if (undoingOrRedoing || event.isChanging() || maxUndoHierarchySize.get() <= 0)
+			return;
+
+		// During processing, we have ParallelTileObjects changing to show which part of the image is being handled
+		// - but we don't want to record these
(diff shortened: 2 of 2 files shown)


petebankhead:
Ah, thank you for investigating this!

The hierarchy event system is horrible, and traces back to the early versions of QuPath where the hierarchy was much more important, undo/redo didn't exist, and we didn't have a good spatial cache from JTS. Then, it was a continual fight to avoid very expensive processing every time an object changed - because `resolveHierarchy()` was effectively being called automatically, and lots of things needed to update.

I thinks this shows event system is *very* overdue a major revision.

> I'm sure there is reason behind it, but here a hierarchy changed event is emitted, instead of an object added event.

From your post and my failing memory, I suspect that the idea was that the 'object added' event would just handle a single object, but if multiple objects were added then we fired a more general structure change event.

I guess this was because, when adding a single object, we knew that it could only effect ancestor and descendent objects in QuPath v0.1.2 and earlier. But if we changed multiple objects, then all the *potential* resolved parent/child relationships between objects would be too complex to decipher. Instead, it was easier and safer to fire an event that basically said: _'something big changed, don't try to figure out exactly what, but just update to handle the hierarchy as it now is'_.

I'm reluctant to switch to `addObjects` firing an event that doesn't include all the objects that were added, in case there is any legacy code that might be sensitive to the change. Which leads to...

> So, [this condition](https://github.com/qupath/qupath/blob/3544e613b40fd123236936d76e2cb5ee08d855f7/qupath-gui-fx/src/main/java/qupath/lib/gui/UndoRedoManager.java#L474) is true, the event is ignored, and the UndoRedoManager doesn't update its state.

It looks like I failed to recognize that 
```java
event.getChangedObjects().stream().allMatch(p -> p instanceof ParallelTileObject)
```
would return true when `getChangedObjects().isEmpty()`.

What it be sufficient to change the problematic condition to this?

```java
                // Try to avoid calling too often
		if (undoingOrRedoing || event.isChanging() || maxUndoHierarchySize.get() <= 0)
			return;

		// During processing, we have ParallelTileObjects changing to show which part of the image is being handled
		// - but we don't want to record these
		if (!event.getChangedObjects().isEmpty() && event.getChangedObjects().stream().allMatch(p -> p instanceof ParallelTileObject))
			return;
```

From a quick try, I think that's enough to fix #1487 but you're deeper into the code than me.

I expect that we could *dramatically* simplify `PathObjectHierarchyEvent` for future versions - but if we can make undo/redo less bad with a small change, that would be excellent.

PR #1579, opened 2024-08-08
Avoid server loading during DefaultProject.saveImageData() if it was never loaded before

carlocastoldi:
see https://github.com/qupath/qupath/pull/1489#issuecomment-2276459548


petebankhead:
Thanks, this looks good. I made a few javadoc edits and will merge soon.

carlocastoldi:
Thanks!

And generally, thank you a lot for taking the time and the chance to do such a core change to QuPath's architecture. I am certain it will have big impacts on the users, however i hope it won't weight too much on the maintenance burdening.

PR #1583, opened 2024-08-09
Added possibility to copy measurement map color map

### LINKED ISSUE #1221: Feature request: Colorbar for measurements maps
Hi

It would be nice to have an option to show colorbar/calibration bar for measurement map in the image panel similar to showing the scalebar.

Thanks
Ofra

Rylern:
A proposal to fix #1221.

It simply adds the possibility to right click on the color map of the measurement map window, this opens a context menu with a `Copy` option, and clicking on this option will copy the color map to the clipboard.


petebankhead:
Looks good, I think this will be helpful for people creating figures. They should be able to resize / rotate etc. without worrying about aspect ratio to get the appearance they need.

Issue #1537, opened 2024-05-30
Improvements to project sorting


alanocallaghan:
## Feature request

**Is your feature request related to a problem? Please describe.**
Project sorting is not persistent across QuPath instances, and does not update based on updated metadata values.
Nice-to-have would be to add metadata values when images are dragged and dropped, but I feel that's going to be a stretch goal.

**Describe the solution you'd like**
In order of priority/feasibility:

1. Project sorting persists
2. Sorting is updated when metadata values change
3. Dragging & dropping a file into a project sorted by metadata value sets the corresponding value.

**Describe alternatives you've considered**
Manually re-sorting on project opening and when values change, and manually setting metadata values.

**Additional context**

ImageSC posts:

- https://forum.image.sc/t/qupath-feature-request-persist-image-sorting-per-metadata/91406/4
- https://forum.image.sc/t/qupath-sort-by/85398

alanocallaghan:
Can somewhat resolve item 2 by adding this:

```java
				ImageRow selectedImageRow = getSelectedImageRow();
				refreshTree(selectedImageRow);
```
to the end of `actionAddMetadataValue`, although this isn't a particularly robust way of doing it in case metadata are edited elsewhere

petebankhead:
Yes. We can make changes, but need to be cautious and ideally keep the `Project` interface as simple as we can.

The thinking is that we want projects to not *necessarily* be backed by files on the local file system. So you might have a QuPath project that behaves like a 'view' on an OMERO project (for example).

I think this idea is already used by QuPath Edu - see [here](https://github.com/openmicroanatomy/qupath-edu-extension/blob/dd67a5edccd8e548dd0739c6bde624fff276b21f/src/main/java/qupath/edu/EduProject.java#L54).

alanocallaghan:
An alternative could be to smuggle in an implementation of stored sorting with a `Manager`, but that seems like a bit of abuse of the existing facilities.

The obvious way to make one change that permits any future extensions is to add a `Map<[String?], [Object?]>`, but that's got its own set of problems (security would be a big one).

petebankhead:
Yes, map is tempting but a bit scary. 
Somewhat relevant: https://github.com/qupath/qupath/issues/1289

alanocallaghan:
> smuggle in an implementation of stored sorting with a Manager, but that seems like a bit of abuse of the existing facilities.

[...]

Issue #1289, opened 2023-08-11
Adding a tag field to handle OMERO tags


Rdornier:
## Feature request

**Issue**
Metadata fields are perfectly suitable for OMERO Key-value pairs but there is no way to propertly handle OMERO tags in QuPath.

**Description of the solution**
As tags are very useful in OMERO to search for images, it could be great to also have a tag field in QuPath.

Tag field attributes
- value (unique => no duplicate possible)
- used to sort the images within the project
- maybe other attributes mirorred from the Metadata field object that I'm not aware of.

petebankhead:
Can you explain in more detail what would need to change to support this?

I don't have much experience of OMERO tags, but it sounds like what you're suggesting could either be handled entirely in the OMERO extension already (using QuPath's existing metadata support for project entries), or else could risk projects becoming more OMERO-specific (which I'd like to avoid).

Rdornier:
Actually, tags are one key-word annotation. 
Currently, the metadata field works as a dictionnary. My suggestion is to add a one-keyword metadata field

> it sounds like what you're suggesting could either be handled entirely in the OMERO extension already

Maybe...  I don't think it is optimal nut having the same value as the key could work, using existing QuPath's metadata

petebankhead:
> Actually, tags are one key-word annotation.

Ah... makes sense.

> My suggestion is to add a one-keyword metadata field

Sounds sensible, but I wonder is it possible that people will then want tags with different interpretations?

The alternative I'm thinking is to simply have a dictionary entry 
```json
{
   "omeroTags": "tag1, tag2, tag3"
}
```
but I don't know whether it's safe to use a comma (or anything else) as a delimiter.

Otherwise, I guess there could be a general `tags` list in a QuPath project entry.

Do you think either option is strongly preferable to the other?

Rdornier:
> but I wonder is it possible that people will then want tags with different interpretations?

Yes, this is also true.

> The alternative I'm thinking is to simply have a dictionary entry
> {
>    "omeroTags": "tag1, tag2, tag3"
> }
> 
> but I don't know whether it's safe to use a comma (or anything else) as a delimiter.

Actually, I already thought of the solution but our final goal would be to sort the images within the current project according to one or more tags.

> Otherwise, I guess there could be a general tags list in a QuPath project entry.

I would rather go for this solution as I think it is the best way to make them independant and searchable.

petebankhead:
How would the sorting look? Would there need to be multi-level sorting?

[...]

PR #1584, opened 2024-08-09
Project sorting improvements

Rylern:
This PR is linked with issues #1537 and #1289 and implements [Alan's idea](https://github.com/qupath/qupath/issues/1537#issuecomment-2141451238).

It adds a [`Manager`](https://github.com/qupath/qupath/blob/7b272e5ee4c79d27dcf4e332d26ea4d0e5c80e3a/qupath-core/src/main/java/qupath/lib/projects/ResourceManager.java#L69) to the [`Project`](https://github.com/qupath/qupath/blob/main/qupath-core/src/main/java/qupath/lib/projects/Project.java) interface through the `Project.getSortingKeys()` function. The names of this manager correspond to sorting keys, while their values are string representations of booleans corresponding to whether the key is in ascending or descending order.

The current state of this PR makes project sorting persists when projects are closed / reopened.

It also enable to store several sorting keys and to define the sorting order (ascending or descending) for each of these keys, so that we could achieve this:
image
 
However, the UI has to be changed to allow this. I can work on it, but I wanted to have your opinions on the current state of this PR first.

### DIFF
--- a/qupath-core/src/main/java/qupath/lib/projects/Project.java
+++ b/qupath-core/src/main/java/qupath/lib/projects/Project.java
@@ -268,6 +268,20 @@ public interface Project<T> {
 	public default Manager<PixelClassifier> getPixelClassifiers() {
 		return getResources(PixelClassifier.PROJECT_LOCATION, PixelClassifier.class, "json");
 	}
+
+	/**
+	 * Get the manager for sort keys saved within this project.
+	 * <p>
+	 * Sort keys can be used to group and sort project entries.
+	 * <p>
+	 * The names of this manager correspond to keys, while their values are string representations
+	 * of booleans corresponding to whether the key is in ascending or descending order.
+	 *
+	 * @return the manager for sort keys
+	 */
+	default Manager<String> getSortKeys() {
+		return getResources("sort_keys", String.class, "txt");
+	}
 	
 	/**
 	 * Get a manager for objects of a specified class within this project.
--- a/qupath-core/src/main/java/qupath/lib/projects/Project.java
+++ b/qupath-core/src/main/java/qupath/lib/projects/Project.java
@@ -277,7 +277,7 @@ public interface Project<T> {
 	 * The names of this manager correspond to keys, while their values are string representations
 	 * of booleans corresponding to whether the key is in ascending or descending order.
 	 *
-	 * @return the manager for sort keys
+	 * @return the manager for sort keys, or {@code null} if the project does not support storing sort keys
 	 */
 	default Manager<String> getSortKeys() {
 		return getResources("sort_keys", String.class, "txt");
--- a/qupath-core/src/main/java/qupath/lib/projects/Project.java
+++ b/qupath-core/src/main/java/qupath/lib/projects/Project.java
@@ -28,6 +28,7 @@ import java.net.URI;
 import java.nio.file.Path;
 import java.util.Collection;
 import java.util.List;
+import java.util.Optional;
 
 import qupath.lib.classifiers.object.ObjectClassifier;
 import qupath.lib.classifiers.pixel.PixelClassifier;
@@ -268,20 +269,6 @@ public interface Project<T> {
 	public default Manager<PixelClassifier> getPixelClassifiers() {
 		return getResources(PixelClassifier.PROJECT_LOCATION, PixelClassifier.class, "json");
 	}
-
-	/**
-	 * Get the manager for sorting keys saved within this project.
-	 * <p>
-	 * Sorting keys can be used to group and sort project entries.
-	 * <p>
-	 * The names of this manager correspond to sorting keys, while their values are string representations
-	 * of booleans corresponding to whether the key is in ascending or descending order.
-	 *
-	 * @return the manager for sorting keys, or {@code null} if the project does not support storing sorting keys
-	 */
-	default Manager<String> getSortingKeys() {
-		return getResources("sorting_keys", String.class, "txt");
-	}
 	
 	/**
 	 * Get a manager for objects of a specified class within this project.
@@ -297,13 +284,29 @@ public interface Project<T> {
 	public default <S, R extends S> Manager<R> getResources(String location, Class<S> cls, String ext) {
 		return null;
 	}
-	
-	
-//	public List<String> listPixelClassifiers();
-//	
-//	public PixelClassifier loadPixelClassifier(String name);
-//	
-//	public void savePixelClassifier(String name, String PixelClassifier);
 
-	
+	/**
+	 * Store a metadata key-value pair in this project.
+	 * If a value already exists for the provided key, it will be overridden.
+	 *
+	 * @param key the key to store
+	 * @param value the value to store
+     */
+	void putMetadataValue(String key, String value);
+
+	/**
+	 * Request a metadata value stored in this project.
+	 *
+	 * @param key the key associated with the value to retrieve
+	 * @return the value associated with the provided key, or an empty Optional if no such value exists
+	 */
+	Optional<String> getMetadataValue(String key);
+
+	/**
+	 * Remove a metadata key-value pair from this project.
+	 * This function does nothing if the provided key is not present in this project.
+	 *
+	 * @param key the key associated with the value to remove
+	 */
+	void removeMetadataValue(String key);
 }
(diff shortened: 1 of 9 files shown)


petebankhead:
The use of a manager still feels a bit strange to me, but I'm open to being convinced.

It's a reasonable mechanism if we want to ensure that sorting is something handled *entirely* by the UI, and isn't an inherent part of projects. But then we should use the existing [`Project.getResources(String, Class, String)`](https://qupath.github.io/javadoc/docs/qupath/lib/projects/Project.html#getResources(java.lang.String,java.lang.Class,java.lang.String)) rather than introduce a new `Project.getSortingKeys()` to the `Project` interface.

I think it's also preferable to store all such resources within a `resources` subdirectory, as done [here](https://github.com/qupath/qupath/blob/db71980322fbdcb47ad54599ef71d21d8f2c5f94/qupath-gui-fx/src/main/java/qupath/lib/display/settings/DisplaySettingUtils.java#L201). With this PR, I get a new `sorting_keys` directory directly within the project. I also got a file `SORT_KEY[IMAGE_NAME].txt` with the only content `true`; I was expecting e.g. a `sort_keys.txt` file with a list of the keys that are used.

The purpose of `ResourceManager` (for me at least) is that it supports having several named resources, typically implementations of a specific serializable class. So we might have something like this:

```java
class SortingKeys {
   List<String> keys;

   List<String> getKeys() {
      return Collections.unmodifiableList(keys);
   }

}

var sortingManager = project.getResources("resources/sorting_keys", SortingKeys.class, "json");
var sortKeys = sortingManager.contains("sortKeys") ? sortingManager.getResource("sortKeys).getKeys() : Collections.emtpyList();
```
but then it is still 'unconventional' to use a `Manager` when we only have a single resource with a fixed name (here, `"sortKeys"`).

---

What not use the alternative approach of extending the `Project` interface to implement [`MetadataStore`](https://github.com/qupath/qupath/blob/df21dcbaab953837d40aabd856d14b79faca6b7b/qupath-core/src/main/java/qupath/lib/objects/MetadataStore.java#L37) - or adding `String` metadata support similar to how it already works with [`ProjectImageEntry`](https://github.com/qupath/qupath/blob/df21dcbaab953837d40aabd856d14b79faca6b7b/qupath-core/src/main/java/qupath/lib/projects/ProjectImageEntry.java#L49)?

Then the UI can store keys as a list of strings easily, with any key of its choosing and the result serialized within the project.

To me, a `Map<String, String>` feels simpler and more intuitive. The advantages I see of a `Manager` are:
1. It avoids adding to the `project.qpproj` file size (in case someone attempts to add ridiculously-large strings)
2. The sort order wouldn't be lost if a project is saved in an older version of QuPath

But these are only an issue of the `Map` is serialized directly within `project.qpproj`. We *could* potentially overcome both of these by storing the `Map` in a separate `.json` file anyway within `resources/sorting` - without needing to rely on `Manager` to do so.

What do you think?

qupath-core/src/main/java/qupath/lib/projects/Project.java
```diff
@@ -283,13 +284,29 @@ public default Manager<PixelClassifier> getPixelClassifiers() {
...
+     */
+	void putMetadataValue(String key, String value);
+
+	/**
+	 * Request a metadata value stored in this project.
+	 *
+	 * @param key the key associated with the value to retrieve
+	 * @return the value associated with the provided key, or an empty Optional if no such value exists
+	 */
+	Optional<String> getMetadataValue(String key);
```
petebankhead, line 303: I'm not sure `Optional` is required - the defined behavior for a `Map` is to return `null` if there is no value for the key.

The `Optional` doesn't resolve the ambiguity as to whether there is no value, or if the stored value is `null` (and there is no method to query `contains`).

petebankhead (changes requested):
I've added some comments.

One lingering question is to what extent this should be modelled on the [`Map` interface](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/Map.html), with similar behavior.

I wasn't entirely consistent myself, and we currently have:
* A [`MetadataStore`](https://github.com/qupath/qupath/blob/df21dcbaab953837d40aabd856d14b79faca6b7b/qupath-core/src/main/java/qupath/lib/objects/MetadataStore.java#L37), which is used by `PathObject` implementations
* Similar metadata support within [`ProjectImageEntry`](https://github.com/qupath/qupath/blob/df21dcbaab953837d40aabd856d14b79faca6b7b/qupath-core/src/main/java/qupath/lib/projects/ProjectImageEntry.java#L49)... but without implementing `MetadataStore`.

I think we should try to use this as an opportunity to improve consistency rather than have `Project` diverge further.

With that in mind, I think here we should do one of these:
* Use the `MetadataStore` approach, i.e. `Project extends MetadataStore`
* Use the `ProjectImageEntry` approach (without extending anything, but ensuring consistency of method names and signatures)
* Do *both* by updating `ProjectImageEntry` to extend `MetadataStore`
* Ignore the existing QuPath-specific strategies, and instead add a `Map<String, String> getMetadata()` method. This would return a 'proper' `Map` implementation that the caller can freely modify, and then rely on `project.syncChanges()` to make any metadata changes persistent.

The last option is quite convenient, since it provides useful `putAll` and `computeIfAbsent` methods - but may lead to `project.qpproj` being written more than it needs to be.

Rylern:
> Do both by updating ProjectImageEntry to extend MetadataStore

I think I'd prefer that option. The only problem I had with `MetadataStore` is that it doesn't have a method to delete a value. Can I add one?

petebankhead:
We can potentially be more ruthless than that... [`MetadataStore`](https://qupath.github.io/javadoc/docs/qupath/lib/objects/MetadataStore.html) currently supports non-String objects - but, weirdly, contains the method
```java
Map<String, String> getMetadataMap()
```
that assumes `String` only. The map is documented to be unmodifiable, but [at least once in QuPath's code it is assumed to be modifiable](https://github.com/qupath/qupath/blob/df21dcbaab953837d40aabd856d14b79faca6b7b/qupath-core/src/main/java/qupath/lib/objects/PathObjectTools.java#L1527).

*If* we don't need non-string objects (we'd need to check usages), I am thinking that we could deprecate all other methods in `MetadataStore`, leaving only `getMetadataMap()` to return a *modifiable* `Map<String, String>` - and provide default implementations for the other methods.

Then, in this case, you could potentially provide a `Map` implementation that takes care of JSON serialization by subclassing [`AbstractMap`](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/Map.html#get(java.lang.Object)).

Since I was responsible for the original `MetadataStore` mess (and the associated [`MetadataMap`](https://github.com/qupath/qupath/blob/df21dcbaab953837d40aabd856d14b79faca6b7b/qupath-core/src/main/java/qupath/lib/objects/MetadataMap.java#L49)), I can look into how changing the interface might impact `PathObject`, where it is currently used.

*If* you think this is a sensible approach, then you could proceed here by adding
```java
public Map<String, String> getMetadataMap() // Or getMetadata() - see below
```
and handling its implementation and (JSON) serialization - then any common interface that includes that method can be added later to `Project`, `ProjectImageEntry` and `PathObject`.

What do you think?

One benefit of using a standard `Map` implementation is that Groovy scripting is nicer, e.g. see
* https://github.com/qupath/qupath/issues/1085

which is why we can now use `pathObject.measurements['Something'] = 2.0`.

In fact, if we simplified the name further to `getMetadata()` the this should work in Groovy:
```
pathObject.metadata['Something'] = 'A value'
project.metadata['Something'] = 'A value'
entry.metadata['Something'] = 'A value'
```

Rylern:
OK. I just have a question: if a class implementing `MetadataStore` returns a modifiable `Map<String, String>`, then the class won't know when the Map is modified. Isn't it a problem for situations where modifying the map should have side effects (such as saving the map to a file for example)?

Rylern:
I updated the `MetadataStore` and changed `Project` and `DefaultProject` accordingly.

I didn't update other classes that were using `MetadataStore` like `TMACoreObject`, and I didn't make `ProjectImageEntry` implements `MetadataStore`. Should I do that here, or create another PR?

petebankhead:
> Should I do that here, or create another PR?

No, I think this PR should focus on adding `Project.getMetadata()` - anything else can/should be done later.

Before going further with the general metadata support, I'd like to make sure that I was never using an `Object` with the `MetadataStore` within `PathObject`, in case that could be an issue. We also discussed having a new repo and module (`qupath-lib` perhaps?) similar to `qupath-fxtras` that contains generally-useful stuff that we can make permissively-licensed (i.e. Apache rather than GPL). I'd like to investigate adding a new, simple `MetadataStore` interface to that module, so that we can deprecate the old interface entirely - but I don't want to rush that, and need to find an hour or two to think about it properly.

petebankhead (approved):

PR #1588, opened 2024-08-16
PathClass in TMA grid view

alanocallaghan:
Should externalise strings and also call it "Classification" rather than "PathClass"


petebankhead:
I tried this with a TMA grid, and 2 cores classified: one tumor, one stroma.

The behavior was unexpected, in that there didn't see a way for me to view all the unclassified cores as well: only two appeared.

Note that this affects *Measure → Grid views... → Show annotation grid view* as well.

Apart from that, this change looks very useful.

alanocallaghan:
That was in keeping with the existing behavior for missing measurements, I think, although I agree it's not optimal

petebankhead:
Ah, maybe we should make it possible to show missing measurements. I don't have a strong opinion on that, since sorting with NaNs is a whole new consideration.

But I think having unclassified objects is going to be much common than having missing measurements. Maybe just adding that as an option in the combo box is enough, with 'Unclassified' turned on by default.

alanocallaghan:
Sorting NaNs was already implemented, it was just whether or not we filter them out first, which I think nah. Although I'd prefer keeping them always at the back which is not current behaviour.

Can write up properly in PR description and changelog before merging if you like

petebankhead:
Mmmm, I find I have no way to hide the unclassified TMA cores. And sorting my ascending/descending seems to only impact the two that I gave classes to - the rest remain stubbornly as they were (both before and after the classified core) - so I initially didn't think sorting was implemented.

Since we aren't dealing with NaNs, I have the feeling that string sorting would make sense here... even if that means 'Unclassified' being sorted as a string.

And actually... what do you think about having 'Name' as another potential 'Measurement'... which we could relabel 'Sort by'? Then we could change 'Show measurement' to 'Show value' - both to save space, and because we're no longer limited to measurements.

The name is especially informative for TMA cores, but it looks like we don't currently have a way to display it.

Minor thing: when I assign a classification to a core _after_ displaying the grid, this becomes available in the 'Classes' drop-down - which is nice. The less nice thing is that all the classes there appear deselected.... but aren't necessarily not visible.

alanocallaghan:
> Mmmm, I find I have no way to hide the unclassified TMA cores. And sorting my ascending/descending seems to only impact the two that I gave classes to - the rest remain stubbornly as they were (both before and after the classified core) - so I initially didn't think sorting was implemented.

Ah yes, I think I misunderstood what my sorting logic was doing; should now be fixed.

> Since we aren't dealing with NaNs, I have the feeling that string sorting would make sense here... even if that means 'Unclassified' being sorted as a string.

That's intuitive to me, although it does somewhat point to the odd behaviour about unclassified things having a null class, or `PathClass.NULL_CLASS`...

> And actually... what do you think about having 'Name' as another potential 'Measurement'... which we could relabel 'Sort by'? Then we could change 'Show measurement' to 'Show value' - both to save space, and because we're no longer limited to measurements.

Sounds good, done.

> Minor thing: when I assign a classification to a core after displaying the grid, this becomes available in the 'Classes' drop-down - which is nice. The less nice thing is that all the classes there appear deselected.... but aren't necessarily not visible.

This should be fixed now, with only minor cursing at ControlsFX

Issue #1590, opened 2024-08-16
Questionable behavior with Delaunay cluster features


petebankhead:
## Bug report

**Describe the bug**
The behavior of *Analyze → Spatial analysis → Delaunay cluster features 2D* is not well-defined when applying the command to annotations that are nested (in the hierarchy) or overlapping.

For 'flat' annotation arrangements the behavior is *mostly ok*, except that it does rely on *hierarchy relationships* rather than *spatial relationships*.
Which is to say, it hasn't really changed since v0.1.2.

**To Reproduce**
Create an arrangement of objects similar to that shown below:

![0-orig](https://github.com/user-attachments/assets/cf1a87fc-c73a-4ac6-be24-d830f0014604)

1. If I run *Delaunay cluster features 2D* with **_all_ annotations selected**, I see triangulation lines which *do not cross the boundary between the annotations containing cells*.

![1-all selected](https://github.com/user-attachments/assets/3309e1e3-76ca-4ec7-bf5c-e0c76b8f2d96)

But I **also** get a `ConcurrentModificationException`:
```
15:47:01.829	[Plugin thread]	ERROR	qupath.lib.plugins.AbstractTaskRunner	Error running plugin: java.util.ConcurrentModificationException	java.util.concurrent.ExecutionException: java.util.ConcurrentModificationException
	at java.base/java.util.concurrent.FutureTask.report(Unknown Source)
	at java.base/java.util.concurrent.FutureTask.get(Unknown Source)
	at qupath.lib.plugins.AbstractTaskRunner.awaitCompletion(AbstractTaskRunner.java:147)
	at qupath.lib.plugins.AbstractTaskRunner.runTasks(AbstractTaskRunner.java:117)
	at qupath.lib.gui.TaskRunnerFX.runTasks(TaskRunnerFX.java:106)
	at qupath.lib.plugins.AbstractPlugin.runPlugin(AbstractPlugin.java:147)
	at qupath.lib.gui.ParameterDialogWrapper$1.run(ParameterDialogWrapper.java:177)
	at java.base/java.lang.Thread.run(Unknown Source)
Caused by: java.util.ConcurrentModificationException
	at java.base/java.util.ArrayList$Itr.checkForComodification(Unknown Source)
	at java.base/java.util.ArrayList$Itr.next(Unknown Source)
	at java.base/java.util.Collections$UnmodifiableCollection$1.next(Unknown Source)
	at java.base/java.util.AbstractCollection.addAll(Unknown Source)
[...]

PR #1597, opened 2024-08-20
Add findNeighbors support to hierarchy

petebankhead:
Proposed support to make it easy to find 'neighbors' for objects based on a Delaunay triangulation.

This lazily builds a subdivision based on object centroids within the `PathObjectHierarchy` class. Even for >200k objects this only takes a couple of seconds on my computer, and subsequent queries are very fast.

Here's an example script that counts the number of neighbors, and adds it as a measurement to each object:
```groovy
def hierarchy = getCurrentHierarchy()
def cells = getCellObjects()
for (cell in cells) {
    cell.measurements['Num neighbors'] = hierarchy.findAllNeighbors(cell).size()
}
```

Or this adds measurements giving counts of immediate neighbors with specific classifications:
```groovy
def hierarchy = getCurrentHierarchy()
def cells = getCellObjects()
def allClassifications = getCellObjects().collect(c -> c.classification) as TreeSet
for (cell in cells) {
    def counts = hierarchy.findAllNeighbors(cell).countBy(n -> n.classification)
    for (classification in allClassifications) {
        cell.measurements["Num ${classification} neighbors"] = counts.getOrDefault(classification, 0)
    }
}
```

Given that the existing Delaunay command is rather buggy https://github.com/qupath/qupath/issues/1590 this could give a convenient alternative. If we need tests such as ensuring the same parent object then this can be implemented, e.g.
```
def sharedParent = hierarchy.findAllNeighbors(cell).findAll {c -> c.parent == cell.parent}
```

## Assumptions & limitations

In this current (draft) form, there are some assumptions:
* The triangulation involves all objects with the exact same Java class (e.g. `PathDetectionObject`, `PathCellObject`, `PathTileObject`, `PathAnnotationObject`)
* There is nothing to prevent connections being made *across parent objects*. So if there were disconnected pieces of tissue - or even TMA cores - neighbors would still be found across these.
* Everything is done in 2D (this is all JTS's Delaunay triangulation supports anyway)
* All measurements are in pixel units, *and pixels are assumed to be square*

The last of these is because the calculations are done within `PathObjectHierarchy`, which has no access to the `PixelCalibration` object that stores pixel sizes.

We could move this into `ImageData` so that pixel calibration could be accessed, but it feels to me less intuitive to have it there.

I'm not sure if this could become problematic at some point though, so the PR is flagged as a draft - subject to more development & discussion.

## Possible improvements

Potential developments to consider:
* Prevent connections being made across TMA cores
* Use annotation boundaries, not centroids
* What should happen with points annotations...?
* Introduce other spatial querying to `PathObjectHierarchy` (e.g. find the closest object meeting specific criteria)
* Enable a Delaunay overlay using these computations, rather than the existing (awkward, buggy-plugin-derived) strategy

## Alternative

`DelaunayTools` can already be used to create a `Subdivision`, and this used to get all the same info - and that *does* provide a way to incorporate `PixelCalibration`, and introduce more criteria to get more exactly.

So the main decision is whether it is worthwhile having a simple version of Delaunay triangulation built-in to the `PathObjectHierarchy` to make it very easy to use for common cases... or whether it just complicates the class too much.


petebankhead:
The last commit adds a toolbar button to show detection 'connections' (neighbors based on a Delaunay triangulation):

image

This was previously available under the *View* command, but only actively did anything if the *Delaunay cluster features 2D* command had been run. Now, it can be used at any time and will show connecting lines for cells (if available), or detections (if there are no cells).

Currently, it is only a visual tool - using the new features require scripts such as those described above.

lacan:
I am encountering a bug
1. Detect cells using something (StarDist, I did it with no cell expansion)
2. Run the script 
```
def hierarchy = getCurrentHierarchy()
def cells = getDetectionObjects().findAll{ sel.getROI().contains( it.getROI().getCentroidX(), it.getROI().getCentroidY() ) }
for (cell in cells) {
    cell.measurements['Num neighbors'] = hierarchy.findAllNeighbors(cell).size()       
}
```
3. Try to remove the neighbors in another script
```
def hierarchy = getCurrentHierarchy()
hierarchy.resetNeighbors()
```
4. Delete the cells and recompute them (by running StarDist again on the same region)
5. The cells still have the connections shown
6. Run cell detection in a *new* region: The new cells also have connections
7. Open another image that already had cells detected. These cells also have connections

petebankhead:
Have you run the (deprecated) _Analyze → Spatial analysis → Delaunay cluster features 2D_ command on any of these images, at any time?

If so, this can mess things up...

When using the new command only, connections should be generated any time they are requested (including for visualisation). Resetting them only serves to request that they are recalculated on the next request.

This behavior is entirely different from the old _Delaunay cluster features 2D_ command, where you had to explicitly request connections to be made. The goal here was to avoid the need for that extra step, so that querying connections is _always_ possible.

(But this is also why the work is unfinished... because we don't have a good way to add in extra criteria for the visualisation, such as only generating connections less than a specified distance)

If you've run _Delaunay cluster features 2D_ then this causes trouble because it sets a property within `ImageData`, and currently this is visualised instead when it's present. Sometimes the property gets out of sync with the original data, which is even worse, but that isn't a new bug.. and contributes to the reasons why _Delaunay cluster features 2D_ is deprecated.

lacan:
I had not run the command to my knowledge, but seeing as it was an old QuPath "Test project" that I keep around it is entirely possible that I *had* run Delaunay at one point in the past. Thanks for the information.

I see, so once the neighborhoot information has been asked once, then it will keep on computing it, no matter what, and then we can do what we want with it, with the current limitations you mentioned. 
Thank you for the information

Oli

PR #1601, opened 2024-08-23
Make viewer options persistent

petebankhead:
Addresses https://forum.image.sc/t/removing-the-cursor-location-ui-element-via-script/100008


alanocallaghan:
What's the motivation in switching from 
```java
public SimpleBooleanProperty(boolean initialValue)
```
to
```java
public SimpleBooleanProperty(Object bean,
                             String name,
                             boolean initialValue)
```
?

petebankhead:
The lazy code further down where I loop through properties and make them persistent there. Using the name makes it easier to assign them a sensible persistent key.

alanocallaghan:
Of course, sorry I thought there was also a `name, initialValue` constructor and thought the null bean had some function

PR #1603, opened 2024-08-25
Update to Java 21

petebankhead:
We already default to using Java 21, but we were still building artifacts with Java 17. This meant we couldn't actually use any features from after Java 17.

Things we can now use (roughly in order of relevance):
- Sequenced collections - https://openjdk.org/jeps/431
- Code snippets in Javadocs - https://openjdk.org/jeps/431
- Pattern matching for switch - https://openjdk.org/jeps/441
- Virtual threads - https://openjdk.org/jeps/444
- Record patterns - https://openjdk.org/jeps/440

Note that this may require updates in any extensions that declare v0.6.0 dependencies, to ensure they also use Java 21 (and have a suitably recent Gradle wrapper).

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -14,14 +14,14 @@
     runs-on: ${{ matrix.platform }}
     steps:
     - uses: actions/checkout@v3
-    - name: Set up JDK 17
+    - name: Set up JDK 21
       uses: actions/setup-java@v3
       with:
-        java-version: '17'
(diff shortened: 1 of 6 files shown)


alanocallaghan:
Nested record patterns seem a feature I would rather avoid

petebankhead:
You don't appreciate the readability of `r instanceof Rectangle(ColoredPoint(Point p, Color c), ColoredPoint lr)`?

PR #1604, opened 2024-08-26
Reinstate mergedJavadocs

petebankhead:
Removed `mergedJavadocs` gradle task with https://github.com/qupath/qupath/pull/1513 but it turns out that we do need it for a while longer.

It is used to generate the javadocs that are hosted online, and as part of the `jpackage` GitHub workflow.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,6 +7,36 @@ plugins {
 // We don't want to generate javadocs for the root project
 javadoc.enabled = false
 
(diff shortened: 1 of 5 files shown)


alanocallaghan:
Ah, yes I was trying to update the javadocs site/action to host multiple versions (a la https://qupath.github.io/qupath-fxtras/stable/qupath.fx/module-summary.html) but without mergedJavadocs, I couldn't get my head around gradle well enough to recreate it from the component pieces

Issue #1586, opened 2024-08-14
QuPath unable to open images using symbolic links


Rafael-Silva-Oliveira:
Hey

I'm trying to open an image through a symlink. I've created symlinks for both the folders with the metadata and the .vsi files, but I'm getting an error:

![image](https://github.com/user-attachments/assets/329325f0-7e2c-4c82-b9f4-76f0745578e6)

How can one create/open symlinks? I can open the images on the original paths just fine, but can't open the symlinks

Thanks

petebankhead:
I've merged a proposed fix at https://github.com/qupath/qupath/pull/1620 which will be in the next QuPath release.

PR #1620, opened 2024-09-04
Better handle symlinks

petebankhead:
Proposed fix for https://github.com/qupath/qupath/issues/1586

@alanocallaghan @Rylern please test this and let me know what you think, I don't use symlinks much...

Ideal tests should involve a range of file types, including images and .qpdata, via drag & drop and opened in other ways.

In particular, the original report related to an image that involves multiple files (.vsi). I don't know of good .vsi examples are public, but .ndpis is similar-ish: https://downloads.openmicroscopy.org/images/Hamamatsu-NDPI/manuel/

Then you have 3 .ndpi files that *could* be opened independently, but if you open the .ndpis then the other files should act as different channels. The trouble for me is that it won't work on an Apple Silicon Mac due to Bio-Formats - it needs an Intel build.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ This is a *work in progress* for the next major release.
     * If you use this command, the calculated connections are displayed instead of the default neighbor connections for compatibility.
       However, this support will be removed in a future version.
 * More viewer options are persistent (e.g. show/hide the overview thumbnail, location text, or scalebar)
+* Better support for symbolic links (https://github.com/qupath/qupath/issues/1586)
 
 ### Bugs fixed
 * Tile export to .ome.tif can convert to 8-bit unnecessarily (https://github.com/qupath/qupath/issues/1494)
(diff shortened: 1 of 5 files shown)


Rylern (approved):
I tested on several images (including zarr), .qpdata, on a qupath project and it worked.

> Then you have 3 .ndpi files that could be opened independently, but if you open the .ndpis then the other files should act as different channels. The trouble for me is that it won't work on an Apple Silicon Mac due to Bio-Formats - it needs an Intel build.

Is opening the `test3.ndpis` file supposed to work?  I've got the no supported image reader found message (the `test3.ndpis` file and the `test3-*.tiff` files are in the same directory).

petebankhead:
`.ndpis` should work on Windows and Linux, and with an Intel build on Mac - but not the Apple Silicon build.

But the extra files should have the extension `.ndpi` - if you downloaded them on a Mac, the OS might automatically have added `.tiff` to the name and this needs to be removed.

Rylern:
Here is what I have:

image

petebankhead:
Can you rename `.tiff` to `.ndpi`? They are `.ndpi` via the link, and that's what they should be.

alanocallaghan:
Opening an NDPI with symlinks already worked, as long as the link had the right extension (regardless of whether the linked file did).

Rylern:
> Can you rename .tiff to .ndpi? They are .ndpi via the link, and that's what they should be.

OK I didn't understand that, but I still got the error:

image

Anyway this is not related to this PR, so we can skip that for now.

PR #1621, opened 2024-09-04
Introduce ObjectProcessor interface

petebankhead:
This enable `ObjectMerger` and `OverlapFixer` to implement the same interface, and means they can easily be applied sequentially.

Then the `PixelProcessor` class was updated to support an `ObjectProcessor` to handle detections made across tiles, rather than an `ObjectMerger` - since that gives more flexibility in dealing with the output.

### DIFF
--- a/qupath-core-processing/src/main/java/qupath/lib/experimental/pixels/PixelProcessor.java
+++ b/qupath-core-processing/src/main/java/qupath/lib/experimental/pixels/PixelProcessor.java
@@ -2,7 +2,7 @@
  * #%L
  * This file is part of QuPath.
  * %%
- * Copyright (C) 2023 QuPath developers, The University of Edinburgh
+ * Copyright (C) 2024 QuPath developers, The University of Edinburgh
  * %%
  * QuPath is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as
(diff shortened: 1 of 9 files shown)


qupath-core-processing/src/main/java/qupath/lib/experimental/pixels/PixelProcessor.java
```diff
@@ -86,14 +87,14 @@ public class PixelProcessor<S, T, U> {
...
-    private final ObjectMerger merger;
+    private final ObjectProcessor objectProcessor;
 
     private PixelProcessor(ImageSupplier<S> imageSupplier,
                           MaskSupplier<S, T> maskSupplier,
                           OutputHandler<S, T, U> outputHandler,
                           Processor<S, T, U> processor,
                           Tiler tiler,
-                          ObjectMerger merger,
+                           ObjectProcessor objectProcessor,
```
alanocallaghan, line 97:
```suggestion
                          ObjectProcessor objectProcessor,
```

alanocallaghan:
Looks good, I think OverlapFixer needs a better name but that's about it

petebankhead:
Yeah, I don't like it either - suggestions welcome!

PR #1623, opened 2024-09-05
Change detection line thickness when upsampling

petebankhead:
Possibly controversial, possibly very welcome...

Usually, the annotation line thickness remains constant at all zoom levels - whereas the detection line thickness increases and decreases when zooming in and out.

This is partly to avoid having thick detection lines when zoomed out, and partly for performance reasons: we cache detections rendered at different resolutions.

*However*, detections *are* painted 'live' (like annotations) when zooming in beyond full resolution.

I find this can be annoying at times, because the default detection line thickness of 2 can be too much - and obscures details. I need to toggle detections on and off to see the pixels that overlap with the border, or else go to the preferences to reduce the line for some images... then go back to increase the line for the next images.

So this PR causes detections to be painted more like annotations when upsampling. The result is that lines narrow to become subpixel when zooming in, and that it is possible to more accurately judge where the detection boundary really lies.

Screenshots below, but they aren't the best examples - and it's better to play around with the behavior using different images to form an opinion as to whether this is better, worse, or pretty neutral.

## Previous

older-rendering

## With this PR

new-rendering

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,13 +34,18 @@ This is a *work in progress* for the next major release.
   * New icon & other drawing icons change to indicate when they are active in selection mode
   * Selection mode works with line ROIs, selecting any intersecting objects
   * Temporarily active 'Selection mode' by pressing the `S` key while interacting with a viewer
+* More viewer options are persistent (e.g. show/hide the overview thumbnail, location text, or scalebar)
+* Better support for symbolic links (https://github.com/qupath/qupath/issues/1586)
+
+### Experimental features
+These features are included for testing and feedback.
+They may change or be removed in future versions.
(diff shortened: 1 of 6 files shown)


petebankhead:
Another example... zoomed in much more, and the distinction is more obvious when the boundaries are tracing pixels.

## Previous

Older-unconvincing

### With this PR

Newer-questionable

Note that when viewing at a downsample ≥ 1, there is no difference in behavior.

petebankhead:
I've added a new preference so that this can be turned on/off, to help us explore if we like it or not.

(If it stays, we might need to tweak the behavior a bit to make detections easier to see in other ways - especially cells, where the boundaries may be too faint currently)

Dynamic thickness

Rylern (approved):
I like the new behaviour.

PR #1625, opened 2024-09-09
Add Zarr support for convert-ome command

Rylern:
Edit: a test is failing on Windows. I'm looking at it now

This PR adds support for writing OME Zarr images with the `convert-ome` command line command (which was only creating OME TIFF files until now).

The OME Zarr writer lacked some features to fulfill the requirements of `convert-ome`, so it has been slightly improved.

An example of such features is about considering only a subset of the existing z-stacks (or timepoints) of the input image. I didn't want to code this inside the Zarr writer because it's not specific to Zarr files, so I created a new `ImageServer` called `SlicedImageServer`. It acts like a `CroppedImageServer`, but on the z and t axis.

This PR also adds tests for the `convert-ome` command, the zarr writer, and the `SlicedImageServer`.


qupath-core/src/main/java/qupath/lib/images/servers/SlicedImageServer.java
```diff
@@ -0,0 +1,113 @@
...
+
+    /**
+     * Create an ImageServer that represents a particular set of z-slices and timepoints of another ImageServer.
+     * <p>
+     * Index parameters of this function are not taken into account if their values is out of range (for example,
+     * a zStart of -1, or a tEnd of 10 with an image having only 5 timepoints).
+     *
+     * @param inputServer the input image to slice
+     * @param zStart the inclusive 0-based index of the first slice to consider
+     * @param zEnd the inclusive 0-based index of the last slide to consider
```
petebankhead, line 30: Why not exclusive?
Making the end point exclusive feels more intuitive to me in Java, so that `0...nSlices` would give the entire image, rather than requiring `0...nSlices-1` to avoid being out of range.

qupath-core/src/main/java/qupath/lib/images/servers/SlicedImageServer.java
```diff
@@ -0,0 +1,113 @@
...
+     * @param tStart the inclusive 0-based index of the first timepoint to consider
+     * @param tEnd the inclusive 0-based index of the last timepoint to consider
+     * @throws IllegalArgumentException when a start index is greater than its corresponding end index
+     */
+    public SlicedImageServer(
```
petebankhead, line 35: I haven't been consistent enough on this, but recent `ImageServer` subclasses tend not to make their constructors public, e.g. [`ConcatChannelsImageServer`](https://github.com/qupath/qupath/blob/ed22edd7f501f04cc66cb749b5228d2dc807bd54/qupath-core/src/main/java/qupath/lib/images/servers/ConcatChannelsImageServer.java).

The expectation is that the caller uses [`TransformedServerBuilder`](https://github.com/qupath/qupath/blob/ed22edd7f501f04cc66cb749b5228d2dc807bd54/qupath-core/src/main/java/qupath/lib/images/servers/TransformedServerBuilder.java#L65).

qupath-core/src/main/java/qupath/lib/images/servers/SlicedImageServer.java
```diff
@@ -0,0 +1,113 @@
...
+import java.io.IOException;
+import java.util.Map;
+
+/**
+ * ImageServer that treats a particular set of z-slices and timepoints of another ImageServer
+ * as a full image.
+ */
+public class SlicedImageServer extends TransformingImageServer<BufferedImage> {
+
+    private final ImageServerMetadata metadata;
```
petebankhead, line 16: What do you think of including `zStep` and `tStep` values, with defaults of 1?

"No" is a valid answer here, but I wonder if subsampling may sometimes be desirable. *However* it may not be and you'd need to update the pixel calibration information if this is supported... so maybe best skip it for now.

petebankhead (changes requested):
Approach looks good, I will try running it soon for some conversions. In the meantime I just have small questions/comments.

petebankhead:
The failing test is a mystery - I can confirm it works locally on my Windows (10) computer as well.

Re. the temp directories, should it clean up afterwards? From [the javadocs](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/nio/file/Files.html#createTempDirectory(java.nio.file.Path,java.lang.String,java.nio.file.attribute.FileAttribute...))

> As with the `createTempFile` methods, this method is only part of a temporary-file facility. A [shutdown-hook](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/Runtime.html#addShutdownHook(java.lang.Thread)), or the [File.deleteOnExit()](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/io/File.html#deleteOnExit()) mechanism may be used to delete the directory automatically.

qupath-core/src/main/java/qupath/lib/images/servers/SlicedImageServer.java
Rylern: Yes. I can include that in a future PR

petebankhead:
Thanks, I applied this to `CMU-1.svs` and the speed when writing in parallel is impressive.

* Could/should we make parallel the default for writing (`-p`)?
* Is it possible to incorporate any other metadata, most importantly pixel size?
* Could/should pyramid levels be retained by default?
* Is it possible to incorporate compression into the ome.zarr export?
   The command line suggests the compression option is only relevant for TIFF.

petebankhead (approved):

qupath-core/src/main/java/qupath/lib/images/servers/SlicedImageServer.java
Rylern: Or instead of the `start`, `step`, and `end`, couldn't we simply give a list of indices to consider?

petebankhead: Sorry, I missed this question originally...

The trouble with giving a list of indices:
1. it's harder to write a reusable command that gives a list of indices whenever the number of z-slices can change
2. often, we will have a z-spacing value in µm; if we have a step size, we can scale the z-spacing by the step size to get the z-spacing value to store in the output metadata. But if someone selects slices 1, 2, 4, 10 then the resulting z-stack will have different spacing between slices. This isn't generally expected, and means that a single z-spacing value can no longer be used.

PR #1630, opened 2024-09-11
Update ConvertCommand

petebankhead:
* Change range is valid.
* Tell user defaults for main options where that is relevant.

### DIFF
--- a/qupath-extension-bioformats/src/main/java/qupath/lib/images/writers/ome/ConvertCommand.java
+++ b/qupath-extension-bioformats/src/main/java/qupath/lib/images/writers/ome/ConvertCommand.java
@@ -106,10 +106,10 @@ public class ConvertCommand implements Runnable, Subcommand {
 	@Option(names = {"--tile-size"}, defaultValue = "-1", description = "Set the tile size (of equal height and width).")
 	private int tileSize;
 	
-	@Option(names = {"--tile-width"}, defaultValue = "512", description = "Set the tile width.")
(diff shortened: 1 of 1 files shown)


qupath-extension-bioformats/src/main/java/qupath/lib/images/writers/ome/ConvertCommand.java
```diff
@@ -341,6 +353,10 @@ private static Range getRange(String rangeText, int maxRange, String rangeLabel)
 		}
 	}
 
+	static boolean isValidRange(Range range, int maxRange) {
```
Rylern, line 356: Why not `private`?

Rylern (approved):

qupath-extension-bioformats/src/main/java/qupath/lib/images/writers/ome/ConvertCommand.java
petebankhead: It depends if any tests will be added - if so, I think it has to be package private?

petebankhead: If not, then yes to private.

Rylern: This will be tested via the convert command, not via the `isValidRange()` function. `isValidRange()` is an implementation detail, while tests are only testing the public behavior of a class.

Also, tests should never make the main code less robust (by leaking implementation details here).

PR #1638, opened 2024-09-13
Support remote ome-zarr

petebankhead:
First steps to handle remote OME-Zarr.

See https://forum.image.sc/t/qupath-0-6-0rc1-unable-to-read-remote-ome-zarr-by-url/101605 for details.

This currently 'works' for https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0073A/9798462.zarr but is too slow to be very usable.


petebankhead:
Merging for now, but still need to improve performance.

PR #1641, opened 2024-09-16
Zarr writer and ConvertCommand improvements

Rylern:
* Cleaned up images created by `TestConvertCommand`.
* In `ConvertCommand`, deleted existing output directories whenever `—overwrite` is specified.
* Created `image.zarr/OME/METADATA.ome.xml` in Zarr writer. This allows to save the entire QuPath metadata of images.
* Added tests for the Zarr writer.


petebankhead:
This looks really good, thanks.

I've added one small commit to handle ARGB conversion.

Initially, when I converted `CMU-1.svs` to `ome.zarr` I was unable to read it again. This was because `BioFormatsImageServer` recognized that the image should be RGB, and therefore created an `ARGB` `ColorModel` (compatible with a packed int representation). But `OMEPixelParser` created an incompatible `BufferedImage` because it received byte arrays, and didn't previously convert to packed int.

The reason that QuPath converts *all* (A)RGB images to a packed int representation is somewhat historical, and because I believe this is more optimized in Java's rendering pipeline. It's possible that this belief is out-of-date though, and we should profile one day to check: perhaps ARGB conversion is no longer required.

PR #1653, opened 2024-09-24
Set BioFormats local file only pref false by default

Rylern:


petebankhead:
We need to explain why, so I've added a line in the changelog that links to this.

For more context, see https://forum.image.sc/t/qupath-0-6-0rc1-unable-to-read-remote-ome-zarr-by-url/101605/14

PR #1654, opened 2024-09-24
Ensure no platform thread exception in pathobjectgridview

alanocallaghan:
Also does some linting


alanocallaghan:
Converted to a draft because I think this may still cause application thread exceptions sometimes - sortAndFilter is the real culprit, I think...

petebankhead:
Is it because we're handling image data property or hierarchy change events on the thread that posts them?
(I am guessing)

PR #1655, opened 2024-09-25
Grid view padding and threading

### LINKED ISSUE #1657: TMA core grid view sort by name doesn't handle numbers as expected
### Describe the bug
Somewhere between a bug report and enhancement request...

The ability to view classifications and names in the grid view is very handy https://github.com/qupath/qupath/pull/1588

The trouble is that, for TMA cores, sorting by name gives `A-1`, `A-10`, `A-11`, `A-2`... which is kinda right, but not what the user likely wants.

### To Reproduce
1. Create a TMA grid with 10 or more columns
2. Run _Measure → Grid views... → Show TMA core grid view_

### Expected behavior
Cores are ordered using numbers in a 'smarter' way.

 - OS: All
 - QuPath Version: 0.6.0-SNAPSHOT

### Additional context
`GeneralTools` has a method that is sort of relevant [here](https://github.com/qupath/qupath/blob/e15df73ff60ef167eac82237586caf1f2e81ceab/qupath-core/src/main/java/qupath/lib/common/GeneralTools.java#L910).

It feels like something an existing dependency library might have, but I haven't found it yet.

> *Extra thing: when setting the classification of a TMA core *after* the grid is shown, the color doesn't update in the thumbnail image.*

alanocallaghan:
#1654 
Resolve #1657

### DIFF
--- a/qupath-gui-fx/src/main/java/qupath/lib/gui/commands/PathObjectGridView.java
+++ b/qupath-gui-fx/src/main/java/qupath/lib/gui/commands/PathObjectGridView.java
@@ -505,7 +505,7 @@ public class PathObjectGridView implements ChangeListener<ImageData<BufferedImag
 						// pathclass is present and selected, or missing and we're showing unclassifier
 						&& (selectedClasses.contains(p.getPathClass()) || (p.getPathClass() == null && selectedClasses.contains(PathClass.NULL_CLASS)))
 		);
-		grid.getItems().setAll(filteredList);
+		Platform.runLater(() -> grid.getItems().setAll(filteredList));
 	}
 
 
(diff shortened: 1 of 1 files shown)


alanocallaghan:
Sorry, it took me far too long to understand the padding code, honestly not sure why as it's not particularly complex

qupath-gui-fx/src/main/java/qupath/lib/gui/commands/PathObjectGridView.java
```diff
@@ -505,14 +496,12 @@ private void sortAndFilter() {
 						// pathclass is present and selected, or missing and we're showing unclassifier
 						&& (selectedClasses.contains(p.getPathClass()) || (p.getPathClass() == null && selectedClasses.contains(PathClass.NULL_CLASS)))
 		);
-		grid.getItems().setAll(filteredList);
+		Platform.runLater(() -> grid.getItems().setAll(filteredList));
```
petebankhead, line 499: Show this always `runLater`, even if `Platform.isFxApplicationThread()` returns `true`?

qupath-gui-fx/src/main/java/qupath/lib/gui/commands/PathObjectGridView.java
```diff
@@ -584,24 +571,22 @@ private void updateChildren() {
 			}
 			List<Node> images = new ArrayList<>();
 			for (PathObject pathObject : list) {
-				Label viewNode = nodeMap.get(pathObject);
-				if (viewNode == null) {
-					
+				Label viewNode = nodeMap.computeIfAbsent(pathObject, po -> {
```
petebankhead, line 574: Admittedly the starting code wasn't great and the switch to `computeIfAbsent` makes sense, but could this be simplified?

It's quite hard to read the long lambda, and then there is another lambda nested inside to handle the mouse clicks. Is there any reason this can't be split out into its own method?

qupath-gui-fx/src/main/java/qupath/lib/gui/commands/PathObjectGridView.java
alanocallaghan: I figured it's unimportant as long as it is run later? I did add a check though if not

Issue #1660, opened 2024-09-30
No way to hide/minimize script editor without closing scripts


petebankhead:
### Describe the bug
After https://github.com/qupath/qupath/pull/1524 (I think) it's no longer possible to close the script editor without closing scripts.

This was flagged as a potential annoyance at https://github.com/qupath/qupath/pull/1524#issuecomment-2133544271 and then reported as an actual annoyance at https://forum.image.sc/t/feedback-on-qupath-v0-6-0-rc2/103187/3?u=petebankhead

### To Reproduce
1. Open a script editor in v0.6.0-rc2
2. Type something, anything
3. Try to close the script editor window
4. See it isn't possible to close the window without saving the script
5. Reopen the script editor - the script is gone

### Expected behavior
Not sure. 

v0.5.x and before permitted the script editor to be closed without requesting the user to save anything.
The trouble with that was that people could easily lose edited scripts without being prompted to save changes.

Two possible options:
1. Provide an option to hide the script editor without closing scripts - _but still prompt to save changes when QuPath is being closed_
2. Keep current behaviour, but retain the last script when reopening the script editor window

The first option is perhaps the most useful, since the second would only permit retaining one script at a time - and multiple scripts may be open.

### Desktop (please complete the following information):
 - OS: All
 - QuPath Version: v0.6.0-rc2

PR #1663, opened 2024-09-30
Hide script editor on close requests

petebankhead:
Potential fix for https://github.com/qupath/qupath/issues/1660

This hides the script editor when either pressing the corner button to close the window, or calling 'File > Close editor' - but does *not* close the scripts.

However, when QuPath is being quit entirely, if there are unsaved scripts then the script editor should appear again - with prompts to save.

### DIFF
--- a/qupath-gui-fx/src/main/java/qupath/lib/gui/scripting/DefaultScriptEditor.java
+++ b/qupath-gui-fx/src/main/java/qupath/lib/gui/scripting/DefaultScriptEditor.java
@@ -539,6 +539,9 @@ public class DefaultScriptEditor implements ScriptEditor {
 			if (tab == null) {
 				break;
 			}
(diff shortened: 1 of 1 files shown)


qupath-gui-fx/src/main/java/qupath/lib/gui/scripting/DefaultScriptEditor.java
```diff
@@ -1998,7 +2002,9 @@ Action createFindAction(final String name) {
 	
 	Action createExitAction(final String name) {
 		Action action = new Action(name, e -> {
-			requestClose();
+//			requestClose();
```
Rylern, line 2005: Can this line be removed?

Rylern (approved):

alanocallaghan (approved):
Looks good, a bit friendlier than the previous implementation

PR #1669, opened 2024-10-03
Handle color changes for RGB fluorescence images

petebankhead:
This revises the approach in https://github.com/qupath/qupath/pull/1659 in the latest attempt to fix https://github.com/qupath/qupath-extension-omero/issues/25

## Purpose

With this PR, RGB images that are set to have the type 'Fluorescence' now behave much more similarly to non-RGB images, and it is possible to set both channel names and colors.

However all the other image type options remain, e.g. Brightfield H&E. If one of these is selected, then all channel name/color customisations are ignored - and the image behaves as a regular RGB.

## Risks & assumptions
* Querying channel names & colors for an RGB `ImageServer` is no longer guaranteed to return `Red`, `Green` or `Blue`
* Custom channel names for an RGB `ImageServer` are also not guaranteed to be represented in the viewer; rather, this only occurs if the image type is `Fluorescence`
* This does *not* provide a way to convert a BGR image to RGB from the perspective of any analysis. `ImageServer.readRegion()` requests should still return the same packed int (A)RGB image as before - the main difference is how things are rendered in the viewer.
* A pixel classifier that is trained on an image with custom RGB channel names will only work on an image with the same custom channel names*

> *-Caveat: this can give unexpected results if a pixel classifier is applied to two RGB images exist in a project, and one has different channel names. Then, if the classifier is applied successfully to one image, it can *appear* to work on the second image because cached tiles are used - so the mismatched channel names are not identified.

## Example

Example using the ImageJ 'Fluorescent Cells' image, flattened to RGB.

### Image type set to 'Fluorescence', channels adjusted
RGB-fluorescence

### Same image with type set to 'Other'
RGB-Other

### DIFF
--- a/qupath-core-processing/src/main/java/qupath/lib/scripting/QP.java
+++ b/qupath-core-processing/src/main/java/qupath/lib/scripting/QP.java
@@ -1399,9 +1399,6 @@ public class QP {
 	/**
 	 * Set the channels for the specified ImageData.
 	 * Note that number of channels provided must match the number of channels of the current image.
-	 * <p>
-	 * Also, currently it is not possible to set channel colors for RGB images - attempting to do so
-	 * will throw an IllegalArgumentException.
 	 * 
 	 * @param imageData 
 	 * @param channels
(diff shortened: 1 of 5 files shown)


alanocallaghan:
This works great! Should be a very helpful accessibility feature for colorblind folks :)

finglis (approved):

PR #1671, opened 2024-10-03
Support system light/dark mode

petebankhead:
Also use the system style by default.

This seems to be working for me on macOS - @alanocallaghan and @finglis could you let me know if it works ok on Linux & Windows too please?

style_light-1

style_dark-1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ Some things may be added, some things may be removed, and some things may look d
 ### Enhancements
 (These are not yet ordered by interestingness)
 
+* Support system light/dark color themes
 * Improved display of annotation names (https://github.com/qupath/qupath/pull/1532)
 * Support log10 counts with histograms (no longer use natural log) (https://github.com/qupath/qupath/pull/1540)
   * Log counts also now available in measurement tables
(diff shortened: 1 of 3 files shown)


alanocallaghan:
Works great on my Linux box, an Ubuntu 22.04 derivative with GNOME desktop environment:

```
Description:	Pop!_OS 22.04 LTS
Release:	22.04
Codename:	jammy
```

Would probably be worthwhile checking a KDE plasma version, will look into that quickly

finglis:
Windows is responding as expected

petebankhead:
Ah, that's a bit of a shame... but thanks for testing - there might be some other [platform preferences](https://openjfx.io/javadoc/23/javafx.graphics/javafx/application/Platform.Preferences.html) that help, although I'd be tempted to just leave is a bug that may magically get fixed by dependency updates. We still have the option to choose light or dark explicitly, so as long as that still works no one should be stuck.

alanocallaghan:
In fairness, I tested on Plasma and it seems like `Platform.getPreferences().getColorScheme()` is indeed returning light when dark and dark when light, so there's not a whole lot we can do to handle that

PR #1679, opened 2024-10-15
Update ImageJ

petebankhead:
Seeing if this works... some dependencies in between caused failing tests when run headlessly.


petebankhead:
See https://wsr.imagej.net/notes.html
Regression that caused the exceptions was introduced in v1.54i and fixed in 1.54k

> Thanks to Felix Rudolphi and Curtis Rueden, fixed a 1.54i regression that caused FileOpener to throw HeadlessExceptions in headless environments.

Issue #1652, opened 2024-09-20
Javadoc jars are missing from QuPath installations


petebankhead:
### Describe the bug
Because of some [JavaFX WebView awkwardness](https://github.com/qupath/qupath/issues/1497), we stop merging all the QuPath javadocs into a single jar, in favor of trying to bundle all the relevant jars with the app.

Unfortunately, for v0.6.0-rc1 and v0.6.0-rc2 lots of jars are missing.

### To Reproduce
1. Go to your QuPath installation
2. Look inside `app/docs`
3. Ask yourself "Is that all there is?"
4. Answer yourself "Yes, but there really should be others"

### Expected behavior
Javadoc jars for all the main QuPath classes are included.

### Desktop (please complete the following information):
 - OS: All (presumably; confirmed on macOS and Windows)
 - QuPath Version: v0.6.0-rc1, v0.6.0-rc2

alanocallaghan:
With
```
./gradlew clean jpackage
```

I end up with what I guess is the full list:

```
ij-1.54k-javadoc.jar
    jts-core-1.20.0-javadoc.jar
    qupath-app-0.6.0-SNAPSHOT-javadoc.jar
    qupath-bioimageio-spec-0.1.0-javadoc.jar
    qupath-core-0.6.0-SNAPSHOT-javadoc.jar
    qupath-core-processing-0.6.0-SNAPSHOT-javadoc.jar
    qupath-extension-bioformats-0.6.0-SNAPSHOT-javadoc.jar
    qupath-extension-openslide-0.6.0-SNAPSHOT-javadoc.jar
    qupath-extension-processing-0.6.0-SNAPSHOT-javadoc.jar
    qupath-extension-script-editor-0.6.0-SNAPSHOT-javadoc.jar
    qupath-extension-svg-0.6.0-SNAPSHOT-javadoc.jar
    qupath-fxtras-0.1.5-javadoc.jar
    qupath-gui-fx-0.6.0-SNAPSHOT-javadoc.jar
```

If I add `-P package=installer` then I do wonder "Is this the life?" (by which I mean, I only get the shorter list).

petebankhead:
Ah... that may explain why I was convinced it was fixed - until v0.6.0-rc3 proved it wasn't :(

Seems really bizarre. Strong chance I'm doing truly awful things [here](https://github.com/qupath/qupath/blob/main/qupath-app/build.gradle) but I couldn't figure it out.

alanocallaghan:
[...]

PR #1694, opened 2024-11-05
Add assemblejavadoc to jpackage action

alanocallaghan:
For some reason, when running `gradlew jpackage` you get the full set of qupath javadocs, but with `gradlew jpackage -P package=installer`, you get just a small subset. However you can bodge around this by explicitly asking for all the javadocs as part of the same call; it'll still copy the directory.

Temporarily address #1652; won't consider it closed until jpackage contains all javadoc jars by default.


alanocallaghan:
Tested on my fork and seems to work, but definitely one to watch in future releases. Will merge for now

PR #1696, opened 2024-11-08
Use Kotlin DSL with Gradle

petebankhead:
This PR proposes to switch from using Groovy to the Kotlin DSL for Gradle.

Reasons include:
1. The current build scripts are pretty unreadable and unmaintainable
2. Since April last year, [Kotlin has been the default for Gradle](https://blog.gradle.org/kotlin-dsl-is-now-the-default-for-new-gradle-builds)
3. Kotlin is stricter, so hopefully that means the scripts will be a lot more comprehensible... and the IDE can help more
4. Kotlin seems very popular & powerful, so worth knowing about

It's possible to mix between Kotlin and Groovy, even within the same project, so it's not essential that extensions or other projects follow this.

Most of the build has been kotlinized here, but the plugins haven't been yet.

---

## Explanation of the changes

The main change is the move from Groovy to Kotlin, but this PR also includes extensive refactoring.

Previously, `qupath-app` contained most of the logic to build the QuPath packages - and the `qupath-app.jar` classpath included everything.

Now, `qupath-app` builds only the launcher jar, and the classpath is not specified within the jar (although this might change in the future).

Instead, packages are build through `build.gradle.kts` in the root project - with most of the configuration of custom plugins handled within [convention plugins](https://docs.gradle.org/current/samples/sample_convention_plugins.html#compiling_convention_plugins) (implemented as pre-compiled script plugins... I think).

Therefore if you want to modify the packaging, you probably want to go to `qupath.jpackage-conventions` (although `build.gradle.kts` is still a reasonable starting point).

Alternatively, if it's the license reporting then try `qupath.license-conventions`.

Or if it's the bundled JavaFX modules then go to `qupath.javafx-conventions`.

Additionally, `settings.gradle.kts` includes more logic related to command line properties - partly to help this be more findable, and help make the defaults easier to identify.

For example
```kotlin
// By default, create a per-user installer on Windows
gradle.extra["qupath.package.per-user"] = providers.gradleProperty("per-user-install").getOrElse("true")
```
indicates that calling `gradlew package -Pper-user-install=false` would make it possible to generate a system-wide installer, related to https://github.com/qupath/qupath/issues/1703


Rylern (approved):
The usual Gradle tasks (e.g. `jpackage`) work correctly on Linux

alanocallaghan:
I can't say I understand the build system entirely but this does seem a big improvement in readability and hopefully maintainability moving forward

PR #1708, opened 2024-11-29
Parameterise version in gradle and GHA

### LINKED ISSUE #1707: GitHub Action (probably) needs updated for new build scripts
Since the build scripts were updated to use Kotlin https://github.com/qupath/qupath/pull/1696 I suspect at least [this GitHub action](https://github.com/qupath/qupath/blob/main/.github/workflows/github-release.yml#L26) will need to be updated to request the QuPath version some other way.

@alanocallaghan could/should we store the version somewhere else entirely?

The Kotlin scripts might change again before release...

alanocallaghan:
This PR adds a required field for the JPackage and github-release actions, so when drafting a new release, you're required to enter the target version. Also adds an optional way to set the version in the build script via the `qupathVersion` property.

Resolve #1707

### DIFF
--- a/.github/workflows/github-release.yml
+++ b/.github/workflows/github-release.yml
@@ -1,7 +1,13 @@
 
 name: Draft a GitHub release
 
-on: workflow_dispatch
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: The QuPath version that will be built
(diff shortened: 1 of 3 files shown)


alanocallaghan:
#1710

PR #1710, opened 2024-11-30
Simplify config for version & extensions

petebankhead:
Relates to
- https://github.com/qupath/qupath/issues/1707
- https://github.com/qupath/qupath/pull/1709

This attempts to simplify configuration firstly by introducing a `VERSION` text file containing the current version.

This is read by `settings.gradle.kts`, but could potentially also be used in GitHub actions.

Additionally, support is added for an optional `include-extra.txt` file to make it easier to develop and debug extensions.
The syntax is

```
[includeBuild]
../qupath-extension-template
../qupath-extension-instanseg

[dependencies]
# Comments are supported, if you like
io.github.qupath:qupath-extension-template
io.github.qupath:qupath-extension-instanseg
```

If this file is present in the root directory, it will be parsed within `settings.gradle.kts`.

The specified builds will be included, and the dependencies passed on to the root project's `build.gradle.kts` for inclusion there.

This makes it easier to develop arbitrary lists of extensions - and potentially explore the effect of introducing other dependencies - but just creating and editing this one file.


alanocallaghan (approved):
include-extra looks good. I *think* I prefer the idea of version being an optional property, but looking at other projects, it seems like it's usually defined in code, so this solution is more conventional.

finglis (approved):

Rylern (approved):
I like this approach. The README of https://github.com/qupath/qupath-extension-template should be updated with this new way of proceeding

PR #1713, opened 2024-12-05
Remove not clear

alanocallaghan:
Change the QP scripting API to use "remove" for all static methods that remove objects from the hierarchy, or remove measurements from objects. Previously "clear" was used for some methods that reset the selection in the current hierarchy, some methods that remove objects, and some methods that remove measurements from objects.

Deprecate old "clear" methods and add equivalent API using more specific terms. See https://github.com/qupath/planner/issues/86

### DIFF
--- a/.github/workflows/github-release.yml
+++ b/.github/workflows/github-release.yml
@@ -20,13 +20,10 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
-      - name: Get qupath version from gradle
+      - name: Set QuPath version
         shell: bash
         run: |
-          # grep reads the version line, sed strips everything before =
-          # GITHUB_ENV means it's stored in env.QUPATH_VERSION persistently
-          # not just for the current step...                
-          echo "QUPATH_VERSION=$(grep 'qupathVersion = \".*\"' settings.gradle  | sed -e 's/.*= "\(.*\)"/\1/')" >> $GITHUB_ENV
+          echo "QUPATH_VERSION=$(cat VERSION)" >> $GITHUB_ENV
 
       - uses: actions/download-artifact@v4
         with:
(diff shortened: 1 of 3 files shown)


petebankhead:
I also dislike the use of 'clear', although there are other places where this will matter... and inconsistency remains, since 'remove' doesn't match with our command names, e.g. *Objects → Delete... → Delete all annotations*

See also where this appears in code, and adds scripts to the workflow
https://github.com/qupath/qupath/blob/e04f0d428ed40f3036c536d4c50c5833dbbac5f5/qupath-gui-fx/src/main/java/qupath/lib/gui/commands/Commands.java#L1246

While less important, we also use `CLEAR` in field names [here](https://github.com/qupath/qupath/blob/e04f0d428ed40f3036c536d4c50c5833dbbac5f5/qupath-gui-fx/src/main/java/qupath/lib/gui/actions/menus/ObjectsMenuActions.java#L103).

Because this will affect script compatibility across versions, I think we need to err on the side of all-or-nothing.

alanocallaghan:
I think "remove" is the best balance between clarity and correctness, because objects removed from the hierarchy aren't necessarily deleted, excluding normal garbage collector rules.

I did assume there'd be other uses yeah, and agree we should be consistent in the UI and script commands.

petebankhead:
Hmmm, the garbage collection point seems quite subtle... but now you mention it, 'remove' feels more correct to me in code, while 'delete' feels more comprehensible to me in the UI. I also agree that we should be consistent in UI and script. My mental faculties seem to have reached an impasse.

qupath-core-processing/src/main/java/qupath/lib/plugins/objects/TileClassificationsToAnnotationsPlugin.java
```diff
@@ -97,7 +97,7 @@ public Collection<Class<? extends PathObject>> getSupportedParentObjectClasses()
 	@Override
 	protected void addRunnableTasks(final ImageData<T> imageData, final PathObject parentObject, List<Runnable> tasks) {
 		ParameterList params = getParameterList(imageData);
-		if (params.getBooleanParameterValue("clearAnnotations")) {
+		if (params.getBooleanParameterValue("clearAnnotations") || params.getBooleanParameterValue("removeAnnotations")) {
```
alanocallaghan, line 100: Was hoping this would be sufficient to avoid things breaking between versions, but I think I'm wrong

PR #1715, opened 2024-12-05
add isSpatiallyTransformed to ImageServerMetadata

carlocastoldi:
When working with annotations coming from different programs, the user might need flip, crop, rotate or transform them according to QuPath's transformation before importing them to a QuPath project.
Thanks to https://github.com/qupath/qupath/pull/1489 users now can do this extremely fast, without loading or reading the actual `ImageServer`, _on the assumption that no spatial transformation was applied.

With this pull request I intend to add an additional metadata to image servers in order to be able to query the server if any spatial transformation was actually applied. If not, loading the server can be skipped.

### DIFF
--- a/qupath-core/src/main/java/qupath/lib/images/servers/AffineTransformImageServer.java
+++ b/qupath-core/src/main/java/qupath/lib/images/servers/AffineTransformImageServer.java
@@ -46,7 +46,7 @@ import qupath.lib.regions.RegionRequest;
  * @author Pete Bankhead
  *
  */
-public class AffineTransformImageServer extends TransformingImageServer<BufferedImage> {
+public class AffineTransformImageServer extends SpatiallyTransformingImageServer<BufferedImage> {
 	
 	private static Logger logger = LoggerFactory.getLogger(AffineTransformImageServer.class);
 	
@@ -83,23 +83,24 @@ public class AffineTransformImageServer extends TransformingImageServer<Buffered
 		var levelBuilder = new ImageServerMetadata.ImageResolutionLevel.Builder(region.getWidth(), region.getHeight());
 		boolean fullServer = server.getWidth() == region.getWidth() && server.getHeight() == region.getHeight();
 		int i = 0;
+		ImageServerMetadata embeddedMetadata = super.getMetadata();
 		do {
-			var originalLevel = server.getMetadata().getLevel(i);
+			var originalLevel = embeddedMetadata.getLevel(i);
 			if (fullServer)
 				levelBuilder.addLevel(originalLevel);
 			else
 				levelBuilder.addLevelByDownsample(originalLevel.getDownsample());
 			i++;
 		} while (i < server.nResolutions() && 
-				region.getWidth() >= server.getMetadata().getPreferredTileWidth() && 
-				region.getHeight() >= server.getMetadata().getPreferredTileHeight());
+				region.getWidth() >= embeddedMetadata.getPreferredTileWidth() &&
+				region.getHeight() >= embeddedMetadata.getPreferredTileHeight());
 		
 		// TODO: Apply AffineTransform to pixel sizes! Perhaps create a Shape or point and transform that?
-		var builder = new ImageServerMetadata.Builder(server.getMetadata())
+		var builder = new ImageServerMetadata.Builder(embeddedMetadata)
 //				.path(server.getPath() + ": Affine " + transform.toString())
 				.width(region.getWidth())
 				.height(region.getHeight())
-				.name(String.format("%s (%s)", server.getMetadata().getName(), transform.toString()))
+				.name(String.format("%s (%s)", embeddedMetadata.getName(), transform.toString()))
 				.levels(levelBuilder.build());
 		
 		// TODO: Handle pixel sizes in units other than microns
@@ -109,7 +110,7 @@ public class AffineTransformImageServer extends TransformingImageServer<Buffered
 				logger.debug("Pixel calibration updated to {}", calUpdated);
 			builder.pixelSizeMicrons(calUpdated.getPixelWidthMicrons(), calUpdated.getPixelHeightMicrons());
 		}
-				
+
 		metadata = builder.build();
 	}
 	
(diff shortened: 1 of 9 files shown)


petebankhead:
Thanks, it is very unlikely that we can incorporate any unplanned major change for v0.6.0, and it will probably take a while before we can evaluate this properly.

In the meantime, can you describe in more detail exactly what problem this PR should solve?

carlocastoldi:
Thanks, I figured it would have probably been too late for v0.6.0 but it was worth a trying as i feel it's heavily coupled with the lazy loading of  `ImageData`.
Would it be eligible for 0.6.1?

>In the meantime, can you describe in more detail exactly what problem this PR should solve?

It is specifically meant for ABBA, a software for aligning multiple brain sections to a brain atlas. Its final step exports the brain region annotations to an associated QuPath project through an extension. Both the ABBA project and the QuPath project work on the same image files, but since QuPath may change the orientation of the images, ABBA requires to adjust the orientation of brain region annotations accordingly.
This step of aligning, exporting to QuPath and improving the alignment to a brain atlas is often iterative and works on lots of huge histological images (~100GB total). Accessing the imageservers should be avoided whenever it's possible, or the process of re-alignment could take tens of minutes just to import a registration to the atlas which may very well need changes.

Apart from this application, I could also see this being usefull when working on the same image files with different programs—i.e. QuPaht and another one (e.g. OMERO)—where one draws the annotations and the other does some other kind of analysis.

petebankhead:
Thanks, I'm struggling to understand - perhaps because I haven't used ABBA, and I don't remember where *"the assumption that no spatial transformation was applied"* originates.

Is there any minimal example to show the problem that this should solve?

And does it relate to any *bug* with the lazy-loading, or is it rather than some images can't be lazy-loaded when it would be desirable that they could?

A PR like this that introduces new classes wouldn't really belong in a v0.0.x release. If I could understand the problem better, maybe there's a simpler way to get the desired outcome without a major API change.

carlocastoldi:
> And does it relate to any bug with the lazy-loading, or is it rather than some images can't be lazy-loaded when it would be desirable that they could?

the latter, yes.

> Thanks, I'm struggling to understand - perhaps because I haven't used ABBA, and I don't remember where "the assumption that no spatial transformation was applied" originates.

Thank you for your patience—let me try to clarify further.

Since lazy loading was introduced, annotations and detections can be handled without actually reading the image pixels. This is particularly useful for tasks like _quickly_ exporting results or quickly creating a classifier from an entire project as training-set (benefits introduced and possible since #1489). However, lazy loading could also simplify importing annotations from outside QuPath.
If the annotations being imported were drawn over the same images used in a QuPath project, they should theoretically be importable without reading the image pixels. This is because the coordinate systems are either identical or _can be derived with minimal effort_.
However, the coordinate system of annotations can change if the corresponding image in QuPath was flipped, cropped, or otherwise spatially transformed. In QuPath, such transformations are encoded as a new `ImageServer`, typically wrapping the original image server (e.g., as a `TransformedImageServer`). The issue is that there’s currently no way to determine whether such a transformation has occurred without reading the `ImageServer`.

To address this, my proposal is to add a new metadata field that explicitly indicates whether the ImageServer applies any spatial transformation to the original image. This metadata would eliminate the need to read the image server when no transformation is applied.

> Is there any minimal example to show the problem that this should solve?

Unfortunately I can't think of any simple one at the moment. I should spend some time to build one. Perhaps you could try to look [at the small commit](https://github.com/carlocastoldi/qupath-extension-abba/commit/2dda12329cac6f030db6f465d1eb1741514a96dd) that i did in order to adapt ABBA's extension to the API changes? What i'd like you to notice is that the `imageData.getImageServer()` is not called by default.

If it wasn't clear from the commit, please let me know. I'll try my best to build a minimal example asap.

carlocastoldi:
P.S. apart from the introduction of the `isSpatiallyTransformed` metadata, I feel like this PR also refactors the code of some `TransformingImageServers` nicely, by separating the server creation, the metadata definition and eliminating the need of an internal representation of the metadata for each `TransformingImageServer`.

Might be an interesting non-API-change nonetheless.

petebankhead:
Thanks for the explanation. If I understand correctly, there isn't currently a bug - and the benefit is really for the ABBA extension - and mostly for `RotatedImageServer` at this time.

While a bit messy, I think a workaround requesting the `ServerBuilder` would enable you to get most of the benefits without this PR - and perhaps even avoid having to build the `RotatedImageServer` too. Specifically, here's a Groovy script that checks whether an image would be built to give a `RotatedImageServer`, and also queries which rotation would be used.

```groovy
var builder = getCurrentImageData().getServerBuilder()

println builder.getClass().getSimpleName()

if (builder.getClass().getSimpleName().startsWith("Rotated")) {
    println builder.rotation
}
```

The API isn't public so you'd need to use reflection to access the `rotation` field, making it all rather brittle - and it isn't straightforward to handle other kinds of transform (or perhaps a `RotatedImageServer` wrapped in some other server). But it isn't *that* brittle, because it's awkward for us to change `ServerBuilder` implementations given how they are serialised within projects - so it shouldn't happen often.

We're really stretched trying to get v0.6.0 ready as soon as possible, and I won't have the bandwidth to review this PR thoroughly and think through it's implications in the near future - especially when it doesn't relate to a bug in the current software, or offer any obvious benefit for the majority of users. So even if there could be other benefits in the PR, I think a workaround is a better option if you want to incorporate the change in any time soon.

carlocastoldi:
Thanks, that's actually a good workaround for my use case!
Following your imput I should have also adapted the code to handle the recursivity of the `TransformingImageServer`!

I would still keep this PR up for when you will have time to see it. I think the small refactoring of the `TransformatingImageServer` makes sense nonetheless of its initial reasoning

petebankhead:
I'm closing the PR because there is a workaround, and we don't have the bandwidth to fully review and test this to be confident that it doesn't introduce any compatibility or maintenance troubles. I don't like how QuPath currently serialises the metadata for each server, but I'm nervous about changing this without a lot of exploration to see the effects (especially when opening the same project across different QuPath versions).

We're starting to think about a big rewrite of lots of aspects of QuPath. We'll certainly keep the insights here in mind, to try to make the new design more flexible - thanks again.

PR #1716, opened 2024-12-05
Extension manager integration

Rylern:
This PR replaces the current extension manager of QuPath by https://github.com/qupath/extension-manager.

To run this PR, you need to install https://github.com/qupath/extension-manager alongside `qupath` and add a `include-extra.txt` in the `qupath` directory with:

```
[includeBuild]
../extension-manager

[dependencies]
io.github.qupath:extensionmanager
```
This will be needed until https://github.com/qupath/extension-manager is on maven.

This PR is a draft because:

- [x] Tests need to be done.
- [x] Strings must be externalized.
- [x] https://github.com/qupath/extension-manager should be on maven.

A quick recap about https://github.com/qupath/extension-manager:
* It provides a `qupath.ext.extensionmanager.core.ExtensionIndexManager` class that manages indexes and extensions (without any UI element): it can add/remove indexes, add/remove extensions, list/remove manually installed extensions, automatically load extensions with a custom class loader.
* It provides a `qupath.ext.extensionmanager.gui.ExtensionManager` class that is a window providing controls to manage indexes and extensions. It's like the user interface controlling an `ExtensionIndexManager`.

A few decisions made that might require some discussions:
* A class (`qupath.lib.gui.ExtensionLoader`) was created to install extensions (basically calling `QuPathExtension.installExtension(QuPathGUI)` on detected extensions) and provide UI feedback. This class is not part of https://github.com/qupath/extension-manager because of its deep link with QuPath.
* An `ExtensionIndexManager` class (see above) is a class that manages extensions and indexes. It is not part of the UI. Currently, this class is instantiated by `qupath.lib.gui.QuPathGUI` and `qupath.ScriptCommand`. I don't know if this is the right approach. Since extensions always need to be manageable (whether or not the GUI is used), shouldn't `ExtensionIndexManager` be created in the `qupath.QuPath` class?
* Unlike the previous approach, there is no "`refresh`" function to load all extensions. Instead, they are loaded as soon as they are detected. Thus, there is no property `extensionManager.refreshingExtensions()` anymore. I wasn't able to see if that changed something important (it was only used [here](https://github.com/qupath/qupath/blob/454549e3b6889ca20b44d0128fc73cb8f43948e0/qupath-gui-fx/src/main/java/qupath/lib/gui/QuPathGUI.java#L353)).
* The `qupath.lib.gui.UpdateManager` class was heavily refactored to integrate the new changes. I also defined its UI in a dedicated FXML file (see `qupath.lib.gui.UpdateManagerContainer`).
* **Something quite important**: before, you could access the extension classloader with a [static method](https://github.com/qupath/qupath/blob/454549e3b6889ca20b44d0128fc73cb8f43948e0/qupath-gui-fx/src/main/java/qupath/lib/gui/ExtensionClassLoader.java#L68). Therefore, it was available everywhere. In the new approach, the extension class loader is linked with an `ExtensionIndexManager`, which means you need to have a reference to an instance of `ExtensionIndexManager` to get the extension class loader. This causes a problem in[`QuPathResources`](https://github.com/qupath/qupath/blob/454549e3b6889ca20b44d0128fc73cb8f43948e0/qupath-gui-fx/src/main/java/qupath/lib/gui/localization/QuPathResources.java), because this class contains static methods that uses the extension class loader.
  * Do they really need to access the extension class loader? Cannot they use the default class loader? The strings contained in `qupath-gui-strings` will never be used in the extensions, right?
  * In the PR, I replaced the use of the extension class loader in `QuPathResources` by the default class loader. If that's not possible, then it means that we'll need to make a static function to access the `ExtensionIndexManager`.


alanocallaghan:
1. Can you provide a URL to indexes that work with this version?
2. Should we include an index of QuPath extensions by default?
3. The extension manager windows opens near the top left of the QuPath window, rather than centrally as other QuPath sub-windows do. Is this intentional?

Rylern:
> Can you provide a URL to index(es) that work with this version?

Yes, https://github.com/Rylern/test-index

> Should we include an index of QuPath extensions by default?

Yes. I'll create one in the qupath GitHub group (which will probably be similar to https://github.com/Rylern/test-index)

> The extension manager windows opens near the top left of the QuPath window, rather than centrally as other QuPath sub-windows do. Is this intentional?

This should be fixed in the last commit.

alanocallaghan:
- Clicking add with empty index field gives an `illegal argument exception (unknown scheme null in )`
- If I add two indexes with the same extension (https://github.com/alanocallaghan/test-index/blob/main/index.json), then I can install eg the DJL extension from one, and then the other, and it then shows both as installed. Which one will be loaded?
- If installing an extension that's installed already (via index or manually) should we ask for confirmation? If I have a local stardist extension and install one from the index, both remain, but which is used?

Rylern:
> Clicking add with empty index field gives an illegal argument exception (unknown scheme null in )

Fixed in last commit (of https://github.com/qupath/extension-manager)

> If I add two indexes with the same extension (https://github.com/alanocallaghan/test-index/blob/main/index.json), then I can install eg the DJL extension from one, and then the other, and it then shows both as installed. Which one will be loaded?

Both will be loaded by the extension class loader, so I think Java [will only pick the first](https://stackoverflow.com/questions/6935705/two-classes-with-same-name-in-classpath)? 

> If installing an extension that's installed already (via index or manually) should we ask for confirmation? If I have a local stardist extension and install one from the index, both remain, but which is used?

Yes, I'll see if I can implement that next week

alanocallaghan:
> Both will be loaded by the extension class loader, so I think Java [will only pick the first](https://stackoverflow.com/questions/6935705/two-classes-with-same-name-in-classpath)?

Ideally would avoid/warn about this - in the simplest case we could just warn if there's any duplicated jar names?

Rylern:
> Ideally would avoid/warn about this - in the simplest case we could just warn if there's any duplicated jar names?

Added in last commit. However, the warning won't appear if two different versions of the same extension are installed, because they will have a different jar name.

> If installing an extension that's installed already (via index or manually) should we ask for confirmation? If I have a local stardist extension and install one from the index, both remain, but which is used?

When installing an extension from an index, if a Jar file with the same name already exists in the extension folder, a confirmation message will be shown to the user. There is the same limitation as above (no message if the two extensions have a different version).

petebankhead:
Overall this is looking great. I spotted a few things / have a few suggestions:

## 1. Exception if no directory set

On a computer that didn't have an user directory set (and where I didn't agree to create one immediately`), I was able to generate the uncaught exception:
```
Caused by: java.lang.NullPointerException: Cannot invoke "java.nio.file.Path.resolve(String)" because the return value of "javafx.beans.property.ReadOnlyObjectProperty.get()" is null
	at qupath.ext.extensionmanager.core.ExtensionFolderManager.getAndCreateCatalogsDirectory(ExtensionFolderManager.java:399)
	at qupath.ext.extensionmanager.core.ExtensionFolderManager.getCatalogDirectoryPath(ExtensionFolderManager.java:220)
	at qupath.ext.extensionmanager.core.ExtensionFolderManager.getExtensionDirectoryPath(ExtensionFolderManager.java:253)
	at qupath.ext.extensionmanager.core.ExtensionCatalogManager.getExtensionDirectory(ExtensionCatalogManager.java:183)
	at qupath.ext.extensionmanager.gui.catalog.ExtensionModificationWindow.<init>(ExtensionModificationWindow.java:126)
	at qupath.ext.extensionmanager.gui.catalog.ExtensionLine.onAddClicked(ExtensionLine.java:134)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	... 63 more
```

## 2. Very long log messages

Some log messages are really long and would ideally be more compact, e.g.

```
06:40:07.77	[pool-6-thread-1]	INFO	qupath.ext.extensionmanager.core.ExtensionCatalogManager	Extension Extension[name=QuPath StarDist extension, description=QuPath extension to run StarDist nucleus detection, author=QuPath, homepage=https://github.com/qupath/qupath-extension-stardist, releases=[Release[name=v0.5.0, mainUrl=https://github.com/qupath/qupath-extension-stardist/releases/download/v0.5.0/qupath-extension-stardist-0.5.0.jar, requiredDependencyUrls=null, optionalDependencyUrls=null, javadocsUrls=null, versionRange=VersionRange[min=v0.5.0, max=null, excludes=null]], Release[name=v0.4.0, mainUrl=https://github.com/qupath/qupath-extension-stardist/releases/download/v0.4.0/qupath-extension-stardist-0.4.0.jar, requiredDependencyUrls=null, optionalDependencyUrls=null, javadocsUrls=null, versionRange=VersionRange[min=v0.4.0, max=null, excludes=null]], Release[name=v0.3.2, mainUrl=https://github.com/qupath/qupath-extension-stardist/releases/download/v0.3.2/qupath-extension-stardist-0.3.2.jar, requiredDependencyUrls=null, optionalDependencyUrls=null, javadocsUrls=null, versionRange=VersionRange[min=v0.3.0, max=null, excludes=null]], Release[name=v0.3.1, mainUrl=https://github.com/qupath/qupath-extension-stardist/releases/download/v0.3.1/qupath-extension-stardist-0.3.1.jar, requiredDependencyUrls=null, optionalDependencyUrls=null, javadocsUrls=null, versionRange=VersionRange[min=v0.3.0, max=null, excludes=null]], Release[name=v0.3.0, mainUrl=https://github.com/qupath/qupath-extension-stardist/releases/download/v0.3.0/qupath-extension-stardist-0.3.0.jar, requiredDependencyUrls=null, optionalDependencyUrls=null, javadocsUrls=null, versionRange=VersionRange[min=v0.3.0, max=null, excludes=null]], Release[name=v0.3.0-rc2, mainUrl=https://github.com/qupath/qupath-extension-stardist/releases/download/v0.3.0-rc2/qupath-extension-stardist-0.3.0-rc2.jar, requiredDependencyUrls=null, optionalDependencyUrls=null, javadocsUrls=null, versionRange=VersionRange[min=v0.3.0-rc2, max=null, excludes=null]]]] of SavedCatalog[name=QuPath catalog, description=Extensions maintained by the QuPath team, uri=https://github.com/qupath/qupath-catalog, rawUri=https://raw.githubusercontent.com/qupath/qupath-catalog/refs/heads/main/catalog.json, deletable=false] installed
```

## 3. Can't tell which files were downloaded

The dialog that shows the files to be downloaded can disappear before it's possible to read the URLs. It does not seem possible for the user to verify that they are happy that the URLs look reasonable, either before or after download.

A quick partial solution would be to simply log them, so the user can check afterwards.

An alternative/additional option would be to show a dialog that presents the URLs to the user *before* download within a `TextArea`, e.g.

Download 3 files?
[list of the files in TextArea]

There is then an opportunity to confirm (in which case the progress bar is shown and download proceeds) or cancel without downloading anything.

## 4. Display of warning message

The warning message can be quite long, and includes '(if it exists)' when it should be possible to check that, so the message only shows when the directory *does* exist.

warn-delete

I suggest the following center-aligned
```
This will delete the directory
/path/to/directory
and all its contents
```
because it separates the lengthy path into its own line(s).

I'd also remove the word 'Warning' because it takes horizontal space, and therefore squeezes the path further. Instead you could style it using CSS to use a relevant color (e.g. the ones used in QuPath are [here](https://github.com/qupath/qupath/blob/main/qupath-gui-fx/src/main/resources/css/main.css#L12)).

Rylern:
I updated https://github.com/qupath/extension-manager and the related maven snapshot. This should fix the four points

qupath-gui-fx/src/main/java/qupath/lib/gui/QuPathGUI.java
```diff
@@ -342,15 +344,27 @@ private QuPathGUI(final Stage stage, final HostServices hostServices, boolean sh
...
+						"QuPath catalog",
+						"Extensions maintained by the QuPath team",
+						URI.create("https://github.com/qupath/qupath-catalog"),
+						URI.create("https://raw.githubusercontent.com/qupath/qupath-catalog/refs/heads/main/catalog.json"),
+						false
+				)))
+		);
+		ExtensionLoader.loadFromManager(extensionCatalogManager, this);
+
+        // Add scripts menu (delayed to here, since it takes a bit longer)
```
alanocallaghan, line 361:
```suggestion
                // Add scripts menu (delayed to here, since it takes a bit longer)
```

alanocallaghan:
Looks great! Few minor points:

- Should the information be a popover, similar to the implementation in WSInfer or InstanSeg extensions? Popup window is a bit blocking and clunky
- I'd still like a warning about duplicate extensions installed locally and in an extension or across catalogs, if possible, even if this is only based on string matching
- Should we make it more obvious that you can remove catalogs?

petebankhead:
I agree with all @alanocallaghan's points - including that it looks great.

From my side it could be merged and anything further handled in subsequent PRs.

My minor points:

### On popovers
I agree [popover would fit](https://controlsfx.github.io/features/popover/), although I remember they can be a bit weird either when occurring close to the screen bounds or in dark mode. The weirdness isn't crucial so we could use it anyway; otherwise, if the popup dialog was at least as easy as possible to dismiss then that would be ok (e.g. big 'Close' or 'OK' button, and responsive to `Esc` or `Ctrl+W`, see https://github.com/qupath/qupath-fxtras/pull/34)

### On dependencies
Because the dependency strings are very long, could/should they be in a `TextArea`?

It would have the advantage of making it possible to copy/paste the list, and also include a horizontal scrollbar local to the text area. But it might also just looks worse... would be keen to see a mock-up of how it looks and have a vote on whether the label or text area is preferable.

install-dependencies

Similarly, is there anything to prevent the dialog growing too large horizontally if the URLs / paths to delete are too long?

I'd also rephrase slightly to `"This will replace the existing directory [dir] and all of its content"` because it could be confusing to declare a directory is deleted when it *appears* to still be there afterwards.

delete-existing

Rylern:
> Should the information be a popover, similar to the implementation in WSInfer or InstanSeg extensions? Popup window is a bit blocking and clunky

> I agree [popover would fit](https://controlsfx.github.io/features/popover/), although I remember they can be a bit weird either when occurring close to the screen bounds or in dark mode. The weirdness isn't crucial so we could use it anyway; otherwise, if the popup dialog was at least as easy as possible to dismiss then that would be ok (e.g. big 'Close' or 'OK' button, and responsive to Esc or Ctrl+W, see https://github.com/qupath/qupath-fxtras/pull/34)

I initially used a popover, but switched to a popup window for the reasons Pete mentionned. I added a close button and responsiveness to Esc or Ctrl+W, do you think that's enough?

> I'd still like a warning about duplicate extensions installed locally and in an extension or across catalogs, if possible, even if this is only based on string matching

When I install an extension with one index and try to install the same extension (with the same version) with another index, here is the message I have:

image

Is that what you have in mind?

> Should we make it more obvious that you can remove catalogs?

I added a `Remove` button in the catalog manager:

image

> Because the dependency strings are very long, could/should they be in a TextArea?

I also prefer that. Here is how it looks:
image

(with wrapping)

image

(without wrapping)

Which one do you prefer? Also, should we also use a TextArea for the warning message in orange?

> I'd also rephrase slightly to "This will replace the existing directory [dir] and all of its content" because it could be confusing to declare a directory is deleted when it appears to still be there afterwards.

Yes

petebankhead:
> I initially used a popover, but switched to a popup window for the reasons Pete mentionned. I added a close button and responsiveness to Esc or Ctrl+W, do you think that's enough?

Sounds good to me.

My only other idea is that we could replicate the information text in a tooltip for the info button (the current text is `Get information on the extension`). Then the click isn't necessary - and if a power-user notices, they will realise they don't need to click at all. Not sure if this is useful or not, and we can always add it later if we decide it's worth it - I think the fact there is a live hyperlink justifies the use of the 'heavier' dialog component.


> When I install an extension with one index and try to install the same extension (with the same version) with another index, here is the message I have:

For me, this text is too long for a dialog, at least without newlines:
```
Catalog.ExtensionModificationWindow.extensionAlreadyInstalledDetails = An extension with the same JAR file is already installed, which may mean that this extension won't be loaded. It is recommended to delete the existing extension before proceeding. Continue?
```

Possible alternative
```
Catalog.ExtensionModificationWindow.extensionAlreadyInstalledDetails = An extension with the same JAR file is already installed.\nIt is recommended to delete the existing extension before proceeding.
```

Then you could put `Continue anyway` as the button text, instead of `OK`.

As it currently stands, as a user I wouldn't be 100% sure if clicking `OK` would install the extension as a duplicate, or if it would follow its own recommendation to delete the existing one. I'd expect it _probably_ adds a duplicate (but then the user isn't instructed *how* they could delete the duplicate).

> Which one do you prefer?

I prefer **without** wrapping.

> Also, should we also use a TextArea for the warning message in orange?

Not sure, I don't think so. But it could be center-aligned, with the directory name clearly on its own line.
I'd also like to revise the text again to
```
This will replace the following directory and all of its content:
[dir path]
```
Reasoning:
1. it is easier to localize the string if the path is not embedded
2. when the directory contains spaces it isn't totally obvious where a space is at the end of the path or splits a component of the path

An extra idea: the path itself could be given as a hyperlink, not embedded in the orange string. This would make it visually distinct, and also enable the user to open the directory to check its contents easily in the file system (would require call to `Desktop`, and must be on the correct thread [as used here](https://github.com/qupath/qupath/blob/60898ed353b2eda253d0f0800232204bde11ecc3/qupath-gui-fx/src/main/java/qupath/lib/gui/tools/GuiTools.java#L220)).

Rylern:
Thanks, all comments were addressed

petebankhead:
That's great, thanks - I'll merge now.

qupath-gui-fx/src/main/java/qupath/lib/gui/scripting/languages/ScriptLanguageProvider.java
```diff
@@ -106,8 +106,8 @@ private static List<ScriptLanguage> getDefaultLanguages() {
 	}
 	
 	
-	private static ExtensionClassLoader getExtensionClassLoader() {
-		return ExtensionClassLoader.getInstance();
+	private static ClassLoader getExtensionClassLoader() {
+		return QuPathGUI.getInstance().getExtensionCatalogManager().getClassLoader();
```
petebankhead, line 110: @Rylern I missed this when merging the PR: this breaks command line scripts,  because then `QuPathGUI` doesn't exist.

We need a method of getting the extension `ClassLoader` that has no dependency on the UI.

PR #1721, opened 2024-12-06
Add linux-arm64 platform

petebankhead:
In preparation for it becoming useful at some point when relative dependencies are available.


alanocallaghan:
Openslide snapshot on scijava maven: https://maven.scijava.org/content/repositories/snapshots/io/github/qupath/openslide/4.0.0.6-SNAPSHOT/

PR #1724, opened 2024-12-06
Update OpenSlide/c-blosc

petebankhead:
Use 1.21.6+01 for blosc (includes linux arm64) and 4.0.0.6 release for openslide


cbuscaron:
Thank you Pete. We will test this as well.

Issue #1733, opened 2024-12-17
x86_64 macOS Unable to Use Latest PyTorch via DJL in QuPath 0.6


jaredbrewer:
Before submitting your bug report, please check the following:

* [x] I've definitely found a bug (it you're not sure, please use [image.sc](https://forum.image.sc/tags/qupath) instead)
* [x] I've checked https://qupath.github.io for a new release that might already have fixed the issue
* [x] I've checked the [Changelog](https://github.com/qupath/qupath/blob/master/CHANGELOG.md) to see if the bug has already been fixed in the next release
* [x] I've checked for existing GitHub issues describing the same problem

### Describe the bug
QuPath very nicely continues to support x86-based macOS systems but the implementation of the DeepJavaLibrary in tandem with PyTorch's deprecation of x86-based macOS support means that PyTorch does not work within QuPath and, thus, InstanSeg.

### To Reproduce
On a macOS x86_64-based system, attempt to download PyTorch from the Extension > DeepJavaLibrary > Manage DJL engines dialog. This will fail without a meaningful error beyond "Unable to initialize PyTorch, sorry."

### Expected behavior
There are a few potential options here:
- Automatically download the latest compatible version of PyTorch for a given operating system and architecture
- Provide additional instruction about how to download and use a different version of PyTorch within QuPath
- Create a TensorFlow-compatible version of InstanSeg, which remains compatible with Intel Macs and there are no plans (to my knowledge) to deprecate this support in the immediate future.

### Desktop (please complete the following information):
 - OS: macOS 15.2 Sequoia
 - QuPath Version: 0.6.0-rc3

### Additional context
While Apple no longer sells new Macs running on Intel, there remains a large install base and specific use cases where an ARM-based Mac is not a viable alternative. PyTorch's choice to deprecate these systems in early 2024 was disappointing but most of the major features and models remain freely compatible among PyTorch 2.x version (for now). It would be useful to provide as much support as possible for the systems that remain in use.

petebankhead:
I've removed the 'bug' label since it was a conscious decision, but I've made a PR to address this at https://github.com/qupath/qupath/pull/1734

[...]

PR #1734, opened 2024-12-18
Mac x64

petebankhead:
Possible strategy for https://github.com/qupath/qupath/issues/1733

PyTorch has dropped support for Intel Macs, and so too has DJL since v0.30.0. 

Downgrading to DJL v0.29.0 works, but only if we also launch with a system property that requests PyTorch 2.2.2 (rather than the default choice of v2.3.1).

This PR therefore sets the DJL version to be v0.29.0, and sets the system property *only* when building on an Intel Mac.

Advantages:
* It makes it possible to run InstanSeg (slowly) for users of Intel Macs. It also helps a significant proportion of Apple Silicon users; for as long as Bio-Formats doesn't fully support Apple Silicon (e.g. https://github.com/ome/bioformats/issues/3858) there people with some slide scanners (e.g. Zeiss Axionscan) that are forced to resort to the Intel builds.

Problems with this:
* There is no GPU support, so inference is *slow*
* There is no possibility to upgrade beyond this; we can't use more recent DJL releases without dropping support for Intel Macs, or causing them to use a different version from other platforms.
* Because the DJL version affects all engines, it also locks us to a specific TensorFlow version. This is a reason we can't use v0.28.0 or earlier, because then we've had no support for TensorFlow on Apple Silicon.

Because it seems certain that we'll need to upgrade to new DJL releases before long, I've also added support for a system property to specify a different DJL when building, e.g.

```
gradlew run -Ddjl.version=0.24.0\!\!
```
The `\!\!` is optional, but helps force through the choice in case another dependency might request a different version.

We can't commit to supporting this, but it means that if we end up having to 'break' Intel Mac support in the next release then we can suggest a temporary workaround to create a custom build for someone who needs it... and who really wants PyTorch, at the expense of TensorFlow.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -113,7 +113,7 @@ They may change or be removed in future versions.
 ### Dependency updates
 * Bio-Formats 8.0.1
 * Commonmark 0.24.0
-* DeepJavaLibrary 0.30.0
+* DeepJavaLibrary 0.31.0
 * Groovy 4.0.22
 * Gson 2.11.0
 * Guava 33.3.1-jre
--- a/buildSrc/src/main/java/io/github/qupath/gradle/PlatformPlugin.java
+++ b/buildSrc/src/main/java/io/github/qupath/gradle/PlatformPlugin.java
@@ -32,7 +32,7 @@ public class PlatformPlugin implements Plugin<Project> {
      */
     public enum Platform {
         WINDOWS("windows", "win32-x86_64", "ico", "msi"),
-        MAC("macosx", "darwin-x86_64", "icns", "pkg"),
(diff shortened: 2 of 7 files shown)


petebankhead:
With this PR, I can run InstanSeg via Rosetta2 on my M3 MacBook Pro.

However I *can't* run it on my Intel 2014 MacBook Pro because it's limited to macOS 11. As described in the [DJL docs](https://djl.ai/engines/pytorch/pytorch-engine/)

> PyTorch 1.13+ doesn’t support mac 11 any more, you must use DJL 0.19.0 ane lower version.

This is *much* too far a leap back, so the DJL extension won't load properly. Therefore even this PR can only help a subset of Intel Mac users.

Issue #1738, opened 2024-12-20
QuPath window closes while removing an image from the project


marbre2:
Hello, 
I did not find this bug in the issues. I hope you can find the origin of the bug and fix it.
Thank you for your time and development,

Best regards,

Marine 
### Describe the bug
I am testing the latest version of QuPath with InstanSeg.
I have created a project with multiplex images.
I created a small training image from the bigger image, but then notice it is not representative of my task.
So I close that small training image, and then ask to remove it from the project.
The software ask me if I want to delete all info, I click on Yes, and the software immediatly after crashes. I repeated the experience twice.

Please fin the error log in attachement

### Desktop (please complete the following information):
 - OS: Windows 11 Pro, version 24H2
 - QuPath Version: 0.6.0-rc3

### Additional context
[hs_err_pid21132.log](https://github.com/user-attachments/files/18208699/hs_err_pid21132.log)
[hs_err_pid5776.log](https://github.com/user-attachments/files/18208700/hs_err_pid5776.log)

petebankhead:
Hi Marine, I'm not able to replicate the error but I do not have a Windows 11 computer to test on - I can only try Windows 10 or Mac.

I made a change at https://github.com/qupath/qupath/pull/1739 that *might* help, but I have no way to be sure.

The error log shows that the crash occurs when attempting to move the data directory to the Windows recycle bin. Because that happens deep within the operating system, we have very little control within QuPath's code. We already ask whether the system supports moving the file to the trash, and only attempt it if the answer is 'yes'... so we aren't able to predict or handle the failure.

The only thing I can think of that could help is to request the deletion from a different thread. That's what the code change does.

petebankhead:
Hi @marbre2 could you confirm if this is still happening - and if it happened with earlier QuPath releases?

None of us can replicate it, but none of us have Windows 11.

marbre2:
HI @petebankhead,

Thank you very much for your quick reply and investigation.

My apologies for the delay of reply. I have replicated the problem with QuPath 0.5.1
So it seems this is not exclusively related to v0.6

Although, QuPath (both v0.6.0_rc3 and v0.5.1) did not crash when I click "No" to the question "Delete all associated data?"
Could this information be of any help?

Best regards,

petebankhead:
Thanks, in that case it *might* be something specific to your system - since I don't think it has been reported before, and v0.5.1 has been available for a long time.

[...]

PR #1739, opened 2024-12-20
Move project data to trash in EDT

petebankhead:
May help with https://github.com/qupath/qupath/issues/1738 - but could not replicate the problem, so not sure.

If it *does* help, then `GeneralTools.deleteFile(File)` should also be updated - but this is more difficult due to the need to return a boolean indicating whether the delete was successful.

### DIFF
--- a/qupath-core/src/main/java/qupath/lib/common/GeneralTools.java
+++ b/qupath-core/src/main/java/qupath/lib/common/GeneralTools.java
@@ -4,7 +4,7 @@
  * %%
  * Copyright (C) 2014 - 2016 The Queen's University of Belfast, Northern Ireland
  * Contact: IP Management (ipmanagement@qub.ac.uk)
- * Copyright (C) 2018 - 2022 QuPath developers, The University of Edinburgh
+ * Copyright (C) 2018 - 2025 QuPath developers, The University of Edinburgh
  * %%
  * QuPath is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as
@@ -68,6 +68,8 @@ import org.slf4j.LoggerFactory;
 import com.google.common.io.CharStreams;
 import com.google.gson.Gson;
 import com.google.gson.reflect.TypeToken;
+
+import javax.swing.SwingUtilities;
 
 /**
  * Collection of generally useful static methods.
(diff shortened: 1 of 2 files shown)


petebankhead:
Could someone with access to Windows 11 please check if you can replicate the original problem, and if so whether this makes any difference?

If `Desktop.moveToTrash` fails with `EXCEPTION_ACCESS_VIOLATION` routinely on Windows 11 then that would be a *major* problem that I'd expect to affect other things - including extension management.

petebankhead:
Updated to ensure that `GeneralTools.deleteFile` uses the EDT on Windows. That is problematic on Mac though, so the code paths need to be different.

Here's a script to quickly test if it is working
```groovy
var file = File.createTempFile("my-temp", ".tmp")
println file
println "Exists before: " + file.exists()
println "Success deleting: " + GeneralTools.deleteFile(file, true)
println "Exists after: " + file.exists()
```

By default, this will run in its own thread. We also need to confirm that the method works when called from the JavaFX application thread; to do that easily, add `guiscript=true` at the top, i.e.

```groovy
guiscript=true

var file = File.createTempFile("my-temp", ".tmp")
println file
println "Exists before: " + file.exists()
println "Success deleting: " + GeneralTools.deleteFile(file, true)
println "Exists after: " + file.exists()
```
This is the variation would fail on macOS if we aren't careful, because the JavaFX application thread and Swing EDT are really the same thread at the operating system level.

petebankhead:
@alanocallaghan and @finglis could you check this hasn't broken things for you on Linux or Windows please?

See example scripts in my last post. I'm not sure it'll help, but I'd merge it if it doesn't hurt.

Rationale: using `Desktop` from a different thread usually seems fine, but it was problematic once before on Linux so my guess is that there *could* be more system-dependent badness: https://github.com/qupath/qupath/commit/5564eff277e2b051afd990c53a61e15ec0fec105

alanocallaghan:
Script runs the same for me on current main and on this branch which I guess means nothing is horribly broken?

petebankhead:
Both with `guiscript=true` and without it, right? i.e. testing it on the JavaFX Application thread and the 'regular' script thread.

PR #1744, opened 2025-01-10
Update BioimageIoTools to handle 0.5 models

alanocallaghan:
See https://github.com/qupath/qupath-bioimageio-spec/pull/13 for details on the spec updates.

Also removes `BioImageioSpec` from the core classes - this is unlikely to have any major impacts, but should be mentioned in the release notes.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,7 +1,7 @@
 [versions]
 
 bioformats      = "8.0.1"
-bioimageIoSpec  = "0.2.0"
+bioimageIoSpec  = "0.2.1"
 omeZarrReader   = "0.5.2"
 blosc           = "1.21.6+01"
 
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,7 +1,7 @@
 [versions]
 
 bioformats      = "8.0.1"
-bioimageIoSpec  = "0.2.1"
+bioimageIoSpec  = "0.2.0-SNAPSHOT"
 omeZarrReader   = "0.5.2"
 blosc           = "1.21.6+01"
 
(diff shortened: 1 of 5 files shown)


Rylern (approved):
The code looks good, however I haven't tested it (I'm not very familiar with all of that)

gradle/libs.versions.toml
```diff
@@ -1,7 +1,7 @@
 [versions]
 
 bioformats      = "8.0.1"
-bioimageIoSpec  = "0.2.0-20250131.094718-2"
+bioimageIoSpec  = "0.2.0-20250204.102546-3"
```
petebankhead, line 4: Why not use `0.2.0-SNAPSHOT` and we fix it at the time of release?

qupath-core-processing/src/main/java/qupath/opencv/ml/PatchClassifierParams.java
```diff
@@ -267,36 +266,36 @@ private Builder(PatchClassifierParams params) {
 		
 		/**
 		 * Define the input channels using channel names.
-		 * @param channels
-		 * @return
+		 * @param channels The input channel names.
+		 * @return A modified builder.
```
petebankhead, line 270: I find this misleading - should clarify it returns *this* builder.

A caller who has ever encountered a builder would expect it to be modified if we know that `this` is returned, but from the javadoc here it's not clear to me if builders are mutable or immutable. I wouldn't know if I *must* work with the returned builder (because it's different), rather than that I *can* use the returned builder (because it's the same object... just updated).

petebankhead (changes requested):
I haven't tried running this, but the changes all make sense - just two comments to consider.

petebankhead (approved):

PR #1747, opened 2025-01-16
Refactor summary measurements, add tooltips

petebankhead:
## Background

QuPath has two main tables showing object measurements:
* The summary table for individual objects (under the 'Annotation' and 'Hierarchy' tabs)
* The summary table for all objects of a specific type (under the 'Measure' menu or toolbar button)

Screenshot 2025-01-23 at 18 11 59

Historically, these have relied upon one monster class: `ObservableMeasurementTableData`, which does a *lot* of work:

* Given an image and collection of objects, figures out which measurements will be relevant, e.g.
  * live ROI areas if we have annotations
  * detection counts only if we have detections
  * H-scores, positive cell counts only if we have detections with the right kinds of classifications
  * measurements from an object's `MeasurementList` only if these are available
* Defines nested classes for dynamic measurements where needed
  * e.g. H-score, positive percentage, ROI area calculated on demand
* Handles caching results for some expensive measurements - but resetting these when they are outdated
* Manages JavaFX bindings and lists so all these stuff can easily be shown in a table

## The problems

* [`ObservableMeasurementTableData`](https://github.com/qupath/qupath/blob/v0.6.0-rc3/qupath-gui-fx/src/main/java/qupath/lib/gui/measure/ObservableMeasurementTableData.java) was huge & unmaintainable
  * Almost 2,000 lines
  * Lots of both static & non-static nested classed
  * Really hard to decipher, test & improve
* Stuck with JavaFX dependencies
  * There was *no way* to get these measurements from `qupath-core` or `qupath-core-processing` - they always required pulling in JavaFX
* Can't make measurements 'smarter' without hacking `ObservableMeasurementTableData` even more
  * e.g. create dynamic measurements that are only relevant for different image types (e.g. multiplexed, electron microscopy)

## This PR

This PR was motivated by the desire to add tooltips to measurements to help explain what they do.

Ultimately, it led to breaking up `ObservableMeasurementTableData` entirely and removing the JavaFX dependency for individual measurements.

*In its current form it is far from complete. We should go further - but may need to wait so we don't delay v0.6.0*

### Introducing `LazyValue`

An old (non-public) interface `MeasurementBuilder` used to create JavaFX bindings for each measurement; this is replaced by `LazyValue<S, T>`.

Its 3 most important methods are:
```java
    /**
     * Calculate a value from the input.
     * @param input the input that should be measured
     * @return the output value
     */
    T getValue(final S input);

    /**
     * Optional help text that explained the value.
     * This may be displayed in a tooltip.
     * @return the help text, or null if no help text is available
     */
    String getHelpText();

    /**
     * The name of the value.
     * When showing a measurement table, this would be the column header.
     * @return the name of the value
     */
    String getName();
```
There are some other methods (including default methods) to request string representations of the value and query the output type.

This completely removes the JavaFX dependency, and so has moved out of the UI module to `qupath-core-processing`.

It also doesn't assume that the input will be a `PathObject` - so we may reuse the interface for project entries, for example.

But if you need a JavaFX binding, this should be easy to create.
For example, if you know the input is a specific `PathObject` and the output is a `Number` then you might use

```java
Bindings.createDoubleBinding(() -> lazyValue.getValue(pathObject));
```

### Obtaining a specific `LazyValue`

The `PathObjectLazyValues` provides a lot of `LazyValue` implementations relevant for `PathObject` input.

In some cases, a singleton is enough: the value can be calculated based on the object and nothing else.

In other cases, the `LazyValue` needs to be created by also passing an `ImageData` (e.g. so that it can request pixel size information to calibrate ROI measurements).

> The `ImageData` is passed to the constructor, and stored internally; this means that the `LazyValue.getValue(PathObject)` still only takes one input parameter, and can still be used even if new objects are added to the image.

### Figuring out what to show in a table

Currently, `ObservableMeasurementTableData` is still used to generate the underlying data for QuPath's JavaFX tables.

It does this by taking an `ImageData` and a collection of `PathObject` and asking a `PathObjectValueFactory` to select/create a list of `LazyValue` that are relevant.

> For now, we have one main `PathObjectValueFactory` implementation as far as the user is concerned.
> But using a different `PathObjectValueFactory` would allow the table to show different dynamically-generated measurements - see future plans below.

For a main measurement table (*Measure → Show detection measurements*, which displays values for lots of objects)
* each row in the table corresponds to a `PathObject`
* each column corresponds to a `LazyValue` (and the name of the column is the name of the `LazyValue`)
* each cell of the table is the result of applying `LazyValue.getValue(pathObject)` using the column and row
  * the tooltip comes from the `LazyValue` help text

For the single-object measurement tables under the 'Hierarchy' and 'Annotation' tabs, the idea is the same but the `LazyValue` are shown as rows instead of columns.

### Caching

Where possible, `LazyValue.getValue(S)` should compute the value quickly and only on demand, so that it is always current.

One time when this doesn't happen is for detection counts (e.g. finding all the detections inside an annotation with a specific classification).
When many classifications are used, there can be lots of these measurements - and testing which detections fall inside an annotation ROI can be expensive.

A measurement table can potentially have hundreds of thousands of objects in it, so we need the calculations to be fast.

Therefore these values are calculated in a more complicated way, with the help of `DetectionClassificationCounter` to look through the object hierarchy once.

But there needs to be a way to determine when the cached value can be used, and when it might be out of date.

This depends upon `PathObjectHierarchy`, but each measurement adding a listener would add a lot of overhead - and complicate the code because the listeners would need to be removed.

With this PR, a simple pragmatic approach is used by adding a new method `PathObjectHierarchy.getEventCount()`.
Then we can can check on demand if more hierarchy events have been fired since the last value was cached; if yes, then the cache is discarded. If no, we can return any cached value with confidence that it should still be correct.

## Current advantages

The most visible advantage currently are:
* measurements have tooltips
* more measurements update automatically when interactively moving annotations

For developers, it's much easier to track down the code to calculate any specific measurement.

401944726-13f7a081-6cb7-435f-86c2-4aed42744cd9

These improvements are pretty modest, but it should matter more in the future.

### Streamline export
Move more measurement-related code out of the UI, and streamline measurement export - including across multiple images.

Currently, there is duplication of logic across
* https://github.com/qupath/qupath/blob/a8041181efbb298f60297315fda6f573bc35b8ee/qupath-gui-fx/src/main/java/qupath/lib/gui/commands/MeasurementExportCommand.java#L99
* https://github.com/qupath/qupath/blob/a8041181efbb298f60297315fda6f573bc35b8ee/qupath-gui-fx/src/main/java/qupath/lib/gui/tools/MeasurementExporter.java#L60

This will also help customize the export, e.g. [specifying the number of decimal places](https://forum.image.sc/t/measurements-differ-from-actual-detection-measurements-seen-in-annotation-window/107445/7).

### Smarter measurements
A goal of this is that we can more easily adapt to give different kinds of measurements for different applications.

For example, we might use different logic to figure out which measurements to show if we have a highly-multiplexed images rather than a brightfield H&E or IHC.

Adding better support for things like EM or cell tracking would also require measurement tables that are different.
`PathObjectValueFactory` helps with that.

### Custom measurements
It should be possible for scripts and extensions to add custom dynamic measurements.
Again, `PathObjectValueFactory` can help.

As proof-of-concept, this is already possible with the following (unsupported, not recommended) call to a non-public method:

```groovy
import qupath.lib.gui.measure.*
import qupath.lib.lazy.interfaces.*

Platform.runLater {
    
    ObservableMeasurementTableData.addFactory(wrapper -> {
        if (wrapper.containsAnnotations()) {
            return [LazyValue.createNumeric(
                        "Dynamic solidity",
                        "Dynamic solidity help text!", 
                        p -> p.getROI().getSolidity()
                    )]
        } else {
            return []
        }
    });
    
}
```

This adds a new solidity measurement to measurement tables showing annotations only, without changing the objects directly.

### Distinguish where measurements come from
Some table values are computed dynamically, some are extracted from `PathObject.getMetadata()` and others come from `PathObject.getMeasurementList()`.

Values coming from `getMeasurementList()` have some important features:
* they can be used to interactively train object classifiers
* they do *not* update automatically

Therefore it would be helpful for tables to show these differently.

Here's a proof-of-concept to show how it could work (although it's not currently in the code, partly because I'm not sure which icon we should use):

Screenshot 2025-01-23 at 18 20 50

### Better testing

Previously, we could test measurements only via `ObservableMeasurementTableData`.
With these changes, we can (and should) add more tests for individual measurements.


alanocallaghan:
The interfaces look sensible and everything seems to work as expected. The detailed doc here will surely come in handy in future. Looks great to me; I wouldn't have further comments unless I spent a lot of time digging into the new API

Rylern:
Same, looks good to me

petebankhead:
Excellent, thanks. Will merge now & continue working on it at some point.

Issue #1644, opened 2024-09-18
Shorten Classifications menu for easier access


lacan:
The most common things to do in the classifications list in my experience is to
1. Create or delete classes
2. Show or hide classes

One option for Show/Hide would be to have the right click menu be context sensitive:
- If one or multiple classifications are selected and visible, the context menu becomes "Hide selected classes" (Instead of show/Hide > Hide classes in viewer)
- If one or multiple classifications are selected and hidden, the context menu becomes "Show selected classes" (Instead of show/Hide > Show classes in viewer)

One option for Add/remove would be that the right click menu:
- Includes an "Add new classification" at the root of the menu
- If one or multiple classifications are selected, change the menu to include "Remove selected classifications" (Instead of Add/remove > Remove class, which is a slight misnomer as it's possible to remove multiple classes if they are selected)

While I know that there is the <kbd>Spacebar</kbd> shortcut for the Hide/Show classes that is way more used, I thought I'd suggest these changes, especially for adding classifications.

Best

Oli

petebankhead:
> One option for Show/Hide would be to have the right click menu be context sensitive:
> 
> * If one or multiple classifications are selected and visible, the context menu becomes "Hide selected classes" (Instead of show/Hide > Hide classes in viewer)
> * If one or multiple classifications are selected and hidden, the context menu becomes "Show selected classes" (Instead of show/Hide > Show classes in viewer)

What about if multiple classifications are selected, some shown and some hidden?

I'm apprehensive about trying to make the code too clever, and confusing people more by the menu changing when they click it.

If the core issue is that the common things are too hard to access, would simply moving `Show/hide` higher up the menu be a solution?

My hesitation with that is that currently the top part of the menu (above the divider) is all concerned with adding/removing classifications. So `Show/Hidez would have to go to the *very* top, to avoid interrupting this logical grouping... and, as you say, spacebar does that job, so `Add/Remove` might be needed more often.

lacan:
> I'm apprehensive about trying to make the code too clever, and confusing people more by the menu changing when they click it.

Yeah that makes sense! Maybe context sensitive isn't the way to go. Thank you for considering it though! :)

> If the core issue is that the common things are too hard to access, would simply moving Show/hide higher up the menu be a solution?

This could definitely work as well. I would be super happy to have a one menu access for "Add/Remove", in the same vibe as "Estimate stain vectors" was given its own high-leven location instead of a submenu.

petebankhead:
[...]

PR #1752, opened 2025-01-24
More control to show/hide objects based on class (& other criteria)

petebankhead:
This PR cleans up the classification list and introduces icons so that it is easier to see that visibility can be toggled.

We still need to figure out what the right behavior is when toggling...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,12 @@ Some things may be added, some things may be removed, and some things may look d
 * Read and write OME-Zarr images (https://github.com/qupath/qupath/pull/1474)
   * Use `convert-ome` to write OME-Zarr from a command line
   * Use *File → Export images... → OME-Zarr* to export images from the user interface
+* Add, remove & upgrade extensions easily with *Extensions → Manage extensions* (https://github.com/qupath/qupath/pull/1716)
+  * See also https://github.com/qupath/extension-manager
+* More control over showing/hiding objects by classification (https://github.com/qupath/qupath/pull/1752)
+* New ImageJ script runner (https://github.com/qupath/qupath/pull/1682)
+* Improved measurements - with tooltips & scatterplots (https://github.com/qupath/qupath/pull/1747 & https://github.com/qupath/qupath/pull/1544)
+* Optionally build QuPath *and* Fiji together (https://github.com/qupath/qupath/pull/1728)
 
 ### Enhancements
 (These are not yet ordered by interestingness)
@@ -53,6 +59,7 @@ Some things may be added, some things may be removed, and some things may look d
   * Note that channel colors still only be set for non-RGB images
 * Improved ImageJ integration (https://github.com/qupath/qupath/pull/1676 https://github.com/qupath/qupath/pull/1677)
 * 'Selection mode' now supports 'deselecting' objects by pressing the 'Alt' key (https://github.com/qupath/qupath/issues/1704)
+* More 'filter boxes' in the user interface to find things in long lists or tables (https://github.com/qupath/qupath/pull/1742)
 
 ### Experimental features
 These features are included for testing and feedback.
--- a/qupath-core/src/main/java/qupath/lib/objects/classes/PathClass.java
+++ b/qupath-core/src/main/java/qupath/lib/objects/classes/PathClass.java
@@ -176,12 +176,12 @@ public final class PathClass implements Comparable<PathClass>, Serializable {
 	/**
 	 * Cache the set representation
 	 */
-	private transient Set<String> set;
+	private transient volatile Set<String> set;
 	
 	/**
 	 * Cache the list representation
 	 */
-	private transient List<String> list;
+	private transient volatile List<String> list;
 
 	private PathClass(UUID mySecret) {
 		if (!Objects.equals(secret, mySecret))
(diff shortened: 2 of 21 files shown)


petebankhead:
I've made further changes & tried to tackle https://github.com/qupath/qupath/issues/1644

There are now buttons to add & remove classifications. This makes it easier for users to find and means that the right-click popup menus can be much shorter.

Classification toggling is now more clearly defined - as shown in the screenshots.

hiding

An icon with an open eye will indicate the classification is visible, an eye with a stroke will indicate it is not.

Taking an example of `Tumor`, then setting the visibility to false will mean that *all* objects with `Tumor ` *anywhere* in their classification will be hidden. This includes `Tumor `, `Tumor: Positive`, `Positive: Tumor`... but not `Stroma` or `Positive`.

Similarly hiding `Positive` will result in *all* objects with `Positive` in their classifications to be hidden - regardless of whether they are `Tumor`, `Stroma` or something else. Previously this wasn't possible.

This is the same for multiplexed images, where classifications often have multiple parts.

It is assumed that, most of the time, all the possible combinations won't be shown in this list: usually the list will have 'single' classifications, one for each marker (and not for combinations of markers).

Taking an example of `CD8`, then setting the visibility to false will mean that *all* objects with `CD8` *anywhere* in their classification will be hidden. This includes `CD8`, `CD3: CD8`, `CD8: CD3`... but not `CD68`.

But derived classifications *can* be shown, as in this screenshot:

better-eyes

Setting a derived classification to be hidden means that any object with all the parts of the hidden classification will be hidden. In the screenshot, `CD8: CD68` being hidden means that `CD8: CD68: PD1` is automatically hidden as well... but `CD8` or `CD68` on their own would not be.

This is indicated with an `X` icon: `CD8: CD68: PD1` isn't explicitly hidden, but any objects with that classification won't be visible because of other classifications chosen to be hidden.

The tooltip text attempts to explain this, making the information also available under the context help.

This may seem complicated and won't deal with every scenario, but it's the best I've come up with so far. I think it feels intuitive to use, but need more people trying it for longer to be sure of that. Certainly, the v0.5.x behavior *didn't* feel intuitive to me at all for multiplexed images - because it had much more tortured logic to deal with intensity subclassifications as a special case...

> The big limitation with this approach is that you can't hide `CD8` but show `CD3: CD8`. Alternative ways to do that might be added later.

alanocallaghan:
The X looks to me like clicking it will delete that class

petebankhead:
Fair, any ideas for a better icon? Easiest if it's from https://fontawesome.com/v5/search?ic=free - but ControlsFX doesn't have all of them, so we're even more restricted than it first appears (unless we add another dependency... or can draw it ourselves)

(There's one `low-vision` that I wanted, but not in ControlsFX)

lauranicolass:
Empty circle vs circle with check?
image
image

petebankhead:
Actually it is possible via ControlsFX by using
```java
GlyphFontRegistry.font("FontAwesome")
	.create('\uf2a8')
	.size(QuPathGUI.TOOLBAR_ICON_SIZE)
```

low-vision

petebankhead:
Hi @lauranicolass sorry, missed you’d posted this in between. I quite like the circle-based icons - and there are more options (checks, crosses, filled, unfilled). Would you replace the eyes entirely? Which icons specifically for each of these 3 conditions?

1. Visible
2. Hidden (explicitly)
3. Hidden (because a related class is hidden)

petebankhead:
Continuing with this, I've introduced titled panes for both the class and annotation lists.
The annotation list now has a filter box as well.

The titled panes contain buttons, which allow us to save some space at the bottom.

I think this starts to make more functionality visible in the UI and easier to access.

annotation-titles

petebankhead:
Added `OverlayOptions.showObjectPredicateProperty()`.

This aims to make it possible to control which objects are visible with more detail than any other methods. It should be used sparingly (and triggers a warning in the context help), but can be useful.

Example script:

```groovy
Platform.runLater {
    var o = getQuPath().getOverlayOptions()
    o.setShowObjectPredicate {
        p -> p.measurements['Nucleus: Area'] > 100
    }
    // Call this when done!
//    o.resetShowObjectPredicate()
}
```

And screenshot:
Screenshot 2025-01-26 at 18 01 39

Although it adds some complexity, I think the benefits may be worth it as it can be very powerful.

Here's another example, this time intended for multiplexed images, to show all objects with a *neighbor* that is CD8 positive (based on a Delaunay triangulation using centroids):

```groovy
Platform.runLater {
    var hierarchy = getCurrentHierarchy()
    var o = getQuPath().getOverlayOptions()
    o.setShowObjectPredicate {
        p -> hierarchy.findAllNeighbors(p).find(n -> 'CD8' in n.classifications) != null
    }
}
```

alanocallaghan:
This looks great! Only question I have is whether the option to be more or less liberal with showing/hiding detections should be a persistent preference

alanocallaghan:
I err on the side of persistent, yeah - certainly I can imagine grumbling if I always want it one way or the other. Persistent may confuse people who don't realise it's set, but I think that's less problematic given there's good UI indications anyway

MichaelSNelson:
One more thought while you guys are working on this, it would be nice if there were (and may already be, I have not dug through everything) a quick way to filter by derived class length. Be it a sort by length, filter by 0, 1, ... n colons, or what have you. We can already 'Populate' base classes or all classes, but it would be nicer to be able to change and adjust the visibility rather than repopulating the list - especially in cases where not all classes will be present in every image. 
If it isn't too difficult!

petebankhead:
I don't think I understand, but it sounds like it might be possible using the regex feature in the filter field under the list of classifications. Or using the new predicate option of `OverlayOptions`.

May need to be more related changes in a day or two after we've used this a bit more (and been annoyed by its limitations)

MichaelSNelson:
Ah, if the filter accepts regex, then it is definitely already possible. 
Probably not convenient for regular users multiplexing, but then I am not sure there _is_ a convenient way to do all of the sorting you'd want to do with highly multiplex data sets without some scripting or regex. Good luck and I hope we can use this by the end of February! :)

PR #1755, opened 2025-01-29
Make extension manager window not modal

Rylern:
Initially, the extension manager window was modal, but I don't know if there's still a valid reason for that.


alanocallaghan:
I don't have super strong opinions on which dialogs should be modal, so this seems fine

petebankhead:
No very strong opinion from me either. I think modal is justified because extension management isn't something that should be done in parallel with processing: we don't really want people to be trying to remove/overwrite extensions while they may be in use. But I can't think of a problem that would arise from making it non-modal, so if there are advantages in doing so then happy to merge this.

PR #1759, opened 2025-02-03
Fix error when starting a script from the command line

Rylern:
Fix https://github.com/qupath/qupath/pull/1716#discussion_r1939224970 by making `ExtensionCatalogManager` a static member of `QuPathGUI`.

This is not a satisfactory solution because the `ExtensionCatalogManager` instance is created in `QuPathGUI` even when the user interface is not used. However, it was the simplest PR I could think of. A better way would have been to move `ScriptLanguageProvider` away from `qupath-gui-fx` (because it can be used without the UI) and make the methods of `ScriptLanguageProvider` not static.

### DIFF
--- a/qupath-app/src/main/java/qupath/QuPath.java
+++ b/qupath-app/src/main/java/qupath/QuPath.java
@@ -46,14 +46,11 @@ import picocli.CommandLine.IVersionProvider;
 import picocli.CommandLine.Option;
 import picocli.CommandLine.Parameters;
 import picocli.CommandLine.ParseResult;
-import qupath.ext.extensionmanager.core.ExtensionCatalogManager;
-import qupath.ext.extensionmanager.core.savedentities.Registry;
-import qupath.ext.extensionmanager.core.savedentities.SavedCatalog;
 import qupath.lib.common.GeneralTools;
 import qupath.lib.common.Version;
 import qupath.lib.gui.BuildInfo;
 import qupath.lib.gui.QuPathApp;
-import qupath.lib.gui.UserDirectoryManager;
+import qupath.lib.gui.QuPathGUI;
 import qupath.lib.gui.extensions.Subcommand;
 import qupath.lib.gui.images.stores.ImageRegionStoreFactory;
 import qupath.lib.gui.logging.LogManager;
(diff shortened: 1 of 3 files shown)


petebankhead (approved):

petebankhead:
Thanks, I'll merge this. I agree that moving scripting away from the GUI would be beneficial in the long term.

It seems a little odd to me that the `ExtensionClassLoader` is created inside the `ExtensionCatalogManager` and must be requested through it, rather than only being managed / updated by the manager. But maybe I just need to spend more time understanding the design.

In any case, I think this is working now and behavior looks good from a user perspective - thanks.

Rylern:
In the design, I tried to have only one non-UI public class, which is the `ExtensionCatalogManager`. This was to reduce the public API of the library and make it simpler to work with

petebankhead:
That makes sense, although [`ExtensionCatalogManager`](https://github.com/qupath/extension-manager/blob/2a1cf01d3995e479e27a222492f2bd3776712574/extensionmanager/src/main/java/qupath/ext/extensionmanager/core/ExtensionCatalogManager.java) depends upon JavaFX so I think of it as a UI class even though it doesn't directly include any `Node`.

Ultimately, we generally want [`ExtensionClassLoader`](https://github.com/qupath/extension-manager/blob/2a1cf01d3995e479e27a222492f2bd3776712574/extensionmanager/src/main/java/qupath/ext/extensionmanager/core/ExtensionClassLoader.java#L21) - which doesn't depend on the UI or JavaFX, but also isn't public.

`ExtensionClassLoader` used to only be accessible from `QuPathGUI`, but then was moved out of it at https://github.com/qupath/qupath/commit/39eaa5dadbf7ea44646e0915d5f772e7f16e5d7f so that it could be requested separately. This isn't entirely necessary though.

Rylern:
> That makes sense, although [ExtensionCatalogManager](https://github.com/qupath/extension-manager/blob/2a1cf01d3995e479e27a222492f2bd3776712574/extensionmanager/src/main/java/qupath/ext/extensionmanager/core/ExtensionCatalogManager.java) depends upon JavaFX so I think of it as a UI class even though it doesn't directly include any Node.

Should we have a small library outside of JavaFX that provides object properties and observable collections then? I'm often using these from JavaFX for non-UI code

petebankhead:
I don't think it's worth the effort at this point when it's fine for us to use JavaFX. But the properties / observable lists in JavaFX can't be accessed from the core QuPath modules, so anything that uses them needs to be in the UI module.

Even `javafx-base` requires platform-specific jars, although it looks like [the jars are much the same size](https://repo1.maven.org/maven2/org/openjfx/javafx-base/23.0.2/) and I don't see any native libraries included within them... so I'm not sure why that's necessary. It would certainly be nice to have access to properties & observables without introducing extra JavaFX complexity.

PR #1760, opened 2025-02-04
Improve handling of bioimageio models

alanocallaghan:


alanocallaghan:
Sorry just noticed the failing CI, will fix that before it's reviewable

Rylern (approved):

alanocallaghan:
> It looks like this takes the input shape to be the minimum input shape as provided by the bioimage.io shape. That feels dubious to me; if the min shape is small and a model will be tiled with overlap, then this could be incredibly inefficient with loads of redundant calculation. And if the overlap is >= the min size, then it may never finish.

I would tend to agree - it took me a long while to realise that `getShape` was really getting the minimum shape, hence renaming the method. I've no idea why, so I'm reluctant to change just now, but I think we should in future for these reasons and to better handle other types of shape.

> if you understand it now then please add more comments to benefit future Us.

Will do

petebankhead (approved):
Fair enough, certainly helps to make explicit that the minimum shape is being used. Suspect we'll need to revisit it.

alanocallaghan:
Yes, I think especially with the new bioimageio format it would be great to add proper support for at least a few common types of model in a streamlined way

PR #1764, opened 2025-02-05
Fix ProcessBuilder bug in GitCommitPlugin.java

yvesuza:
incorrect usage of Process.inputReader()

replaced inputReader() with a proper BufferedReader wrapping InputStreamReader

### DIFF
--- a/buildSrc/src/main/java/io/github/qupath/gradle/GitCommitPlugin.java
+++ b/buildSrc/src/main/java/io/github/qupath/gradle/GitCommitPlugin.java
@@ -9,7 +9,11 @@ import org.gradle.api.Project;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.io.BufferedReader;
+import java.io.InputStreamReader;
+
 import java.util.stream.Collectors;
+
 
 public class GitCommitPlugin implements Plugin<Project> {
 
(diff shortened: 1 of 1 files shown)


petebankhead:
> incorrect usage of Process.inputReader()

Can you explain this further please?

alanocallaghan:
Checking the java.lang source, this is what Process.inputReader() does anyway

yvesuza:
Hi Pete,

I downloaded the sourcecode and tried to build in IntelliJ IDEA w/ Amazon Corretto JDK. I got stuck on a RTE in GitCommitPlugin.java and fixed it.

Admittedly, I had ChatGPT assist me

On a bigger scale: I’m investigating if we could build a plugin that directly retrieves WSI data from the Philips PIPS tile server (since QuPath doesn’t support the isyntax format).

Kind regards,

Yves

Van: Pete ***@***.***>
Verzonden: woensdag 5 februari 2025 16:04
Aan: qupath/qupath ***@***.***>
CC: Sucaet, Yves ***@***.***>; Author ***@***.***>
Onderwerp: Re: [qupath/qupath] Fix ProcessBuilder bug in GitCommitPlugin.java (PR #1764)


incorrect usage of Process.inputReader()

Can you explain this further please?

petebankhead:
Thanks, `Process.inputReader()` was added in [Java 17](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Process.html#inputReader()). I remember considering at the time whether we needed to make the build scripts compatible with earlier Java versions, specifically because of `inputReader()`, and decided that it was preferable to use the simpler code - especially since QuPath v0.6 will requires Java 21 to run anyway.

Because the Java version used to run Gradle doesn't have to match the Java version used to build QuPath, the use of `inputReader()` means that you'll need to run Gradle with at least Java 17. But it makes sense to run it with Java 21, so that Gradle won't need to download a different Java version for QuPath itself.

Issue #1773, opened 2025-02-25
Ctrl + S shortcut not working


finglis:
## IMPORTANT!

* [x] I've definitely found a bug (it you're not sure, please use [image.sc](https://forum.image.sc/tags/qupath) instead)
* [x] I've checked https://qupath.github.io for a new release that might already have fixed the issue
* [x] I've checked the [Changelog](https://github.com/qupath/qupath/blob/master/CHANGELOG.md) to see if the bug has already been fixed in the next release
* [x] I've checked for existing GitHub issues describing the same problem

### Describe the bug
Ctrl + s is not saving on QuPath for windows (or when the system menubar is not being used on Mac)

### To Reproduce
1. Open an image (within a project or not)(previously edited/saved or newly imported)
2. Make an edit (eg set stain or create annotation)
3. Ctrl (cmd) + s
4. Either open another image, close the QuPath window or press ctrl + r to restore previous saved state. These will all alert you that unsaved changes exist. 

### Expected behavior
Ctrl + s should save the image and no pop up should be prompted when navigating away from the image. 

### Desktop (please complete the following information):
 - OS: Windows, macOS when system menubar is not used. 
 - QuPath Version: 0.6.0-rc1 onwards

### Additional context
When pressing ctrl + S the toolbar buttons change to the selection variety (dotted lines), as expected when "s" is pressed, but not with "ctrl" beforehand. 

Reported by Mike N

PR #1774, opened 2025-02-25
Fix saving accelerator bug

petebankhead:
Fix v0.6.0-rc1 regression whereby the Ctrl+S accelerator wouldn't work on Windows, or Mac with non-system-menubar. Might also have been broken on Linux, who can say?

Edit: Hadn't yet seen bug report, fixes https://github.com/qupath/qupath/issues/1773


alanocallaghan:
I think it was broken in Linux yeah but I always assumed PEBCAK on my part so didn't check

finglis (approved):

PR #1775, opened 2025-02-26
Resolve single measurement range bug (I think)

alanocallaghan:


finglis:
Steps to reproduce the issue:
1. Open an image and create an annotation containing detections (cell detection is fine) in a fresh qupath instance (important)
2. Open "create single measurement classifier", note the default state for "channel filter", "measurement", "threshold" and "classifier name".  
image

3. Select a measurement (eg. area) to fill the histogram and note the range should now be adapted to the measurement
image

4. Change the channel from all to a specific one (eg Hematoxylin) and the measurement will change to a related parameter but the histogram or range wont change to these new values (issue).  
image

5. Change the measurement to something else and histogram and range will adapt to new values as expected 

6. Change the channel back to "no filter (all channels)" and measurement to nucleus area, the range and histogram will change as expected but the classifier name will remain as the channel specific option.
 
image

7. Close and re-open the single measurement classifier and note the range and histogram are between 0-1 (classifier name also remains)
image
image

finglis:
Since PR, a measurement seems to be selected by default at the start resulting in the histogram and range being filled so skips to step 3. 
Step 4 however (changing the channel filter) doesn't result in an update of histogram and range to the new values. 
image
step 5 sorts this by the measurement change prompting the update. However changing back to no channel filter (all channels) also doesn't prompt update

alanocallaghan:
> However changing back to no channel filter (all channels) also doesn't prompt update

I wouldn't expect removing the channel filter to ever change the selected measurement?

alanocallaghan:
> Step 4 however (changing the channel filter) doesn't result in an update of histogram and range to the new values.

I can reproduce this bug (the other behaviour I consider fine):

1. Open an image with detections
2. Open single measurement classifier window
3. Change channel filter
4. Observe selected measurement changes (eg, in a H&E image with cells, you get "Nucleus: hematoxylin OD mean")
5. Revert channel filter
6. Observe name of selected measurement changes (eg, in a H&E image with cells, you get "Cell: hematoxylin OD mean"), but value and histogram do not update.
7. Repeat steps 3-6 as many times as you find amusing

The measurement in step 4 is 7th in the unfiltered measurement list, and the measurement in step 6 is 7th in the filtered list; I suspect this is related.

finglis (approved):

alanocallaghan:
Noting this: https://docs.oracle.com/javase/8/javafx/api/javafx/scene/control/ComboBox.html

We need to be careful with ComboBox as the selectedItemProperty of the selectionModel and the `valueProperty` are not necessarily the same thing. In fact it seems that you can have a set `valueProperty` with no items!

PR #1776, opened 2025-02-26
Use supplier in script menu loader

Rylern:
There's an issue with the script editor. Basically, if I:

* launch QuPath
* open a project
* use Automate > Project scripts... > Select something.groovy

then i get the default script editor, without syntax highlighting.

This happens because `ScriptMenuLoader` stores the `ScriptEditor` before it was set to be a `RichScriptEditor`. This PR fixes that by ensuring `ScriptMenuLoader` always gets the current `ScriptEditor`.


petebankhead:
Great, thanks!

PR #1777, opened 2025-02-26
Much faster creation & rendering of complex ROIs

petebankhead:
Major performance improvements when creating complex ROIs (e.g. from a pixel classifier) and rendering them in the viewer.

## ContourTracer improvements
Building on https://github.com/qupath/qupath/pull/1520, generating complex annotations can now be **much** faster.

For example, I applied a moderately high-resolution thresholder to `OS-2.ndpi` to generate a monster multipolygon with millions of vertices.

image

On the same MacBook Pro:

| Version | Time |
|---|---|
| v0.5.1 | ~494 seconds |
| v0.6.0-rc3 | ~133 seconds |
| With this PR | ~15 seconds |

The key step is to avoid creating polygons for each individual tile and unioning them at the end, because unioning is too expensive.

Instead, we trace all the lines across all the tiles and then pass them all in one big go to a `Polygonizer` at the end.

Some extra tricks and optimizations are used to speed this up, and also to reduce the number of vertices in the final result, without changing the shape.

The annotation above had about 7-8 million coordinates, while with this PR it is just under 5 million.

> This was inspired by CoverageUnion not quite working... it was very fast, but gave invalid results.

> Note that trying to generate regions at an even *higher* resolution is still unreasonably slow - so it's still not advised to apply pixel classification / thresholding to a whole slide image at full (or almost full) resolution: restricting to regions of interest can still be necessary.

## Faster rendering

Generating huge ROIs is one thing, rendering them is another.

This PR also makes this much more efficient using 2 tricks... and then a 3rd trick that renders one of the first two moot.

The tricks are:
1. Speed up `ShapeSimplifier` by replacing Java's built in `PriorityQueue` with something smaller & faster to handle removals and insertions
2. Apply shape simplification progressively & across more resolutions when zooming out in the viewer (rather than simplifying the original shape at a few downsamples only)
3. Skip `ShapeSimplifier` entirely in the viewer, and instead introduce `DownsampledShapeCache`. This uses a much simpler method of shape simplification, while also discarding very small fragments or holes.

The end result looks a bit different. I don't think it's worse, but the rendering might benefit from some tweaks in the future.

More critically, it is *far* more responsive. Previously, *rendering* the ROI might take up to a minute and it would be sluggish when zooming in & out until the simplified shapes were built. Viewing the image at a low resolution was slower than at a high resolution.

Now viewing at a low resolution should be fast, and the lag when zooming should be much less. Images with complex ROIs should also open much faster.

### DIFF
--- a/qupath-core/src/main/java/qupath/lib/analysis/images/ContourTracing.java
+++ b/qupath-core/src/main/java/qupath/lib/analysis/images/ContourTracing.java
@@ -28,21 +28,15 @@ import java.awt.image.Raster;
 import java.awt.image.WritableRaster;
 import java.io.IOException;
 import java.nio.file.Path;
-import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.Comparator;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
-import java.util.Set;
 import java.util.TreeMap;
-import java.util.TreeSet;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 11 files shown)


qupath-core/src/main/java/qupath/lib/analysis/images/ContourTracing.java
```diff
@@ -742,29 +734,69 @@ private static Geometry createTracedGeometry(Raster raster, double minThresholdI
...
+			long endTime = System.currentTimeMillis();
+			logger.debug("Creating lines from pair time: {} ms", endTime - startTime);
+
+			var polygonizer = new Polygonizer(true);
+			polygonizer.add(lineStrings);
+			var geometry = polygonizer.getGeometry();
+			geometry.normalize();
+			return geometry;
+		} catch (Throwable e) {
+			System.err.println("Error in polygonization: " + e.getMessage());
```
Rylern, line 797: `logger` instead of `System.err.println`, with the entire exception (not just the message)

qupath-core/src/main/java/qupath/lib/analysis/images/ContourTracingUtils.java
```diff
@@ -0,0 +1,368 @@
...
+
+    private static final Logger logger = LoggerFactory.getLogger(ContourTracingUtils.class);
+
+    /**
+     * Create a new collection that contains only pairs that did not have any duplicate in the input.
+     * This is different from removing duplicates: if a pair is duplicated, all instances of it are removed.
+     * @param lines
+     * @return
+     */
+    static Collection<CoordinatePair> removeDuplicatesCompletely(Collection<CoordinatePair> lines) {
```
Rylern, line 44: These kind of functions could be unit tested

qupath-core/src/main/java/qupath/lib/analysis/images/CoordinatePair.java
```diff
@@ -0,0 +1,71 @@
+package qupath.lib.analysis.images;
+
+import org.locationtech.jts.geom.Coordinate;
+
+import java.util.Comparator;
+import java.util.Objects;
+
+class CoordinatePair {
```
Rylern, line 8: This could be a `record`, that would remove some boilerplate code

qupath-core/src/main/java/qupath/lib/analysis/images/ContourTracingUtils.java
```diff
@@ -0,0 +1,368 @@
...
+            var list = entry.getValue();
+            var queued = new ArrayDeque<>(buildLineStrings(list, nonMergeable::contains, factory, xOrigin, yOrigin, scale));
+            while (!queued.isEmpty()) {
+                var ls = queued.pop();
+                if (isClosed(ls)) {
+                    lines.add(ls);
+                    continue;
+                }
+                if (Thread.interrupted())
+                    throw new RuntimeException("Interrupted");
```
Rylern, line 205: Why not `InterruptedException`?

PR #1779, opened 2025-03-02
Event faster contour tracing (albeit only slightly this time)

petebankhead:
Addendum to https://github.com/qupath/qupath/pull/1777

Replaces the frequent use of `Coordinate` with `Point2`, since it's a more minimal class and gives us more flexibility to optimize things in the future.

These changes reduced the contour generation time in the previous example further still, from ~15 seconds to under 13 seconds (with some variation).

The use of very large `HashSet` and `HashMap` instances continues to slow things down if the number of coordinates gets much higher; I haven't been able to find a straightforward solution to that.

### DIFF
--- a/qupath-core/src/main/java/qupath/lib/analysis/images/ContourTracingUtils.java
+++ b/qupath-core/src/main/java/qupath/lib/analysis/images/ContourTracingUtils.java
@@ -8,12 +8,12 @@ import org.locationtech.jts.geom.LineString;
 import org.locationtech.jts.simplify.DouglasPeuckerSimplifier;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import qupath.lib.geom.Point2;
 
 import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


qupath-core/src/main/java/qupath/lib/analysis/images/ContourTracingUtils.java
```diff
@@ -135,6 +137,10 @@ private static Set<Coordinate> findNonMergeableCoordinates(Collection<Coordinate
                 count = 1;
             }
         }
+
+        nonMergeable.rebuild();
+//        long endTime = System.currentTimeMillis();
+//        System.err.println("Time to find non-mergeable coordinates: " + (endTime - startTime) + " ms");
```
Rylern, line 143: You could remove the commented out code or keep the message with `logger.trace`

qupath-core/src/main/java/qupath/lib/analysis/images/ContourTracingUtils.java
```diff
@@ -169,8 +175,9 @@ else if (p.isVertical())
         var firstDirection = horizontal.size() <= vertical.size() ? horizontal : vertical;
         var secondDirection = firstDirection == horizontal ? vertical : horizontal;
 
-        Collection<List<Coordinate>> lines = new ArrayList<>();
-        var mergeable = new MinimalCoordinateMap<List<Coordinate>>();
+        Collection<List<Point2>> lines = new ArrayList<>(pairs.size()/10);
+//        Map<Point2, List<Point2>> mergeable = HashMap.newHashMap(pairs.size());
+        var mergeable = new MinimalCoordinateMap<List<Point2>>(Math.max(100, (int)Math.sqrt(pairList.size())));
```
Rylern, line 180: Commented out code

Issue #1780, opened 2025-03-02
Optimise ContourTracing even more


petebankhead:
### Is your feature request related to a problem? Please describe.
I've spent some time improving the performance of `ContourTracing` for v0.6.0.
One of the main PRs is https://github.com/qupath/qupath/pull/1777

The processing time for a multipolygon with ~5 million vertices decreased on my computer from ~500 seconds to < 15 seconds... so quite a lot.

Still, trying to trace contours on an *even larger* polygon (e.g. > 30 million vertices) is not possible. At least, I always force quit before it could finish.

This is likely to be limited by JTS's `Polygonizer`, which we are pushing rather hard... although `HashSet` and `HashMap` also aren't performing well with getting the data to `Polygonizer`. So it's hard to really test its limits.

### Describe the solution you'd like
I'm writing this while the recent changes are fresh in my mind, as I think there is a 'natural' way to improve the performance further still... possibly.

The key idea is that we could store the coordinates more efficiently during processing.

Currently, we trace contours in image tiles, which may be downsampled. These are then translated and rescaled immediately after the tile is processed, with the vertices stored in `Point2` objects, with a `double` or `x` and `y`.

A somewhat surprising bottleneck is that, when we use a huge number of `Point2` as a key in a `HashMap` or value in a `HashSet` then VisualVM shows that `add`, `put`, `contains` and `remove` can be *sloooow*.

It actually sped things up considerably to sometimes use a horrible `Map<Double, Map<Double, T>>` instead of `Map<Point2, T>`... despite all the boxing/unboxing overhead.

The idea is that contour tracing always works with integer vertices, so we *should* be storing `x` and `y` as `int`, and then doing any rescaling at the very last step.

If we do that, then we could define a special class that stores `x` and `y` packed into a single `long`... or even just use `Long` directly.

That should at least cut down memory requirements a bit, but hopefully it would also lead to ways to  speed up the use of some collections.

### Describe alternatives you've considered
We could just... not do this. It isn't *guaranteed* to be faster, and ~5 million vertices in 15 seconds seems respectable.

### Additional context
So far, I've concentrated on creating just *one* ROI. `ContourTracing` will attempt to parallelize things e.g. when using multiple outputs with a pixel classifier, but only with a regular parallel stream. This may have been part of where my 30 million vertices were going wrong... the memory requirements get much bigger, and we really need smarter parallelization.

[...]

PR #1781, opened 2025-03-03
More contours

petebankhead:
Partially addresses https://github.com/qupath/qupath/issues/1780 by replacing `Point2` with `IntPoint2`, which stores its values in a packed `long`.

This doesn't make much difference to performance for the ~5 million vertices tracing, but it *did* make it possible to trace much larger regions - making `Polygonizer` the bottleneck.

For example, I could trace two regions - with 39,230,067 and 21,436,882 respectively, in 532 seconds.

Rendering was then slow again, and I couldn't *save* them because of memory issues... but I think it indicates that this PR improves the efficiency of some parts of the tracing.

### DIFF
--- a/qupath-core/src/main/java/qupath/lib/analysis/images/ContourTracing.java
+++ b/qupath-core/src/main/java/qupath/lib/analysis/images/ContourTracing.java
@@ -798,7 +798,7 @@ public class ContourTracing {
 			logger.debug("Created {} with {} coordinates", geometry.getGeometryType(), geometry.getNumPoints());
 			return geometry;
 		} catch (Throwable e) {
-			logger.error("Error in polygonization: {}", e.getMessage());
+			logger.error("Error in polygonization: {}", e.getMessage(), e);
 			return factory.createEmpty(2);
 		}
 	}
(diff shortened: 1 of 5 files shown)


qupath-core/src/main/java/qupath/lib/analysis/images/IntPoint.java
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    public int getY() {
+        return unpackY(value);
+    }
+
+    public long getLongValue() {
+        return value;
+    }
+
```
alanocallaghan, line 37: I think this is provided by implementation as a record? or maybe `value()`

petebankhead:
I didn't change a lot, but I did switch to use `record` and fixed a comparator.

Running the same script again generated the huge geometries in 268 seconds.

So... good?

alanocallaghan:
I assume the packed long will be what made the biggest difference?

qupath-core/src/main/java/qupath/lib/analysis/images/ContourTracing.java
```diff
@@ -804,7 +798,7 @@ private static Geometry createGeometry(GeometryFactory factory, Collection<Coord
 			logger.debug("Created {} with {} coordinates", geometry.getGeometryType(), geometry.getNumPoints());
 			return geometry;
 		} catch (Throwable e) {
-			System.err.println("Error in polygonization: " + e.getMessage());
+			logger.error("Error in polygonization: {}", e.getMessage());
```
Rylern, line 801: Is there a reason not to use `logger.error("Error in polygonization", e);`? The stack trace might be important to keep if an error occurs

petebankhead: Fixed now, thanks

PR #1783, opened 2025-03-04
Create directories more easily in scripts

petebankhead:
New `createDirectoryInProject` and `createDirectory` methods that return a path *and* call `mkdirs` if no directory already exists for the given path.

Example use:
```groovy
var dir = createDirectoryInProject("Hello", "there", "hello2.txt")
new File(dir, "hello3.txt").text = "Hi!"
```
This replaces `makePathInProject`, which frankly I forgot exists... and which is less clear (is it making a file or directory path?).

It aims to replace the annoying convention found in many QuPath scripts:
```groovy
var outputDir = buildPathInProject("results")
mkdirs(results)
var whatIWanted = buildFilePath(outputDir, "my-results.txt")
```
with
```groovy
var outputDir = createDirectoryInProject("results")
var whatIWanted = buildFilePath(outputDir, "my-results.txt")
```
or even
```groovy
var whatIWanted = createDirectoryInProject("results") + "my-results.txt"
```
because the `create` method ensures that the path ends with a separator.

### DIFF
--- a/qupath-core-processing/src/main/java/qupath/lib/scripting/QP.java
+++ b/qupath-core-processing/src/main/java/qupath/lib/scripting/QP.java
@@ -751,14 +751,13 @@ public class QP {
 	 */
 	public static String buildFilePath(String first, String... more) throws IllegalArgumentException {
 		File file = new File(resolvePath(first));
-		for (int i = 0; i < more.length; i++) {
-			var part = more[i];
-			if (part == null)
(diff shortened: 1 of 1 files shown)


Rylern (approved):
LGTM. Some unit tests could be implemented to test the new functions

alanocallaghan:
Sounds sensible I think

petebankhead:
Thanks, I've just merged this. @Rylern tests would be very welcome, if you have time to add any please do :)

PR #1789, opened 2025-03-07
Minor fixes

petebankhead:
* Fix regression that caused an exception when 2 or more classes were removed
* Fix CSS issue with icons in dark mode
* Add warning when attempting to load & apply an object classifier to an image with no compatible objects


finglis:
"Negative" seems to be missing from the default classes now? Unsure if this is user specific though

PR #1797, opened 2025-03-13
Implement double click to select z/t slice

alanocallaghan:
Enables double-clicking on Z and T sliders to set the position.

Does zero validation of inputs, so negative numbers or non-numbers will probably throw an uncaught exception.

I had to change the handling of backspace because otherwise it would (try to) delete objects if you make a mistake and retype

Incidentally, I think I like showing ticks and tick values on these sliders, but I'm possibly alone there...


MichaelSNelson:
Ticks would be nice, though not sure how the labeling would work, maybe top and bottom with 3 middle ticks unlabeled? I have worked with 10 Z stacks before and am now working with 700, so it would definitely need to be adaptive and not look too weird with at least 3 to 4 digits.

alanocallaghan:
Closing in favour of spinner pr, which hopefully is a bit nicer for lots of time points/slices too

PR #1799, opened 2025-03-13
Implement spinners for Z and T selection in viewer

alanocallaghan:
After discussion in #1797, implement spinners instead of sliders. Can't enter text as that's bound to channel toggling.

Screenshot from 2025-03-13 15-03-28

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -124,7 +124,7 @@ They may change or be removed in future versions.
   * Extension writers can use either Groovy or Kotlin - but may need to make some changes to their build scripts
 
 ### Dependency updates
-* Bio-Formats 8.1.0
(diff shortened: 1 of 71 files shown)


alanocallaghan:
Less bad image

petebankhead:
Ready for review, I think...

* Uses a translucent black background to make more visible in both dark and light mode
* Supports typing the input, clicking the spinner buttons or clicking/dragging on the slider (really a progress bar)

new-z-t-sliders

alanocallaghan:
Seems like a massive improvement, being

- more visible
- more explicit (showing what each slider does, current value and range)
- easier to use
- more accessible (better contrast)
- generally prettier (subjectively)

PR #1800, opened 2025-03-14
Enable running jpackage with extensions

alanocallaghan:
Adds an optional parameter for the jpackage build, which takes a comma-separated list of extensions that should be included in the build. These *must* be under the qupath org, or the build will fail.

Then, it clones qupath to ./qupath and each extension to ./qupath-extension-whatever, and adds each extension to `[includeBuild]` and `[dependencies]` of `include-extra`. Then it runs the gradle build as usual.

Example, setting the input to `qupath-extension-instanseg,qupath-extension-djl`:
jpackage

It's an optional arg, so in case it's not set then the build will just run as before with no extra extensions.


alanocallaghan:
Note to whoever merges: please squash merge so we don't get these beautifully named broken commits in the history...

PR #1803, opened 2025-03-14
Don't build histogram from large image stacks

petebankhead:
QuPath builds a histogram when opening an image, and this is used for Brightness/Contrast settings. This is usually based on a low resolution image and works fine. But it *doesn't* work fine for large z-stacks or time series, especially if these are at a single resolution.

It also isn't fine when opening remote images, e.g. stored through OMERO.

This PR uses a *very* basic approach to select only the middle slice of a z-stack in the event that reading the entire thing would require more than 50 MB to open. This makes it fast to open some images that otherwise throw out-of-memory errors.

For discussion...

### DIFF
--- a/qupath-gui-fx/src/main/java/qupath/lib/display/ImageDisplay.java
+++ b/qupath-gui-fx/src/main/java/qupath/lib/display/ImageDisplay.java
@@ -882,8 +882,14 @@ public class ImageDisplay extends AbstractImageRenderer {
 		int nImages = server.nTimepoints() * server.nZSlices();
 		// Try to get the first image 'normally', so that if there is an exception it can be handled
 		var list = new ArrayList<BufferedImage>();
-		list.add(server.getDefaultThumbnail(0, 0));
-		if (nImages > 1) {
+		list.add(server.getDefaultThumbnail(server.nZSlices()/2, 0));
+		// Ideally, we'd like to build the histogram from all pixels -
+		// but we can't if our image is too big... and shouldn't if it's slow to access (e.g. via OMERO)
+		// So this is a *very* rough way to decide when to accept just a single plane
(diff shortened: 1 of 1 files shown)


petebankhead:
The PR has grown in complexity, but hopefully provides a more satisfactory solution (even if it could be further improved).

The main changes:
* When opening a multidimensional image (z-slices and/or time-points), only the 'middle' slice is initially used to generate a histogram to determine brightness/contrast settings. This can make opening very large images much faster.
* Later, *all cached slices at the lowest resolution* are used when the histogram is requested from `ImageDisplay` (e.g. by the Brightness/Contrast dialog
   * This means that pressing 'auto' or 'reset' for a channel can do different things depending upon how much of the image has been read (and cached). While suboptimal, it seems preferable than either
      1. Reading all the pixels (which is what caused the problems this PR aims to address)
      2. Relying on choosing a 'good' representative slice to base the histogram on (but which might be out-of-focus, and give terrible estimates for min/max values)
* Two previously separate caches (for image tiles and thumbnails) have been brought into a being a single cache for simplicity, and to make the changes above work effectively
* `ImageJServer` can better handle large multidimensional images in two ways:
  1. Use tiles & tile caching (i.e. extend `AbtractTileableImageServer` instead of `AbstractImageServer`
  2. Use a virtual stack for large images
* `ImageDisplay` has been slightly cleaned up and simplified (although it is still too complicated)

It appears to be behaving, but I wouldn't be surprised if some subtle bugs have been introduced. Since I suspect they will only emerge in general use, I'd like to merge the PR soon to give time to identify and fix them.

alanocallaghan:
The only alternative I have to a. and b. would be to read a random sample of every slice, although choosing the number of samples to take from each slice is maybe not straightforward unless you can make some assumptions about the distribution of pixels, and this can in any case be considered for a future PR

PR #1806, opened 2025-03-21
Attempt to fix obscure file deletion bug on macOS

petebankhead:
One proposed approach to fix a deadlock occurring when deleting files on macOS - but only under some circumstances.

This is really a placeholder to make sure we remember to fix it before the next release - @Rylern please feel free to fix in another, better way if you like and we can close this.


petebankhead:
Closed - use https://github.com/qupath/qupath/pull/1808 instead.

PR #1809, opened 2025-03-24
Add tags to project entries

Rylern:
Add tags to project entries (see #1289). This PR doesn't include any UI changes, just support for tags.


Rylern:
Yes, I thought about it at some point but forgot

PR #1821, opened 2025-04-07
Remove objects touching ROI boundary

petebankhead:
New methods int `PathObjectTools` to identify & remove objects that touch the boundary of a specified ROI.

This can be used to identify & discard clipped objects after detection.

Examples:

### Original image
image

### Remove all objects touching the selected object
```groovy
PathObjectTools.removeTouchingBoundary(
    getCurrentHierarchy(),
    getSelectedObject()
)
```

image

### Remove only detections touching the selected object

```groovy
PathObjectTools.removeTouchingBoundary(
    getCurrentHierarchy(),
    getSelectedObject(),
    PathObject::isDetection
)
```

image

### Remove child objects touching the selected object

```groovy
PathObjectTools.removeTouchingBoundary(
    getCurrentHierarchy(),
    getSelectedObject(),
    p -> p.getParent() == getSelectedObject()
)
```

The output in this case is the same as filtering to remove detections only.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,9 @@ Some things may be added, some things may be removed, and some things may look d
 (These are not yet ordered by interestingness)
 
 * Support system light/dark color themes
+* Commands to remove objects touching the bounds of the image or other objects (https://github.com/qupath/qupath/pull/1821)
+  * _Objects → Delete... → Delete objects on image bounds_
+  * _Objects → Delete... → Delete objects touching selected ROI bounds..._
 * Window menu to help find lost windows (https://github.com/qupath/qupath/pull/1790)
 * Improved display of annotation names (https://github.com/qupath/qupath/pull/1532)
 * Better performance for large pixel classifiers (https://github.com/qupath/qupath/pull/1782)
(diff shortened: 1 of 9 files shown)


petebankhead:
There are now two new scriptable commands:

* Objects → Delete... → Delete objects on image bounds
* Objects → Delete... → Delete objects touching selected ROI bounds...

The first runs immediately to remove all objects that touch the boundary of the current image.

The second requires input to choose the type of objects to remove (annotations or detections), and whether to only remove child objects.

I switched from 'boundary' to 'bounds' in the code because the plural was helpful to sidestep the grammatical complication that arises because there might be one or multiple selected objects.

The naming is a bit awkward, was are the method names in scripts... but I couldn't think of better ones.

Example scripts (within `QP`):
```groovy
removeObjectsTouchingImageBounds()
removeChildObjectsTouchingSelectedBounds()
removeObjectsTouchingSelectedBounds()
removeObjectsTouchingSelectedBounds(PathObjectFilter.ANNOTATIONS)
removeObjectsTouchingSelectedBounds(PathObjectFilter.DETECTIONS_ALL)
```

PR #1831, opened 2025-04-10
Offer to open the extension manager

petebankhead:
Add a prompt immediately after the welcome screen to open the extension manager.

This will appear on startup so long as
- no managed extensions are installed
- the user hasn't deselected 'Remind me next time'

The aim is to improve the discoverability of extensions, while highlighting the extension manager for new users (so that they use it instead of drag & drop installation).


alanocallaghan:
Looks great

PR #1836, opened 2025-04-11
Update github-release.yml

alanocallaghan:
Pass an input variable for extensions to the sub-job for building from source


alanocallaghan:
Example workflow run here: https://github.com/alanocallaghan/qupath/actions/runs/14400529518/job/40385078481

PR #1845, opened 2025-04-17
Export docs & extension loading

petebankhead:
Try to fix `DocGenerator` - which was skipping any commands added through extensions, even built-in extensions.

This is what's used to create the reference page in the docs, e.g. https://qupath.readthedocs.io/en/0.5/docs/reference/commands.html

This involved making a change to `ExtensionLoader`, so that it loads extensions immediately if called from the Platform thread, rather than delayed with `runLater`. This means that calling code that creates a new `QuPathGUI` instance can rely on extensions being available immediately (well, assuming that the *extension* hasn't delayed its own initialisation...).

@Rylern does this seem ok to you, or there a reason why `runLater` is essential within `ExtensionLoader`?

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -143,7 +143,7 @@ tasks.register<JavaExec>("exportDocs") {
         println("Making docs dir in ${docsDir.absolutePath}")
         docsDir.mkdirs()
     }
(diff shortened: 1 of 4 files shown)


Rylern (approved):
No, it just needs to be executed from the JavaFX Application Thread

PR #1846, opened 2025-04-22
Clarify wording on detection display menu

alanocallaghan:
I had talked myself into this being a "Me issue" but then I showed it to Fiona


alanocallaghan:
Alternative is to have two toggleButtons, one "show detections" and one "show icons", and when "show detections" is enabled, "show icons" is disabled

finglis (approved):
Certainly an improvement
Agree that it could be narrowed down to be just the two toggle options. Also when the detections are hidden and then the option to "show icons" is selected have it expand the list would be preferred but as discussed with Alan, probably better to have icon control greyed out when detections are hidden anyway.

petebankhead:
This looks fine to me, happy to merge

PR #1847, opened 2025-04-22
Show/hide checks for icons and detections independently

alanocallaghan:
Alternative to #1846


petebankhead:
I prefer https://github.com/qupath/qupath/pull/1846 - I think this introduces more complexity, and when I turn off 'with icons' part of me expects the annotation icons to disappear too (because 'detection' isn't specified in the text of the menu item I'm clicking, but rather the parent menu).

Ultimately, I think the reason to support showing without icons is because at one point there was considerable overhead in creating the ROI icons. I don't *think* that's the case now, although I haven't checked thoroughly.

We may be able to rethink the options here entirely, but I think it's obscure enough that the small change of wording is the best strategy for v0.6.0.

alanocallaghan:
Yes, this was a bodge really; I like the idea but it would require further changes

Issue #1749, opened 2025-01-18
Support OME-XML masks


petebankhead:
### Is your feature request related to a problem? Please describe.
OME-XML masks aren't supported by QuPath.

This request is based on a post on the forum: https://forum.image.sc/t/applications-that-support-ome-xml-masks/106416

### Describe the solution you'd like
Update `BioFormatsImageServer` to implement `PathObjectReader` so that OME-XML masks can be supported.

### Additional context
An example has been posted at https://zenodo.org/records/14685515

For now, this request is to explore the feasibility - and whether OME-XML masks are a good fit for QuPath objects.

PR #1848, opened 2025-04-23
Add support for OME XML shapes

Rylern:
Implement #1749.

This lacks proper testing, but I haven't be able to find another image than https://zenodo.org/records/14685515 (and even on this image I don't know what the expected annotations are).

### DIFF
--- a/qupath-extension-bioformats/src/main/java/qupath/lib/images/servers/bioformats/BioFormatsImageServer.java
+++ b/qupath-extension-bioformats/src/main/java/qupath/lib/images/servers/bioformats/BioFormatsImageServer.java
@@ -44,6 +44,8 @@ import loci.formats.ome.OMEPyramidStore;
 import loci.formats.ome.OMEXMLMetadata;
 import ome.units.UNITS;
 import ome.units.quantity.Length;
+import ome.xml.meta.OMEXMLMetadataRoot;
+import ome.xml.model.Shape;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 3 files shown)


Rylern:
Indeed, the ROIs you get are more plausible.

There's a script that shows a similar conversion [here](https://github.com/glencoesoftware/ome-omero-roitool/blob/7b3896f6f6f3c168f3610a4a416ccf46df1aa278/src/dist/QuPath.scripts/OME_XML_import.groovy#L257) (it's in the same repo than the script you mentionned). I tried to implement it but it wasn't working, as it seems to expect `mask.getWidth() * mask.getHeight() == binData.getLength().getValue()`, so your code seems better.

I'll start from your code and check that the correct `series` is used and handle transforms.

Rylern:
I integrated your code, handled transforms, and added unit tests for the converter from OME shape to QuPath ROI.

I could create a sample image by opening Fiji, opening any sample image, adding ROIs to the ROIs manager, and then `Plugins` -> `Bio-Formats` -> `Bio-Formats Exporter`. This allowed me to check that rectangles, ellipses, polygons, freehand selections, lines, and points work correctly.

About `series`, there is no function to select ROIs of a particular series (unlike on [this script](https://github.com/ome/bioformats/blob/c997de185fc71eef191a1c9d82c35990bbd042c9/components/formats-gpl/utils/PrintROIs.java#L52)), but ROIs are retrieved from a `IFormatReader` and it's possible to set the `series` of a `IFormatReader`, so I assumed it was enough. I didn't find a way to test it though.

PR #1849, opened 2025-04-23
Use viewer gamma in RenderedImageServer

### LINKED ISSUE #1736: Rendered RGB export does not use gamma setting
### Describe the bug
See https://forum.image.sc/t/small-bug-rendered-rgb-export-does-not-use-gamma-setting/106202.

### To Reproduce
1. Open any image.
2. Change `Viewer gamma` in `Brightness and contrast` to something else than 1.
3. `File` > `Export Images` > `Rendered RGB (with overlay)`: the exported image has still a gamma of 1.

### Expected behavior
The Rendered RGB image should take the viewer gamma into account.

### Desktop:
 - OS: macOS but very likely to be any OS.
 - QuPath Version: 0.5.1 and 0.6.0-rc3

Rylern:
Fix #1736 by making `RenderedImageServer` use gamma of QuPath viewer

### DIFF
--- a/qupath-gui-fx/src/main/java/qupath/lib/gui/images/servers/RenderedImageServer.java
+++ b/qupath-gui-fx/src/main/java/qupath/lib/gui/images/servers/RenderedImageServer.java
@@ -24,6 +24,7 @@ package qupath.lib.gui.images.servers;
 import java.awt.AlphaComposite;
 import java.awt.Color;
 import java.awt.image.BufferedImage;
+import java.awt.image.LookupOp;
 import java.io.IOException;
 import java.net.URI;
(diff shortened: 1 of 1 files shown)


petebankhead:
Interesting! I didn't imagine it would be so simple... I also didn't imagine `getGammaOp()` would be a public method in the viewer. I'm not sure it *should* be a public method, but it is, so this looks like a good fast solution for now.

Minimal example script to document use:
```groovy
def path = buildPathInProject("rendered.png")
writeRenderedImage(getCurrentViewer(), path)
```
I confirm it applies the gamma to the image, while the overlay is unchanged.

Issue #1854, opened 2025-04-30
Screenshots created in QuPath can also include the screenshot menu


petebankhead:
### Describe the bug
Screenshots created in QuPath can also include the screenshot menu

### To Reproduce
1. Go to *Edit → Copy to clipboard... → Main window screenshot* (or *File → Export snapshot... → Main window screenshot*)
2. Check if the menu is visible

### Expected behavior
Screenshots *don't* include the still-open menu.

### Screenshots
![Image](https://github.com/user-attachments/assets/e5277e47-9935-4f79-8514-d406903c30a1)

### Desktop (please complete the following information):
 - OS: macOS *but only if the system menubar is turned off*
 - QuPath Version: v0.5.1 (probably earlier)

### Additional context
I haven't checked if this happens on other operating systems.

It does *not* happen on macOS by default, because the system menubar is turned on. However, when turned off the problem occurs.

I'd *expect* the issue to occur for Windows and Linux... but no one has complained, so perhaps it doesn't.

Issue #1309, opened 2023-09-02
Screenshots made using JavaFX have slightly different colors on macOS


petebankhead:
## Bug report

**Describe the bug**
Screenshots make using QuPath v0.4.x have unexpected colors - at least on macOS. This affects
* *Edit &rarr; Copy to clipboard... &rarr; Main window screenshot*
* *Edit &rarr; Copy to clipboard... &rarr; Full screenshot*
* *File &rarr; Export snapshot... &rarr; Main window screenshot*

Snapshots of *specific components* (e.g. the viewer) seem to be fine - suggesting the problem relates to the use of the JavaFX `Robot` class for screen capture, and not color management with JavaFX more generally.

**To Reproduce**
Using one of the above commands can show the issue, but the color shift can be subtle enough to miss.

I've written a script that reveals the issue more clearly, by comparing screenshots made from JavaFX with those made using (the older) Java AWT:
https://gist.github.com/petebankhead/6d8ea33e0137705d33abc1800fd52072

The colors generated using Java AWT's `Robot` class seem correct, whereas those generated with JavaFX's `Robot` class are unexpectedly 'vivid'.

**Expected behavior**
Screenshots generated with QuPath should have colors that faithfully match what is seen on screen, and which would be generated using the operating system's screen capture.

**Screenshots**
Screenshots from running the above script:

### With JavaFX
![jfx](https://github.com/qupath/qupath/assets/4690904/2662a20d-138c-41ec-896d-3d0dd283b2df)

### With AWT
![awt](https://github.com/qupath/qupath/assets/4690904/0756554e-c120-4074-8970-6469de5fdde6)

###  Difference
![diff](https://github.com/qupath/qupath/assets/4690904/1df88b3c-03ae-4620-a0ba-fe2907ac11ec)

**Desktop (please complete the following information):**
 - OS: macOS - with Apple Studio Display
 - QuPath Version: v0.4.x

**Additional context**
I have only seen this on macOS with an Apple Studio Display - we need to check if affects other platforms as well.

(Additionally, based on the current code I'd expect there are issues with multi-screen scenarios... but I usually only have one)

petebankhead:
I have replicate the issue on:
* MacBook Pro (M1, macOS Ventura 13.5.1)
* MacBook Pro (Intel, macOS Big Sur 11.7.9)

I could **not** replicate it on a Windows 10 laptop.

Therefore it seems to affect all macOS releases, but probably not other platforms (although Linux hasn't been checked yet).

PR #1855, opened 2025-04-30
Fix main window snapshots

petebankhead:
Fixes https://github.com/qupath/qupath/issues/1854


Rylern (commented):
I only have my Linux system today, and the issue #1854 doesn't happen on it.

petebankhead:
Interesting, thanks - maybe it's specific to Mac. We do have to make the screenshots differently on Mac, using AWT rather than JavaFX, because of https://github.com/qupath/qupath/issues/1309

I'll still merge the PR though with the assumption that it shouldn't have negative consequences for other platforms.

PR #1856, opened 2025-05-01
Add dock button to keep description pane open

Rylern:
This PR aims at fixing [this issue](https://forum.image.sc/t/improved-descriptions-for-project-images/111854).

It is done by adding a button to keep the description pane of an image entry open:
image

If this button is on, then the description pane will stay visible, even if the selected image entry doesn't have any description.
A preference was added to save the button status when QuPath restarts.

An edit button was also added to easily edit the description. It is only shown when an image is currently selected.

The caret in the description pane is not visible anymore.

### DIFF
--- a/qupath-gui-fx/src/main/java/qupath/lib/gui/panes/ProjectBrowser.java
+++ b/qupath-gui-fx/src/main/java/qupath/lib/gui/panes/ProjectBrowser.java
@@ -242,6 +242,7 @@ public class ProjectBrowser implements ChangeListener<ImageData<BufferedImage>> 
 		TextArea textDescription = new TextArea();
 		textDescription.textProperty().bind(descriptionText);
 		textDescription.setWrapText(true);
+		textDescription.setEditable(false);
 
 		ToggleButton keepDescriptionOpenButton = new ToggleButton(
 				null,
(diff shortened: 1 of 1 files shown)


petebankhead:
The PR looks good. Two very minor things:
* The height of the 'Description' title is larger than the height of comparable titles in other panes (e.g. 'Associated images' under the 'Image' tab, or 'Annotation list' and 'Class list' under the 'Annotations' tab). I think these should be consistent - and preferably the smaller height.
* When pressing the 'Edit' button, I think that the `TextArea` in the dialog should automatically receive the focus and/or also contain some prompt text, e.g.

```java
	static boolean showDescriptionEditor(ProjectImageEntry<?> entry) {
		TextArea editor = new TextArea();
		editor.setWrapText(true);
		editor.setPromptText("Enter description for " + entry.getImageName());
		editor.setText(entry.getDescription());
		Dialog<ButtonType> dialog = new Dialog<>();
		dialog.getDialogPane().getButtonTypes().setAll(ButtonType.OK, ButtonType.CANCEL);
		dialog.setTitle("Image description");
		dialog.getDialogPane().setHeaderText(entry.getImageName());
		dialog.getDialogPane().setContent(editor);
		Platform.runLater(editor::requestFocus);
		Optional<ButtonType> result = dialog.showAndWait();
		if (result.isPresent() && result.get() == ButtonType.OK && editor.getText() != null) {	
			var text = editor.getText();
			entry.setDescription(text.isEmpty() ? null : text);
			return true;
		}
		return false;
	}
```
I suggest and/or because receiving the focus means the prompt text isn't visible anyway... so I don't have a strong opinion on which is preferable.

Rylern:
Both comments were addressed.

PR #1858, opened 2025-05-02
Use the same capitalisation pattern for ImageJ script menu item and stage

alanocallaghan:
Previously one was sentence case and the other title

### DIFF
--- a/qupath-extension-processing/src/main/resources/qupath/imagej/gui/scripts/strings.properties
+++ b/qupath-extension-processing/src/main/resources/qupath/imagej/gui/scripts/strings.properties
@@ -1,4 +1,4 @@
-title = ImageJ Script Runner
+title = ImageJ script runner
 
 # Main UI
 ui.title.script = Script


petebankhead:
Looks fine, will merge.

FWIW I’m less concerned that the capitalization matches the menu item, and more that it’s consistent across title bars. Here, it seems that title vs sentence case is a matter of debate.

Interesting link about it at https://www.everyinteraction.com/articles/title-case-vs-sentence-case-in-ui/

I think I’m on Team Sentence Case, with all the consistency that that statement implies. Probably something we should discuss & have a position on one day.

alanocallaghan:
I generally don't care but if we're writing code that selects menu items and windows by text I want to minimise the gotchas.

I Can't Say I'm a Fan Of Title Case, not least because it's often unclear whether to capitalise prepositions and articles, and it seems Unnecessarily Grand. But whichever we use id prefer consistency

Issue #1867, opened 2025-05-12
gradlew build failing on pc with Locale using commas


francisduvivier:
### Description
Build is failing on pc with a Belgian locale because some unit test depend on the locale using dots in numbers and not commas.

### To Reproduce
1. Set the locale on your pc to en_BE
2. run ./gradlew clean build

### Expected behavior
No tests failing

### Desktop (please complete the following information):
 - OS: MacOS
 - QuPath Version: 1d3f9980804923dfc9e207cbfad2a7f1ff545c37

### Additional context
One of the 2 unit test files failing is TestParameterList.java, in this case the test uses `GeneralTools.parseArgStringValues(json)`, which coverts the values in the json to string values without taking the locale into account. 
This then causes a problem because a bit later in the test, `ParameterList.updateParameterList(paramsFromJson, mapFromJson, Locale.getDefault())` is used, which is using the locale of the host machine.

francisduvivier:
Issue is addressed in https://github.com/qupath/qupath/pull/1868

PR #1868, opened 2025-05-12
Fix locale issues in Unit tests

francisduvivier:
This fixes the build in case of a build environment with a Belgian Locale.
This fixes issue https://github.com/qupath/qupath/issues/1867


francisduvivier:
Looks like there are some more more tests affected by locale issues, so I put the PR in to draft mode to fix these first.

alanocallaghan:
Thanks for raising this! We opted to instead test the formatting on different locales in #1870

PR #1869, opened 2025-05-13
Fix extension-loading freeze

petebankhead:
Try to fix a bug whereby attempting to install an extension causes a freeze - even though the extension is fine with relaunching QuPath.

It appears to happen when `FXUtils.runOnApplicationThread` was called from the `catalogs-directory-watcher` thread... although I'm not entirely sure why.


Rylern:
I think I know why the deadlock occurs.

There are two threads running in parallel: `catalogs-directory-watcher` and `JavaFX Application Thread`.

Here is a simplified version of what is happening when installing Instanseg for example:

```java
// (runs on the catalogs-directory-watcher thread)

synchronized (qupath.ext.extensionmanager.core.ExtensionClassLoader) {
  CountDownLatch latch = new CountDownLatch(1);
  Platform.runLater(() -> {
     // Runs on the JavaFX Application Thread
     InstansegExtension.loadExtension();
     latch.countDown();
  });
  latch.await(); // Waits for the JavaFX thread to finish
}
```

To recap:

* `catalogs-directory-watcher` acquires a lock on `qupath.ext.extensionmanager.core.ExtensionClassLoader`.
* `catalogs-directory-watcher` asks to call `InstansegExtension.loadExtension()` on the `JavaFX Application Thread`.
* `catalogs-directory-watcher` waits for `InstansegExtension.loadExtension()` to complete.
* `JavaFX Application Thread` calls `InstansegExtension.loadExtension()`.
* `InstansegExtension.loadExtension()` is located in a JAR loaded by `qupath.ext.extensionmanager.core.ExtensionClassLoader`. Therefore, before calling `InstansegExtension.loadExtension()`, the `JavaFX Application Thread` has to get a lock on  `qupath.ext.extensionmanager.core.ExtensionClassLoader` (this is the part I'm not 100% sure of but it seems sensible).
* `JavaFX Application Thread` wants the lock on `qupath.ext.extensionmanager.core.ExtensionClassLoader`, but `catalogs-directory-watcher` has already it: `JavaFX Application Thread` is `BLOCKED`.

You PR fixes that by removing the **catalogs-directory-watcher` waits for `InstansegExtension.loadExtension()` to complete.** line. That works, but that doesn't address the real issue, which happens on [this function](https://github.com/qupath/extension-manager/blob/a1e8f7f04853af3b3d61f7e330655fffca8d042d/extensionmanager/src/main/java/qupath/ext/extensionmanager/core/ExtensionClassLoader.java#L48):

```java
public synchronized void addJar(Path jarPath) throws MalformedURLException {
        addURL(jarPath.toUri().toURL());
        logger.debug("File {} loaded by extension class loader", jarPath);

        String filename = jarPath.getFileName().toString();
        if (filenamesAdded.contains(filename)) {
            logger.warn(
                    "A JAR file with the same file name ({}) was already added to this class loader. {} will probably not be loaded",
                    filename,
                    jarPath
            );
        }
        filenamesAdded.add(filename);

        for (Runnable runnable: runnables) {
            try {
                runnable.run();
            } catch (Exception e) {
                logger.error("Error when calling runnable of class loader", e);
            }
        }
    }
```

Here, we call an unknown code (`runnable.run()`) inside a `synchronized` function. This is bad practise:

> In a synchronized block, avoid calling an alien method (method whose behavior is not fully specified by the current class, like methods from other classes or overrideable methods) because such methods can acquire other locks.

(from the concurrency in Java presentation I did last year).

So to recap, the real issue comes from my code in the extension manager, so I suggest accepting the [PR I just created](https://github.com/qupath/extension-manager/pull/16) and closing this PR.

petebankhead:
Interesting - thanks for investigating all this!

Happy to close this. We'll need to bump the version of `ExtensionManager` and then update the dependency here - could you take care of that please?

PR #1871, opened 2025-05-14
Set default file of zarr export

Rylern:
Set a default empty file name when exporting a Zarr image. This fixes a bug on Windows where the default output file name was `Untitled.ome.zarr.ome.zarr.ome.zarr`.


petebankhead:
This looks good to me on Mac. Have you also confirmed it behaves on Linux?

The things I checked:
* It sets the right extension, when there is 1) no extension, 2) part of the extension
* It remembers the last directory that was used, when running the command multiple times

Rylern:
> Have you also confirmed it behaves on Linux?

Yes

PR #1879, opened 2025-05-16
Update to DJL v0.33.0

petebankhead:
@alanocallaghan any thoughts on the wisdom of this for v0.6.0?

For me it seems to work for InstanSeg without trouble - although it will require downloading the updated JNI for PyTorch (but not necessarily `libtorch`).

Release notes: https://github.com/deepjavalibrary/djl/releases/tag/v0.33.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -150,7 +150,7 @@ They may change or be removed in future versions.
 * Bio-Formats 8.1.1
 * Commonmark 0.24.0
 * ControlsFX 11.2.2
-* DeepJavaLibrary 0.31.1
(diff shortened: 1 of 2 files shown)


alanocallaghan:
Seems fine, but the release notes don't really seem overly enticing, so would be happy whether including or not

petebankhead:
Let’s go for it then - there are a few things I suspect could be useful at times (like requesting model signatures from TensorFlow, and the ability to use diff in PyTorch). Having them make slightly more doable through extensions over the coming months, without a need for a new release.

PR #1882, opened 2025-05-21
Add onTileWritten parameter in zarr writer

Rylern:
Add an `onTileWritten` function that will be called by the Zarr writer each time a tile is (successfully or unsuccessfully) written. This is used to have an idea of the progress when writing an image. An associated unit test was also added to this PR.

This PR also set the names of threads created by the Zarr writer.

### DIFF
--- a/qupath-extension-bioformats/src/main/java/qupath/lib/images/writers/ome/zarr/OMEZarrWriter.java
+++ b/qupath-extension-bioformats/src/main/java/qupath/lib/images/writers/ome/zarr/OMEZarrWriter.java
@@ -10,6 +10,7 @@ import com.bc.zarr.ZarrGroup;
 import loci.formats.gui.AWTImageTools;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import qupath.lib.common.ThreadTools;
 import qupath.lib.images.servers.ImageServer;
 import qupath.lib.images.servers.ImageServers;
 import qupath.lib.images.servers.TileRequest;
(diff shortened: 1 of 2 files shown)


petebankhead:
Could [`ThreadTools`](https://github.com/qupath/qupath/blob/96bf0ab9c0610d5f94a386cc70d157a539038d19/qupath-core/src/main/java/qupath/lib/common/ThreadTools.java#L41) be used here?

Rylern:
Yes

qupath-extension-bioformats/src/main/java/qupath/lib/images/writers/ome/zarr/OMEZarrWriter.java
```diff
@@ -179,7 +189,7 @@ public static class Builder {
         private static final String FILE_EXTENSION = ".ome.zarr";
         private final ImageServer<BufferedImage> server;
         private Compressor compressor = CompressorFactory.createDefaultCompressor();
-        private int numberOfThreads = 12;
+        private int numberOfThreads = Runtime.getRuntime().availableProcessors();
```
petebankhead, line 192: Certainly preferable to 12.... maybe switch to [`ThreadTools.getParallelism()`](https://github.com/qupath/qupath/blob/9ce42cbb44870e409ce275b3551bedd62c1f3578/qupath-core/src/main/java/qupath/lib/common/ThreadTools.java#L87)?

This exists to make it possible for code implemented in core modules to respect the number of threads requested in the QuPath UI, without being able to access the actual preference (since we may not have JavaFX).

Rylern: Yes. Ideally we should do some benchmarks to determine a good number, since writing a Zarr file implies a lot of blocking operations (reading and writing to the disk)

PR #1887, opened 2025-05-26
Increase number of decimals of stain values when creating workflow step

### LINKED ISSUE #1851: Setting stains via a script gives (very) slightly different results
### Describe the bug
Color deconvolution stain vectors set via script can be *slightly* different from those set previously, because the script representation is less precise.

An example of a logged script:
```groovy
setColorDeconvolutionStains('{"Name" : "H&E default", "Stain 1" : "Hematoxylin", "Values 1" : "0.65111 0.70119 0.29049", "Stain 2" : "Eosin", "Values 2" : "0.2159 0.8012 0.5581", "Background" : " 255 255 255"}');
```

An example of the stain stored in the image:
```groovy
println getCurrentImageData().getColorDeconvolutionStains().getStain(1).array as List
```
which results in
```groovy
[0.6511078257574492, 0.7011930431234068, 0.29049426072255424]
```

### Expected behavior
Results when setting stains via a script should be identical to those set another way.

 - OS: All
 - QuPath Version: All versions (from the beginning up to v0.6.0-rc4)

### Additional context
This emerged when writing some tests. Since the difference is extremely small, I have no evidence that it matters in practice. Still, it would be desirable to improve the accuracy - and perhaps support an alternative syntax for `setColorDeconvolutionStains` that can take a `Map` rather than JSON string.

Rylern:
An proposal to fix #1851.

It is done by increasing the number of decimals of stain values from 5 to 17 when creating a workflow step.

With this PR, the line `println getCurrentImageData().getColorDeconvolutionStains().getStain(1).array as List` and the corresponding workflow step give the same values.

### DIFF
--- a/qupath-core/src/main/java/qupath/lib/color/ColorDeconvolutionStains.java
+++ b/qupath-core/src/main/java/qupath/lib/color/ColorDeconvolutionStains.java
@@ -588,7 +588,7 @@ public class ColorDeconvolutionStains implements Externalizable {
 	@Override
 	public void writeExternal(ObjectOutput out) throws IOException {
 		out.writeInt(version);
-		out.writeObject(getColorDeconvolutionStainsAsString(this, 8));
+		out.writeObject(getColorDeconvolutionStainsAsString(this, 32));
 	}
 
 
--- a/qupath-core/src/main/java/qupath/lib/images/ImageData.java
+++ b/qupath-core/src/main/java/qupath/lib/images/ImageData.java
@@ -25,12 +25,10 @@ package qupath.lib.images;
 
 import java.beans.PropertyChangeListener;
 import java.beans.PropertyChangeSupport;
-import java.io.IOException;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
-import java.util.function.Supplier;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -360,7 +358,7 @@ public class ImageData<T> implements WorkflowListener, PathObjectHierarchyListen
 			return;
 		}
 		
-		String arg = ColorDeconvolutionStains.getColorDeconvolutionStainsAsString(imageData.getColorDeconvolutionStains(), 17);
+		String arg = ColorDeconvolutionStains.getColorDeconvolutionStainsAsString(imageData.getColorDeconvolutionStains(), 32);
 		Map<String, String> map = GeneralTools.parseArgStringValues(arg);
 		WorkflowStep lastStep = imageData.getHistoryWorkflow().getLastStep();
 		String commandName = "Set color deconvolution stains";
--- a/qupath-core/src/main/java/qupath/lib/images/ImageData.java
+++ b/qupath-core/src/main/java/qupath/lib/images/ImageData.java
@@ -360,7 +360,7 @@ public class ImageData<T> implements WorkflowListener, PathObjectHierarchyListen
 			return;
 		}
 		
-		String arg = ColorDeconvolutionStains.getColorDeconvolutionStainsAsString(imageData.getColorDeconvolutionStains(), 5);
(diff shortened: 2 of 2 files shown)


petebankhead:
Rather than specify a number of decimal places, should we just rely on string concatenation using `double`?
Or, since trailing zeros should be omitted anyway, just go for 32 or some other value that is less surprising than 17?

I'm tentatively in favour of this change as a quick improvement with minimal disruption, but we will need to look at this issue in more detail at some point soon.

The current code is very old, pre-dating the frequent use of Gson within QuPath, and has numerous questionable decisions. Notably, even when serialising we appear to (unnecessarily...) impose 8 decimal places [here](https://github.com/qupath/qupath/blob/a74e6818ec98df4da7827deacbf8653a3734865a/qupath-core/src/main/java/qupath/lib/color/ColorDeconvolutionStains.java#L591). I think that probably we should be using the same limit as in the workflow.

What's more, we are using a weird representation of the stains, which looks more like a JSON representation than it really is:
```groovy
import qupath.lib.color.*

def stains = getCurrentImageData().getColorDeconvolutionStains()

def toString = ColorDeconvolutionStains.getColorDeconvolutionStainsAsString(stains, 1000)
println "String representation:"
println "   " + toString

def json = GsonTools.getInstance(false).toJson(stains)
println "JSON from Gson:"
println "   " + json

// This won't work, because it's not really a JSON representation of the stains object!
println "Deserialized string to stains using Gson:"
println "   " + GsonTools.getInstance(true).fromJson(toString, ColorDeconvolutionStains.class)

// This does print something kind of meaningful
println "Deserialized string to map using Gson:"
println "   " + GsonTools.getInstance(true).fromJson(toString, Map.class)
```

For me this prints
```
String representation:
   {"Name" : "H&E default", "Stain 1" : "Hematoxylin", "Values 1" : "0.6511078257574492 0.7011930431234068 0.29049426072255424", "Stain 2" : "Eosin", "Values 2" : "0.21589893562087106 0.8011960501132093 0.5580972485873467", "Background" : " 255 255 255"}
JSON from Gson:
   {"name":"H\u0026E default","stain1":{"r":0.6511078257574492,"g":0.7011930431234068,"b":0.29049426072255424,"name":"Hematoxylin","isResidual":false},"stain2":{"r":0.21589893562087106,"g":0.8011960501132093,"b":0.5580972485873467,"name":"Eosin","isResidual":false},"stain3":{"r":0.315510575173205,"g":-0.5981592020094374,"b":0.736653681186286,"name":"Residual","isResidual":true},"maxRed":255.0,"maxGreen":255.0,"maxBlue":255.0}
Deserialized string to stains using Gson:
   Color deconvolution stains: null, null, null
Deserialized string to map using Gson:
   [Name:H&E default, Stain 1:Hematoxylin, Values 1:0.6511078257574492 0.7011930431234068 0.29049426072255424, Stain 2:Eosin, Values 2:0.21589893562087106 0.8011960501132093 0.5580972485873467, Background: 255 255 255]
```

We have [`QP.setColorDeconvolutionStains(String)`](https://github.com/qupath/qupath/blob/a74e6818ec98df4da7827deacbf8653a3734865a/qupath-core-processing/src/main/java/qupath/lib/scripting/QP.java#L1901) that takes our weird, custom, JSON-ish version of the stains - but doesn't handle a JSON representation serialised by Gson.

I propose:
1. We update this PR to use the same number of decimal places for serialising to the .qpdata file, then merge for v0.6.0
2. We consider a *separate* PR to write a custom JSON serializer/deserializer for `ColorDeconvolutionStains` so that we can serialise and deserialise with Gson, with a view to removing the custom string stuff entirely
3. We think about whether there is another syntax that would be more readable in a script than a huge string

2 and 3 might be after v0.6.0, although if there is a quick change we could make that would help maintain compatibility then that would be worth considering (as a separate PR).

For 3, I'm imagining something like `QP.setColorDeconvolutionStains(Map)` so we might use
```java
setColorDeconvolutionStains("Hematoxylin": [0.6511078257574492, 0.7011930431234068, 0.29049426072255424], "Eosin": [0.21589893562087106, 0.8011960501132093, 0.5580972485873467], "Background": [255, 255, 255])
```

PR #1888, opened 2025-05-28
Add viewer placeholder

petebankhead:
Make it more obvious that drag & drop can be used to open images, by adding a placeholder to the viewer.

Currently, 2 different messages are included.

## No image or project open

image

## No image open

image

Either way, it looks weird whenever multiple viewers are open.

image

My main questions:
* Any improvements on the wording?
  * If you have a project open, the message should maybe be to double-click an image under the project tab...  but that is wordy, and the fact you can use drag & drop for import may be less obvious
* Any opinions on whether only the active viewer should show the text, or all viewers?


Rylern:
> If you have a project open, the message should maybe be to double-click an image under the project tab... but that is wordy, and the fact you can use drag & drop for import may be less obvious

Have you tried to put both, on two different lines?

Issue #1786, opened 2025-03-05
shared windows paths are badly interpreted when opened on *nix


carlocastoldi:
## Before we begin...

Before submitting your bug report, please check the following:

* [x] I've definitely found a bug (it you're not sure, please use [image.sc](https://forum.image.sc/tags/qupath) instead)
* [x] I've checked https://qupath.github.io for a new release that might already have fixed the issue
* [x] I've checked the [Changelog](https://github.com/qupath/qupath/blob/master/CHANGELOG.md) to see if the bug has already been fixed in the next release
* [x] I've checked for existing GitHub issues describing the same problem

### Describe the bug
This bug arises only on *nix systems when opening a QuPath project that was last opened on Windows and that has images on a [shared network](https://en.wikipedia.org/wiki/Shared_resource#Naming_convention_and_mapping).
On windows, shared resources path follow the following convention:
```
\\ServerComputerName\ShareName\Path\To\Image
```

When opening the project on linux, however, you are not prompted by QuPath that it can't find the associated images. Instead, you get an error when trying to opening an image:

<details>
<summary>java.io.FileNotFoundException</summary>

```
Unable to open file://v7ku1.techosp.it/ricerca/Lab%20Matteoli/Silva/projects/data/APP_PNS/Cohort%204/GABA_cfos/images_females/7728_new/7728_female_GABAcFos.czi: file:/v7ku1.techosp.it/ricerca/Lab%20Matteoli/Silva/projects/data/APP_PNS/Cohort%204/GABA_cfos/images_females/7728_new/7728_female_GABAcFos.czi (No such file or directory)
java.io.FileNotFoundException: file:/v7ku1.techosp.it/ricerca/Lab%20Matteoli/Silva/projects/data/APP_PNS/Cohort%204/GABA_cfos/images_females/7728_new/7728_female_GABAcFos.czi (No such file or directory)
	at java.base/java.io.RandomAccessFile.open0(Native Method)
	at java.base/java.io.RandomAccessFile.open(Unknown Source)
	at java.base/java.io.RandomAccessFile.<init>(Unknown Source)
	at java.base/java.io.RandomAccessFile.<init>(Unknown Source)
	at loci.common.NIOFileHandle.<init>(NIOFileHandle.java:130)
	at loci.common.NIOFileHandle.<init>(NIOFileHandle.java:151)
	at loci.common.NIOFileHandle.<init>(NIOFileHandle.java:165)
	at loci.common.Location.getHandle(Location.java:522)
[...]

PR #1889, opened 2025-05-28
Flag local files as missing instead of unknown in URI updater

Rylern:
Address #1786.

Here is a project containing an image whose path was on a Windows shared network: 
[qupath_project_created_on_shared_resource_folder_windows.zip](https://github.com/user-attachments/files/20490002/qupath_project_created_on_shared_resource_folder_windows.zip)

Before this PR:
* Opening this project on Windows was showing the URI updater window if the path of the image was not found.
* Opening this project on Linux/macOS was not showing the URI updater window even if the path of the image was not found.

With this PR, opening this project on Windows/Linux/macOS shows the URI updater window if the path of the image is not found.

Implementation details:
`file://cmvm.datastore.ed.ac.uk/cmvm/smgphs/users/lleplat/CMU-1.jpg` is an example path to a file on a Windows shared network. `Paths.get(thisUri)` returns a valid path on Windows, while it throws an `IllegalArgumentException` on Linux/macOS. Thus, before this PR, `UriUpdater` was considering this path to be a local file on Windows, but not on Linux/macOS. This explains the difference of behaviour.
This PR considers that an `URI` with `file` as the scheme refers to a local file, whether or not a `Path` can be constructed from it. This enables to have a consistent behaviour between all platforms.

The changed code in `searchDirectoriesRecursive` was to make the function work even if a `Path` is not present (which happens on Linux/macOS with the previous URI).

### DIFF
--- a/qupath-gui-fx/src/main/java/qupath/lib/gui/commands/UpdateUrisCommand.java
+++ b/qupath-gui-fx/src/main/java/qupath/lib/gui/commands/UpdateUrisCommand.java
@@ -24,6 +24,8 @@ package qupath.lib.gui.commands;
 import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.net.URLDecoder;
+import java.nio.charset.StandardCharsets;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.Scanner;
@@ -371,8 +373,9 @@ public class UpdateUrisCommand<T extends UriResource> {
 				setText("");
 				return;
 			}
-			setText(item.toString());
-			tooltip.setText(item.toString());
+			String displayableText = URLDecoder.decode(item.toString(), StandardCharsets.UTF_8);
+			setText(displayableText);
+			tooltip.setText(displayableText);
 			setTooltip(tooltip);
 
 			switch (item.getStatus()) {


carlocastoldi:
Hi!
Weirdly enough, the projects I was collaborating with Windows users on are now pointing to image paths that start with `file:////NEWORK-LOCATION/` (4 forward slashes instead of 2, what this PR fixes).
This means that I cannot directly test it, however if I edit the `qpproj` files to have only 2 forward slashes this PR indeed behaves as expected! It shows the "Update URIs" pop up.

The only questionable difference that I noticed is that it does not render ASCII encodings properly.

Two forward slashes with this PR:
image

Four forward slashes (not triggering this PR):
image

Rylern:
I wasn't able to get image paths that start with `file:////NETWORK-LOCATION/`, could you give some details on how to get that?

Also, what exactly is the issue when you have image paths that start with `file:////NETWORK-LOCATION/`? On your screenshots, it seems that both situations work as expected (`Update URIs` window open and showing invalid URIs).

carlocastoldi:
Oh sorry for explaining myself badly.

There is NO problem with having 4 forward slashes instead of 2. However I don't know how my colleagues managed to have it that way. They're on 0.5.1

Could it be a difference due to win10 vs win11?

PR #1890, opened 2025-05-28
Update bioformats to 8.2.0 and ome zarr reader to 0.6.0

Rylern:
Update bioformats to 8.2.0 and ome zarr reader to 0.6.0


petebankhead:
Could you update these in the changelog too please? We should list any major dependency updates there.

Issue #1892, opened 2025-05-29
QuPath tour issues (darkmode and lag)


finglis:
Before submitting your bug report, please check the following:

* [ x] I've definitely found a bug (it you're not sure, please use [image.sc](https://forum.image.sc/tags/qupath) instead)
* [x ] I've checked https://qupath.github.io for a new release that might already have fixed the issue
* [ x] I've checked the [Changelog](https://github.com/qupath/qupath/blob/master/CHANGELOG.md) to see if the bug has already been fixed in the next release
* [ x] I've checked for existing GitHub issues describing the same problem

### Describe the bug
QuPath tour currently ignores styling from dark mode when opening and additionally there is some lag with the image in the tour. 

### To Reproduce
Steps to reproduce the darkmode behavior:
1. Go to Help → QuPath Tour

Steps to reproduce the darkmode behavior:
1. Open an image and then open Qupath tour
2. Note the image in the tour matches
3. Close QuPath tour and open another image
4. Open QuPath tour and note that the image remains the first image. Pan a few slides and it will update to the recent one. 

### Expected behavior
Darkmode styling is applied and the image in the tour represents the current image open. 

### Screenshots

![Image](https://github.com/user-attachments/assets/b3d796eb-ad5d-471f-887a-33c50001a3f6)

### Desktop (please complete the following information):
 - OS: Windows
 - QuPath Version: 0.6.0 SNAPSHOT

petebankhead:
The image is generated when the page is opened. I suppose we *could* reset it when closing the tour, but I'm not sure if it really matters? You shouldn't have to change a few slides - just change one.

Re. the styling, it works for me on my Mac.... so I'm not able to investigate.

One thing to check is whether there's any difference if dark mode is toggled *after* QuPath has been launched (as opposed to being applied during startup).

PR #1893, opened 2025-05-29
Set webview based on colorscheme not style name content

alanocallaghan:
Fixes https://github.com/qupath/qupath/issues/1892


petebankhead:
Added link to the original issue. Admittedly obscure, but this doesn't change the webview if the colorscheme is changed in the system preferences *while* the webview is being shown (i.e. I choose to use the system colorscheme in QuPath, then switch dark/light in the system itself).

petebankhead:
Yes, I think it should - having a property for the `ColorScheme` that uses a custom binding to query the style and use the system value if the style can't help.

Still, I'll merge this anyway since it fixes the behavior.

Issue #1896, opened 2025-05-29
New resolve hierarchy causes QuPath to crash


saramcardle:
### Describe the bug
The new "Resolve Hierarchy" [function ](https://github.com/qupath/qupath/pull/1824) can cause QuPath to freeze/hang when dealing with many complex annotations, even with no detections. After a few minutes of thinking, my Task Manager shows that QuPath drops to 0% CPU and minimal memory and it never recovers. 

I looked at the log on Trace mode in both versions. In rc5, I get a long series of Cache reconstructed ... Cache reset messages and then it just stops. 

```
09:21:34.868 [JavaFX Application Thread] [DEBUG] q.l.o.hierarchy.PathObjectTileCache - Cache reconstructed in 0 ms
09:21:34.874 [JavaFX Application Thread] [TRACE] q.l.o.hierarchy.PathObjectTileCache - Cache reset!
09:21:34.875 [JavaFX Application Thread] [DEBUG] q.l.o.hierarchy.PathObjectTileCache - Cache reconstructed in 0 ms
```

In rc3, which works with the older version of the function, I see many more of those cache messages, followed by:

```
09:14:45.159 [JavaFX Application Thread] [DEBUG] q.l.g.p.SelectedMeasurementTableView - Measurement table update skipped (not visible)
09:14:45.160 [JavaFX Application Thread] [DEBUG] q.l.gui.panes.ObjectDescriptionPane - Updating details pane for Image
09:14:45.161 [JavaFX Application Thread] [DEBUG] q.l.g.p.SelectedMeasurementTableView - Measurement table update requested
09:14:45.162 [JavaFX Application Thread] [DEBUG] q.l.gui.panes.ObjectDescriptionPane - Updating details pane for Image
09:14:45.248 [JavaFX Application Thread] [TRACE] qupath.lib.objects.MetadataMap - Creating MetadataMap
09:14:45.249 [JavaFX Application Thread] [TRACE] qupath.lib.objects.MetadataMap - Creating MetadataMap

```
And then many more of the MetadataMap lines. So, I think the issue is either in the cache reconstruction step or immediately after that?

### To Reproduce
1. In any large WSI, load the geojson objects here: [Geojson for bug report.zip](https://github.com/user-attachments/files/20507933/Geojson.for.bug.report.zip) 
2. Try to resolve the hierarchy. 

### Expected behavior
Ideally, it successfully resolves the hierarchy as it's expected to. Short of that, it exits the function without crashing. 

[...]

PR #1897, opened 2025-05-29
Improve computeCoversWithTolerance performance

petebankhead:
Attempts to fix https://github.com/qupath/qupath/issues/1896

### DIFF
--- a/qupath-core/src/main/java/qupath/lib/objects/hierarchy/RoiRelate.java
+++ b/qupath-core/src/main/java/qupath/lib/objects/hierarchy/RoiRelate.java
@@ -21,6 +21,7 @@
 
 package qupath.lib.objects.hierarchy;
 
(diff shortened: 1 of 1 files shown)


petebankhead:
Here's a script to show how the maximum inscribed circle looks:
https://gist.github.com/petebankhead/074dd8dad81a536978930f8b156c251a

The thinking is that the 'covers' problem emerges when one 'inner' ROI is outside another 'outer' ROI by a tiny amount, smaller than the precision of JTS's overlay operations, so that it really should still be considered as 'inside'. This can happen when we compute the intersection of two ROIs, where the intersection *should* be inside both the original ROIs - but because of precision errors, it sometimes isn't.

Fitting a circle to the difference between the ROIs tries to give a fast estimate of the 'thickness', and therefore the maximum distance the 'inner' ROI extends beyond the 'outer' ROI. If this is more than the tolerance, we don't need to do any more expensive and detailed checks.

image

PR #1899, opened 2025-06-02
Allow toggling multiple channels simultaneously in B&C

alanocallaghan:
Currently, when multiple items are selected in brightness & contrast, clicking a checkbox will toggle the display of the single item being clicked, not the multiple items selected.

With the proposed change, shift+click on one of the selected items will set all of the selected channels to the inverse of whether the current item is enabled.

This somehow isn't exactly how I would like things to behave, since you can also do multiple selection with Ctrl/Cmd as well as shift, but may be a slight improvement. I couldn't find an easy way to improve the logic for Ctrl at the same time.


petebankhead:
I can see how this would be kind of convenient, although I think the dual purpose of a shift+click is a bit confusing (depending upon whether I'm clicking on an item that is selected or not).

Is this necessary at all, when you can get the effect (with more control) via a right-click?

Screenshot 2025-06-03 at 10 08 26

PR #1900, opened 2025-06-02
Increase min size labels of spinners in thresholder window

alanocallaghan:
Prevents the numbers from jumping around.

https://github.com/user-attachments/assets/30ee33d3-f728-4d49-be6e-8c6f9b7800cd

### DIFF
--- a/qupath-extension-processing/src/main/java/qupath/process/gui/commands/SimpleThresholdCommand.java
+++ b/qupath-extension-processing/src/main/java/qupath/process/gui/commands/SimpleThresholdCommand.java
@@ -232,7 +232,7 @@ public class SimpleThresholdCommand implements Runnable {
 		labelSigma.textProperty().bind(
 				Bindings.createStringBinding(() -> sigma.get() == null ? "" : GeneralTools.formatNumber(sigma.get(), 2), sigma)
 				);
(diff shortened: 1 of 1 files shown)


alanocallaghan:
> I don't know how best to make it obvious to a user that numbers they are typing don't immediately update the display, but rather need to be committed (with a keypress, or by focusing another control).

Is this necessary? We have a lot of textfields that have the same behaviour, where the value needs to be committed before applying.

petebankhead:
> > None of the non-specific specifics seem to me exactly the same
> 
> I have no idea what this means.

There are a lot of things in QuPath, it helps my old brain it's unambiguous. I had to assume you were referring to measurement tables... and take a guess on what 'several others' might be. 

> Yeah fair enough, I just really want the jumping buttons fixed and to me this seemed the easy & obvious fix, but if you have concerns about user analyses I'm happy to try a fix with element sizing (although this could still in theory break, eg in 16 bit images)

Min size might do it? I don't think jumping needs eliminated entirely, if it can at least be reduced.

Longer term, the information might be expressed elsewhere in the dialog - but I think fixing the labels for now is easiest.

alanocallaghan:
Right, increased the min size so we don't get jumping until the threshold is something silly like 10000000. It does eat a bit into the text field space, but unless somebody is setting thresholds down to the 10th decimal place I can't see it being a problem.

PR #1841, opened 2025-04-15
Combine overlay toolbar buttons

petebankhead:
Create space for more buttons by combining those relating to the viewer thumbnail, grid, scalebar and location (which are rarely changed) under a MenuButton.

PR #1901, opened 2025-06-02
Add toolbar button and accelerator to toggle detection display mode

alanocallaghan:
Introduce a toolbar button to toggle the detection display mode.

Ideally, the icon would reflect the current display mode.

This is a "nice to have" for me, but I appreciate we are close to polluting the toolbar at this point...

### DIFF
--- a/qupath-gui-fx/src/main/java/qupath/lib/gui/ToolBarComponent.java
+++ b/qupath-gui-fx/src/main/java/qupath/lib/gui/ToolBarComponent.java
@@ -150,7 +150,6 @@ class ToolBarComponent {
 		nodes.add(createToggleButton(overlayActions.FILL_DETECTIONS));
 		// TODO: Consider removing 'Show connections' button until it becomes more useful
 		nodes.add(createToggleButton(overlayActions.SHOW_CONNECTIONS));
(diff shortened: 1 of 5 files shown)


petebankhead:
Toolbar space is really at a premium - https://github.com/qupath/qupath/pull/1841 was made after @finglis pointed out that the toolbar was an issue in workshops when people have small laptop screens.

I think a `MenuButton` would be a more natural choice here anyway, but would cost any more space... unless it can absorb other toolbar buttons, to give us a net saving. I don't find the `ToggleButton` with differing purposes intuitive, and it seems to be inviting more confusion.

petebankhead:
Another (possibly too radical at this point) idea would be to make the drawing tools a vertical toolbar, thereby freeing up *loads* of space...

alanocallaghan:
> Toggling the cell display may be handy for a pro users, but for most people I think it's better to choose a mode explicitly from the context menu.

I think it may become more popular if somebody were to introduce broadly useful cell and nucleus segmentation methods ;)

> To me, this is the kind of custom behavior that we could ideally make installable through a startup script or extension

Fair enough; I will close this and just make an extension for my own QOL stuff

PR #1910, opened 2025-06-06
Only create pyramidal server if necessary + change default parameters

Rylern:
In the zarr writer:

* Change the default builder parameters to use the ones of the input image server.
* Create a `PyramidGeneratingImageServer` only if necessary.


petebankhead:
Haven’t run the code, but from reading it the PR makes sense to me. One thing I wondered re.
```java
Arrays.equals(downsamples, builder.server.getPreferredDownsamples())
```
Would you still need to create a pyramidalizing server if the downsamples are a subset of the original servers downsamples?

e.g. if you want to export with downsamples `[1.0]` and the server has `[1.0, 2.0, 4.0]` could you still use the server unchanged?

(Not sure if there’s a benefit in this or not, but if this PR aims to reduce the need to create a pyramid-generating server I wondered if it could be reduced even further)

petebankhead:
Ah, I don’t think so - I think creating a pyramidalizing server is the easiest way.

PR #1936, opened 2025-06-26
Improve dark mode icon contrast when focused

alanocallaghan:
Before:

Screenshot from 2025-06-26 11-15-10

With this change:

Screenshot from 2025-06-26 11-15-31


alanocallaghan:
Checked the code and couldn't identify other locations with listviews and pathicons.

PR #1956, opened 2025-07-18
Add on close runnables

Rylern:
With the OMERO extension, I need some code to be executed when the QuPath window is closed (to close all active OMERO connections in order to free up resources on the server side).

I initially tried to use `QuPathGUI.getStage().getScene().getWindow().addEventFilter(WindowEvent.WINDOW_CLOSE_REQUEST, ...)`, but the function was executed when the user **attempts** to close the window, not when the window is actually closed. For example, if an image is open and the user attempts to close QuPath, QuPath will prompt to save data, and if the user clicks on `Cancel`, then QuPath is not closed, but the `WindowEvent.WINDOW_CLOSE_REQUEST` event is triggered.

The only solution I saw was to create this PR.

### DIFF
--- a/qupath-gui-fx/src/main/java/qupath/lib/gui/QuPathGUI.java
+++ b/qupath-gui-fx/src/main/java/qupath/lib/gui/QuPathGUI.java
@@ -261,8 +261,10 @@ public class QuPathGUI {
 	 * can result in exceptions (or even segfaults...).
 	 */
 	private BooleanProperty menusInitializing = new SimpleBooleanProperty(false);
-	
-	
+
+	private final List<Runnable> onCloseRunnables = new ArrayList<>();
+
+
 	/**
 	 * Create a new QuPath instance.
 	 * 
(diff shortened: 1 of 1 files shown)


petebankhead:
Could [`Runtime.addShutdownHook(Thread)`](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/Runtime.html#addShutdownHook(java.lang.Thread)) be used? See also https://github.com/qupath/qupath/blob/9f6e3bfff3fe39901b9ff292381abfc25d5f6f84/qupath-extension-bioformats/src/main/java/qupath/lib/images/servers/bioformats/BioFormatsImageServer.java#L1689

Rylern:
I didn't know this function existed but it seems good. The only thing is that the documentation of the function states:

> Shutdown hooks should also finish their work quickly. When a program invokes [exit](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/Runtime.html#exit(int)), the expectation is that the virtual machine will promptly shut down and exit. When the virtual machine is terminated due to user logoff or system shutdown the underlying operating system may only allow a limited amount of time in which to shut down and exit. It is therefore inadvisable to attempt any user interaction or to perform a long-running computation in a shutdown hook.

while closing an OMERO connection is sending a web request, which can (in the worst case scenario) takes up to 20 seconds per closure. Do you think I should use it still?

PR #1957, opened 2025-07-28
avoid useless loadings of ImageServers

carlocastoldi:
QuPath 0.6.0 always loaded some `ImageServer` just to then close them right after.
This occurred when closing the object classifier's window and when executing a CLI script.

Actually, I expect these precautionary `getServer()` → `close()` to have slipped through in other parts. These two cases jumped to me because CLI scripting and object classifiers are tools that I regularly use.

Related to #1489 and #1579

### DIFF
--- a/qupath-app/src/main/java/qupath/QuPath.java
+++ b/qupath-app/src/main/java/qupath/QuPath.java
@@ -397,7 +397,8 @@ class ScriptCommand implements Runnable {
 						if (imagePath != null && imagePath.equals(entry.getImageName()))
 							throw new RuntimeException(e);
 					} finally {
-						imageData.getServer().close();						
+						if (imageData.isLoaded())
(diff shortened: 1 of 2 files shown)


petebankhead:
Thanks @carlocastoldi could you please review https://github.com/qupath/qupath/pull/1959 and see if it fixes the issue for you? I rarely use these tools & don't have a quick way to test them.

I think the other PR handles more cases, and relying on `ImageData` implementing `AutoCloseable` makes it possible to simplify things slightly (and closing the `ImageData` should [already do much the same thing](https://github.com/qupath/qupath/blob/9f6e3bfff3fe39901b9ff292381abfc25d5f6f84/qupath-core/src/main/java/qupath/lib/images/ImageData.java#L584)).

If the other PR looks good to you, I'd merge it and close this one.

PR #1964, opened 2025-08-07
Pyramidal zarr writer

Rylern:
Add a zarr writer that (in that order):

* Write the full resolution level by looking at the input image.
* Write the second level by looking at the first level pixels that were just written (the input image is not considered anymore).
* Write the third level by looking at the second level pixels that were just written.

This provides less flexibility than the existing `OMEZarrWriter` but is less likely to throw out of memory errors.

This PR only adds this new writer (`PyramidalOMEZarrWriter`). Other changes to existing classes were made to reuse as much code as possible between the two zarr writers.

I create this PR now to discuss about the public API (is only one `writeImage()` function enough?), but it's a draft because tests need to be added.

### DIFF
--- a/qupath-extension-bioformats/src/main/java/qupath/lib/images/writers/ome/zarr/PyramidalOMEZarrWriter.java
+++ b/qupath-extension-bioformats/src/main/java/qupath/lib/images/writers/ome/zarr/PyramidalOMEZarrWriter.java
@@ -50,7 +50,8 @@ public class PyramidalOMEZarrWriter {
 
     private static final Logger logger = LoggerFactory.getLogger(PyramidalOMEZarrWriter.class);
     private final ImageServer<BufferedImage> server;
-    private final List<Double> downsamples;
+    private final double firstDownsampleOnInputImage;
+    private final List<Double> downsamplesOutputImage;
     private final int tileWidth;
     private final int tileHeight;
     private final int numberOfThreads;
@@ -74,7 +75,10 @@ public class PyramidalOMEZarrWriter {
         }
         this.server = transformedServerBuilder.build();
 
-        this.downsamples = builder.downsamples;
+        this.firstDownsampleOnInputImage = builder.downsamples.getFirst();
+        this.downsamplesOutputImage = builder.downsamples.stream()
+                .map(downsample -> downsample / this.firstDownsampleOnInputImage)
+                .toList();
         this.tileWidth = ZarrWriterUtils.getChunkSize(
                 builder.tileWidth > 0 ? builder.tileWidth : builder.server.getMetadata().getPreferredTileWidth(),
                 builder.maxNumberOfChunks,
@@ -97,8 +101,11 @@ public class PyramidalOMEZarrWriter {
         }
 
         this.levels = ZarrWriterUtils.createLevels(
-                server.getMetadata(),
-                downsamples,
+                new ImageServerMetadata.Builder(server.getMetadata())
+                        .width((int) (server.getWidth() / firstDownsampleOnInputImage))
+                        .height((int) (server.getHeight() / firstDownsampleOnInputImage))
+                        .levelsFromDownsamples(downsamplesOutputImage.stream().mapToDouble(Double::doubleValue).toArray())
+                        .build(),
                 root,
                 tileWidth,
                 tileHeight,
@@ -122,41 +129,44 @@ public class PyramidalOMEZarrWriter {
         );
 
         try {
+            int numberOfDownsamples = downsamplesOutputImage.size();
             root.writeAttributes(
                     new OMEZarrAttributesCreator(new ImageServerMetadata.Builder(server.getMetadata())
-                            .levelsFromDownsamples(downsamples.getFirst())
+                            .levelsFromDownsamples(firstDownsampleOnInputImage)
                             .build()
                     ).getGroupAttributes()
             );
             writeLevel(
-                    path.toString(),
+                    server.getPath(),
                     0,
+                    firstDownsampleOnInputImage,
                     levels.get(0),
                     server,
                     executorService,
-                    progress -> onProgress.accept(progress / downsamples.size())
+                    progress -> onProgress.accept(progress / numberOfDownsamples)
             );
 
-            for (int i=1; i<downsamples.size(); i++) {
+            for (int i=1; i<numberOfDownsamples; i++) {
                 try (ImageServer<BufferedImage> server = new BioFormatsImageServer(
                         path.toUri(),
-                        "--series",                         // since all level zarr subgroups are already created (see the constructor), BioFormats treat each subgroup
-                        String.valueOf(downsamples.size() - i)  // as a different series, so the series we are interested in must be specified
+                        "--series",                          // since all level zarr subgroups are already created (see the constructor), BioFormats treat each subgroup
+                        String.valueOf(numberOfDownsamples - i)  // as a different series, so the series we are interested in must be specified
                 )) {
-                    double progressOffset = (double) i / downsamples.size();
+                    double progressOffset = (double) i / numberOfDownsamples;
                     writeLevel(
                             path.toString(),
                             i,
+                            downsamplesOutputImage.get(i),
                             levels.get(i),
                             server,
                             executorService,
-                            progress -> onProgress.accept(progressOffset + progress / downsamples.size())
+                            progress -> onProgress.accept(progressOffset + progress / numberOfDownsamples)
                     );
                 }
 
                 root.writeAttributes(
                         new OMEZarrAttributesCreator(new ImageServerMetadata.Builder(server.getMetadata())
-                                .levelsFromDownsamples(downsamples.stream().limit(i+1).mapToDouble(d -> d).toArray())
+                                .levelsFromDownsamples(downsamplesOutputImage.stream().limit(i+1).mapToDouble(d -> d).toArray())
                                 .build()
                         ).getGroupAttributes()
                 );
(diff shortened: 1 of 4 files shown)


qupath-extension-bioformats/src/main/java/qupath/lib/images/writers/ome/zarr/WriterUtils.java
```diff
@@ -0,0 +1,279 @@
...
+     * </ul>
+     *
+     * @param image the image to convert
+     * @param metadata the metadata of the image to convert. It must match the provided image (e.g. it must be RGB if the
+     *                 provided image has the RGB format, they must have the same number of channels and pixel type), except
+     *                 for the width and height that can be different
+     * @return a one-dimensional array as described above. The pixel located at coordinates [x; y; c] can be accessed with
+     * returnedArray[x + imageWidth * y + imageWidth * imageHeight * c]
+     */
+    public static Object convertBufferedImageToArray(BufferedImage image, ImageServerMetadata metadata) {
```
petebankhead, line 212: `ImageServerMetadata` doesn't look necessary here - and it creates a dependency upon `AWTImageTools.getPixels` behaving in a way that is fully consistent with QuPath's metadata. *Conceivably* `metadata.isRGB()` might one day return true for 16-bit RGB images (although it probably shouldn't).

Here is a suggested alternative implementation that
* Should handle RGB more efficiently _(although it might not do the right thing for exotic RGB/BGR types supported by `BufferedImage`, or for indexed color images! In these cases I don't know what's expected)_
   * This isn't crucial & could be skipped
* Uses the return type of `getPixels` rather than relying upon it matching the metadata
* Doesn't copy single-channel values

```java
public static Object convertBufferedImageToArray(BufferedImage image, ImageServerMetadata metadata) {

        int channelSize = image.getWidth() * image.getHeight();

        // Handle RGB without unnecessary creation of int[] arrays
        if (BufferedImageTools.is8bitColorType(image.getType())) {
            byte[] pixels = new byte[image.getWidth() * image.getHeight() * 3];
            int[] rgb = image.getRGB(0, 0, image.getWidth(), image.getHeight(), null, 0, image.getWidth());
            for (int i = 0; i < channelSize; i++) {
                int val = rgb[i];
                pixels[i] = (byte)ColorTools.red(val);
                pixels[channelSize + i] = (byte)ColorTools.green(val);
                pixels[channelSize*2 + i] = (byte)ColorTools.blue(val);
            }
            return pixels;
        }

        Object pixels = AWTImageTools.getPixels(image);

        // No need to copy for single array
        int nChannels = Array.getLength(pixels);
        if (nChannels == 1)
            return Array.get(pixels, 0);

        if (pixels instanceof byte[][] data) {
            byte[] output = new byte[nChannels * channelSize];
            for (int c=0; c<nChannels; c++) {
                System.arraycopy(data[c], 0, output, c * channelSize, channelSize);
            }
            return output;
        } else if (pixels instanceof short[][] data) {
            short[] output = new short[nChannels * channelSize];
            for (int c=0; c<nChannels; c++) {
                System.arraycopy(data[c], 0, output, c * channelSize, channelSize);
            }
            return output;
        } else if (pixels instanceof int[][] data) {
            int[] output = new int[nChannels * channelSize];
            for (int c=0; c<nChannels; c++) {
                System.arraycopy(data[c], 0, output, c * channelSize, channelSize);
            }
            return output;
        } else if (pixels instanceof float[][] data) {
            float[] output = new float[nChannels * channelSize];
            for (int c=0; c<nChannels; c++) {
                System.arraycopy(data[c], 0, output, c * channelSize, channelSize);
            }
            return output;
        } else if (pixels instanceof double[][] data) {
            double[] output = new double[nChannels * channelSize];
            for (int c=0; c<nChannels; c++) {
                System.arraycopy(data[c], 0, output, c * channelSize, channelSize);
            }
            return output;
        } else
            throw new UnsupportedOperationException("Unknown data type");
    }
```

qupath-extension-bioformats/src/main/java/qupath/lib/images/writers/ome/zarr/WriterUtils.java
```diff
@@ -0,0 +1,279 @@
...
+        Object pixels = AWTImageTools.getPixels(image);
+
+        if (metadata.isRGB()) {
+            int[][] data = (int[][]) pixels;
+            int channelSize = image.getWidth() * image.getHeight();
+
+            int[] output = new int[metadata.getSizeC() * channelSize];
+            for (int c=0; c<metadata.getSizeC(); ++c) {
```
petebankhead, line 220: I think the normal idiom is `c++` - any reason to change to pre-increment?

qupath-extension-bioformats/src/main/java/qupath/lib/images/writers/ome/zarr/PyramidalOMEZarrWriter.java
```diff
@@ -0,0 +1,474 @@
...
+         * Create the builder.
+         *
+         * @param server the image to write. Only the level with downsample 1 will be considered (so it must exist)
+         * @param downsamples the downsamples the output image should have. There must be at least one downsample,
+         *                    and one downsample should be equal to 1
+         * @throws IllegalArgumentException if the downsample list is empty or doesn't contain 1, or if the provided
+         * server doesn't contain a level with downsample 1
+         * @throws NullPointerException if one of the provided parameter is null
+         */
+        public Builder(ImageServer<BufferedImage> server, List<Double> downsamples) {
```
petebankhead, line 206: Could we use `double... downsamples`?

`List<Double>` is actually hard to get right from Groovy - passing `[1, 4, 16]` gives me errors at runtime because it doesn't automatically convert to `Double`... and `[1.0, 4.0, 16.0]` doesn't work because it uses `BigDecimal`.

Using `double...` is easier and means that `null` isn't possible, but if you really want a list then `List<? extends Number>` could help, then call `doubleValue()` on each element when needed.

qupath-extension-bioformats/src/main/java/qupath/lib/images/writers/ome/zarr/PyramidalOMEZarrWriter.java
```diff
@@ -0,0 +1,474 @@
...
+            if (downsamples.isEmpty()) {
+                throw new IllegalArgumentException("At least one downsample should be provided");
+            }
+            if (downsamples.stream().noneMatch(downsample -> downsample == 1)) {
```
petebankhead, line 210: * Is it possible for the downsamples to be out of order - or should they always be ascending?
* Can or should we be checking for invalid or duplicate values?
* Is it necessary to start with 1? I could imagine exporting with `[2, 4, 8]`.

Requiring a downsample of 1 is ok if necessary, but note that the following script will print `2`
```groovy
import qupath.lib.images.servers.*
println ImageServers.pyramidalize(getCurrentServer(), 2, 4).getPreferredDownsamples()[0]
```
i.e. it's not _guaranteed_ that the first resolution returned by an `ImageServer` has a downsample of `1`.... even if it almost always will.

qupath-extension-bioformats/src/main/java/qupath/lib/images/writers/ome/zarr/PyramidalOMEZarrWriter.java
```diff
@@ -0,0 +1,474 @@
...
+                    .levelsFromDownsamples(1)
+                    .build()
+            ).getGroupAttributes());
+            writeLevel(
+                    path,
+                    0,
+                    levels.get(0),
+                    server,
+                    executorService,
+                    progress -> {
```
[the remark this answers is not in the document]
petebankhead: Actually, an implementation that lots sensible percentages only (e.g. `10%`, `20%`...) would be useful - but probably belongs somewhere else, since it would be very general.

petebankhead (changes requested):
This works for me, but I find it a bit confusing that the API is quite different for `OMEZarrWriter` and `PyramidalOMEZarrWriter`, even when it seems like they should be doing something similar.

For example,
* `OMEZarrWriter.Builder` requires downsamples to be passed in a method, whereas `PyramidalOMEZarrWriter` requires them to be passed to the constructor; if they aren't needed at construction, then I think defaulting to a downsample of 1 would be fine.
* `OMEZarrWriter.writeImage` takes no arguments, uses a background thread and returns immediately, while `PyramidalOMEZarrWriter` requires a path and is blocking.

Adding example scripts to the PR showing how the writers *should* be used would be good. I'd expect the scripts to look almost identical, so that it's easy to switch from one implementation to the other. If that's not possible, we'll also need to explain that clearly on ReadTheDocs.

I've added some comments on specific things.

qupath-extension-bioformats/src/main/java/qupath/lib/images/writers/ome/zarr/PyramidalOMEZarrWriter.java
Rylern: > Is it possible for the downsamples to be out of order - or should they always be ascending?

They should always be ascending because levels are written from the highest to the lowest resolution.

> Can or should we be checking for invalid or duplicate values?

Yes, to avoid writing invalid or duplicated levels.

> Is it necessary to start with 1? I could imagine exporting with [2, 4, 8].

Yes. I tried to write an image with downsamples `[2, 6]`, but then when reading it I got downsamples `[1, 4]`. I am not sure what happens but for clarity I think it's best to always start with 1.

> Requiring a downsample of 1 is ok if necessary

The condition on the input server was removed.

Rylern:
It is now possible to provide downsamples without including 1.

PR #1967, opened 2025-08-08
Fixing tests where n might potentially be NULL

zindy:
Hello,

This is to solve a very unlikely edge case where the variable `n` is NULL in two tests (as far as I can tell) in the Brightness & Contrast dialog, `n` being the number of channels. This can happen because the dialog can be opened even if no image is available in the current viewer.

Apparently a classic *autounboxing null problem* (?) in Java:  When `n` is null, this causes a NullPointerException because Java tries to invoke `n.booleanValue()` on a null reference.

I only noticed this because my [autodock script](https://forum.image.sc/t/script-for-docking-any-floating-window-to-the-analysis-pane-results-may-vary/103808/2) crashed with
```
[JavaFX Application Thread]	ERROR	qupath.lib.gui.QuPathUncaughtExceptionHandler	Cannot invoke "java.lang.Boolean.booleanValue()" because "n" is null	java.lang.NullPointerException: Cannot invoke "java.lang.Boolean.booleanValue()" because "n" is null
	at qupath.lib.gui.commands.display.BrightnessContrastChannelPane.lambda$new$0(BrightnessContrastChannelPane.java:149)
	at com.sun.javafx.binding.ExpressionHelper$SingleChange.fireValueChangedEvent(ExpressionHelper.java:192)
	at com.sun.javafx.binding.ExpressionHelper.fireValueChangedEvent(ExpressionHelper.java:91)
	at javafx.beans.binding.ObjectBinding.invalidate(ObjectBinding.java:192)
```
Cheers,
Egor

### DIFF
--- a/qupath-gui-fx/src/main/java/qupath/lib/gui/commands/display/BrightnessContrastChannelPane.java
+++ b/qupath-gui-fx/src/main/java/qupath/lib/gui/commands/display/BrightnessContrastChannelPane.java
@@ -146,7 +146,7 @@ public class BrightnessContrastChannelPane extends BorderPane {
         filteredChannels.predicateProperty().bind(filter.predicateProperty());
         table.setItems(filteredChannels);
         table.sceneProperty().flatMap(Scene::windowProperty).flatMap(Window::showingProperty).addListener((v, o, n) -> {
-            if (n) updateShowTableColumnHeader();
+            if (n != null && n) updateShowTableColumnHeader();
         });
         initializeShowAllCheckbox();
         initializeColorPicker();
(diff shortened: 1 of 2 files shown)


petebankhead:
Would `orElse(Boolean)` solve the problem? Something like
```java
    table.sceneProperty().flatMap(Scene::windowProperty).flatMap(Window::showingProperty).orElse(Boolean.FALSE).addListener((v, o, n) -> {
            if (n) updateShowTableColumnHeader();
        });
```
I didn't actually know of the `orElse` method until just now, but it feels like a better way to handle potential nulls here... if it works for you.

petebankhead:
Ah good, could you update the PR for that please?

I couldn't see a way to trigger the issue in QuPath without your script, so I think under 'normal' circumstances the null check might not be needed. I think the `orElse` is less likely to be removed by some future over-enthusiastic attempt at code cleanup.

PR #1969, opened 2025-08-08
Make moveToTrash public and improve documentation

Rylern:
Make `GeneralTools.moveToTrash()` public and improve `GeneralTools.deleteFile()` documentation

### DIFF
--- a/qupath-core/src/main/java/qupath/lib/common/GeneralTools.java
+++ b/qupath-core/src/main/java/qupath/lib/common/GeneralTools.java
@@ -24,6 +24,7 @@
 package qupath.lib.common;
 
 import java.awt.Desktop;
+import java.awt.GraphicsEnvironment;
 import java.io.BufferedReader;
 import java.io.File;
 import java.io.IOException;
@@ -757,14 +758,20 @@ public final class GeneralTools {
 
 
 	/**
-	 * Delete a file, optionally requesting that it be moved to the trash rather than permanently deleted.
+	 * Delete a file or an empty directory, optionally requesting that it be moved to the trash rather than permanently deleted.
 	 * <p>
-	 * Note that the behavior of this method is system-dependent, and there is no guarantee the file will 
+	 * Note that the behavior of this method is system-dependent, and there is no guarantee the file/empty directory will
 	 * indeed be moved to the trash.
 	 * 
-	 * @param fileToDelete
-	 * @param preferTrash
-	 * @return true if the file is successfully deleted, false otherwise
+	 * @param fileToDelete the file or empty directory to delete
+	 * @param preferTrash whether an attempt should be made to move the provided file/empty directory to trash before deleting it
+	 * @return true if the provided file/empty directory is successfully deleted, false otherwise
+	 * @throws java.awt.HeadlessException if the preferTrash parameter and {@link GraphicsEnvironment#isHeadless()} are true
+	 * @throws UnsupportedOperationException if the preferTrash parameter is true and the {@link Desktop} class is not supported
+	 * on the current platform
+	 * @throws NullPointerException if the provided file/empty directory is null
+	 * @throws SecurityException if the calling thread doesn't have the permission to read or delete the provided file/directory
+	 * @throws IllegalArgumentException if the provided file/directory doesn't exit
 	 */
 	public static boolean deleteFile(File fileToDelete, boolean preferTrash) {
 		if (preferTrash && Desktop.isDesktopSupported()) {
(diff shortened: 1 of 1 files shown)


qupath-core/src/main/java/qupath/lib/common/GeneralTools.java
```diff
@@ -757,14 +758,20 @@ public static boolean isWindows() {
...
-	 * @return true if the file is successfully deleted, false otherwise
+	 * @param fileToDelete the file or empty directory to delete
+	 * @param preferTrash whether an attempt should be made to move the provided file/empty directory to trash before deleting it
+	 * @return true if the provided file/empty directory is successfully deleted, false otherwise
+	 * @throws java.awt.HeadlessException if the preferTrash parameter and {@link GraphicsEnvironment#isHeadless()} are true
+	 * @throws UnsupportedOperationException if the preferTrash parameter is true and the {@link Desktop} class is not supported
+	 * on the current platform
+	 * @throws NullPointerException if the provided file/empty directory is null
+	 * @throws SecurityException if the calling thread doesn't have the permission to read or delete the provided file/directory
+	 * @throws IllegalArgumentException if the provided file/directory doesn't exit
```
petebankhead, line 774: Exist?

qupath-core/src/main/java/qupath/lib/common/GeneralTools.java
```diff
@@ -775,7 +782,18 @@ public static boolean deleteFile(File fileToDelete, boolean preferTrash) {
...
-	private static boolean moveToTrash(Desktop desktop, File fileToDelete) {
+	/**
+	 * Attempt to move the provided file or directory to trash. This may fail depending on the current platform.
+	 *
+	 * @param desktop the desktop to use when moving the provided file/directory to trash
+	 * @param fileToDelete the file or directory to move to trash
+	 * @return whether the provided file/directory was moved to trash
+	 * @throws NullPointerException if one of the provided parameters is null
+	 * @throws SecurityException if the calling thread doesn't have the permission to read or move the provided file/directory
+	 * @throws UnsupportedOperationException if the provided desktop doesn't support the {@link Desktop.Action#MOVE_TO_TRASH} action
```
petebankhead, line 793: Shouldn't we just return `false` here?

I think a public version of this method shouldn't expose that it is using `Desktop` at all. For example, in the future we might switch to [JNA Platform](https://javadoc.io/doc/net.java.dev.jna/jna-platform/latest/com/sun/jna/platform/FileUtils.html#moveToTrash(java.io.File...)).

We could therefore have
```java
private static boolean moveToTrash(Desktop desktop, File fileToDelete) // as before

public static boolean moveToTrash(File fileOrDirectory) // new public method
```

qupath-core/src/main/java/qupath/lib/common/GeneralTools.java
```diff
@@ -775,7 +782,18 @@ public static boolean deleteFile(File fileToDelete, boolean preferTrash) {
...
 
-	private static boolean moveToTrash(Desktop desktop, File fileToDelete) {
+	/**
+	 * Attempt to move the provided file or directory to trash. This may fail depending on the current platform.
+	 *
+	 * @param desktop the desktop to use when moving the provided file/directory to trash
+	 * @param fileToDelete the file or directory to move to trash
+	 * @return whether the provided file/directory was moved to trash
+	 * @throws NullPointerException if one of the provided parameters is null
+	 * @throws SecurityException if the calling thread doesn't have the permission to read or move the provided file/directory
```
petebankhead, line 792: Don't think we need to document `SecurityException`. It is dropped from the [javadoc in Java 24](https://docs.oracle.com/en/java/javase/24/docs/api/java.desktop/java/awt/Desktop.html#moveToTrash(java.io.File)), presumably due to [`SecurityManager` being retired](https://inside.java/2024/12/11/quality-heads-up/).

qupath-core/src/main/java/qupath/lib/common/GeneralTools.java
```diff
@@ -757,14 +758,20 @@ public static boolean isWindows() {
...
+	 * Note that the behavior of this method is system-dependent, and there is no guarantee the file/empty directory will
 	 * indeed be moved to the trash.
 	 * 
-	 * @param fileToDelete
-	 * @param preferTrash
-	 * @return true if the file is successfully deleted, false otherwise
+	 * @param fileToDelete the file or empty directory to delete
+	 * @param preferTrash whether an attempt should be made to move the provided file/empty directory to trash before deleting it
+	 * @return true if the provided file/empty directory is successfully deleted, false otherwise
+	 * @throws java.awt.HeadlessException if the preferTrash parameter and {@link GraphicsEnvironment#isHeadless()} are true
```
petebankhead, line 769: I don't think the fact we are using `Desktop` should be exposed to the caller. I would tend towards throwing fewer exceptions and using the return value to handle failures - logging the reason (although I think NPE and `IllegalArgumentException` make sense to keep).

Issue #2000, opened 2025-09-29
Create Thresholder window too small


finglis:
* [x ] I've definitely found a bug (it you're not sure, please use [image.sc](https://forum.image.sc/tags/qupath) instead)
* [ x] I've checked https://qupath.github.io for a new release that might already have fixed the issue
* [ x] I've checked the [Changelog](https://github.com/qupath/qupath/blob/master/CHANGELOG.md) to see if the bug has already been fixed in the next release
* [ x] I've checked for existing GitHub issues describing the same problem

### Describe the bug
On windows, when opening _Classify → Pixel classification → Create thresholder..._ the window for the tool is not high enough to contain all the content. Additionally, trying to resize it will give the cursor change to the arrows when hovered over, but clicking and dragging doesn't work as you would expect. The window isn't resizable and there is no scroll bar.  

### To Reproduce
1. Go to _Classify → Pixel classification → Create thresholder..._
2. Click on bottom of the window and drag down
3. The window doesnt resize, nor is there a scroll option

### Expected behavior
Window is either resizable or contains all content when opening

### Screenshots

<img width="388" height="413" alt="Image" src="https://github.com/user-attachments/assets/98945d74-fcb2-4e16-b088-9de88451a313" />

### Desktop (please complete the following information):
 - OS: Windows 10
 - v0.6.0

PR #2003, opened 2025-09-30
Create thresholder window height issue

finglis:
As described in issue #2000, the windows height was restricted and prevented access to key buttons. 
This PR removes the max height and width for this window to match the behaviour used by other similar windows throughout QuPath.


petebankhead:
Looks fine to me - behaviour remains ok on the Mac. It's possible to resize the dialog too big but... well... no one makes anyone do that.

On an affected system, does this cause the dialog to open with the buttons visible? Or does it only make it possible to see them by resizing the window, but they are still cropped initially?

Also pinging @alanocallaghan who I think introduced the max size limit initially in https://github.com/qupath/qupath/commit/50c044b2ee53b47bd09223b19ce0de572e83e915

Issue #2001, opened 2025-09-29
Separator not needed in project tab


finglis:
### Describe the bug
When an image is not selected in the project, the right click options in this area have an additional line separation that is not required.
Right click on the project when an image is selected gives you many options. When an image is not selected, then the options is less but the separation line still remains at the very top, looking slightly odd. 

### To Reproduce
1. Make sure an image is not open or selected in an open project
2. Right click on the project tab
3. Note the additional line at the top thats not required

### Expected behavior
No line should appear when an image is not selected. The line should remain when an image is as there will be additional options that do not relate to the ones below. 

### Screenshots

<img width="2046" height="1150" alt="Image" src="https://github.com/user-attachments/assets/59875611-ae1d-43a7-8822-98fa09d7abec" />

### Desktop (please complete the following information):
 - OS: All
 - QuPath Version: 0.6.0

PR #2006, opened 2025-10-01
Removal of separator within right click menu

finglis:
Addresses #2001


alanocallaghan:
It's a fairly dumb regex, so even "does not fix" will probably close the issue. Yes, I prefer "addresses" or "relates to" when the issue will linger on, although any mention will link the issue

PR #2026, opened 2025-10-29
Clean up cell detection and fail fast on invalid input

petebankhead:
This firstly (conservatively) cleans up `WatershedCellDetection` while trying to avoid any change in behavior.

Then it adds new logic to `parseArgument(ImageData<BufferedImage> imageData, String arg)` that attempts to fail fast with an exception in the following cases:
* A brightfield image is processed with non-brightfield parameters
* A non-brightfield image is processed with brightfield parameters
* A non-brightfield image is processed while requesting a channel that doesn't exist

It *is* still possible to process a brightfield image while specifying a channel (stain) that doesn't exist, but I think that might be obscure enough to accept - and it will result in a warning being logged.

The main problem this aims to solve is that the user could use completely invalid parameters and still get sensible results because default channels were chosen.

See also https://github.com/qupath/qupath/pull/2025

---

Along the way, I also changed
```java
ParameterList.updateParameterList(ParameterList params, Map<String, String> mapNew, Locale locale)
```
so that it now throws an exception (rather than only logging a warning) if attempting to set a parameter that doesn't exist. This could potentially impact other commands, but I *hope* that if it does then its impact is positive.

This change could however be reverted, since in the end the cell detection adjustments don't rely upon it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ Version in progress...
 * Improve toggling between point and multipoint tools (https://github.com/qupath/qupath/pull/1972)
 * Update to Deep Java Library 0.35.0 (supports PyTorch 2.7.1)
 * The maximum memory preference now uses a percentage, not a number of GB
-* Cell detection is more strict about valid inputs to reduce risk of unexpected behavior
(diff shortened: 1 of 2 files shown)


alanocallaghan (approved):
I don't really understand a lot of the code still but this seems sensible and certainly resolves the worst examples of silently running on a different channel without even logging a warning.

PR #2033, opened 2025-11-07
Update use of input display

petebankhead:
Draft PR to explore the changes introduces in https://github.com/qupath/qupath-fxtras/pull/63

If accepted, then a new release of qupath-fxtras should be made (this currently uses a snapshot).

The main change is that the input display now makes key presses visible for longer, having them gentle fade out rather than disappear immediately after keys are released.

This is actually really fiddly to try to get behaving well, because a combination like `Ctrl + Shift + C` corresponds to multiple events, and so there needs to be some kind of 'memory' to avoid retaining only the last key that was released.

Changes to behavior would need to be made in qupath-fxtras.

There are also some new options for `InputDisplay`, which can be controlled via scripting:

```groovy
guiscript=true

var id = getQuPath().getInputDisplay()
id.show()

// New properties
id.setShowSymbols(false) // Unicode keyboard symbols (where known)
id.setFadeTo(0.5)        // Opacity of keypresses after fading
id.setFadeDuration(javafx.util.Duration.millis(1000)) // Duration of fade
id.setSkipTextInputControls(true) // Don't show keypresses in text controls
                                  // (the script editor ignores this...)
```

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -47,7 +47,7 @@ openslide       = "4.0.0.6"
 
 picocli         = "4.7.7"
 
-qupath-fxtras   = "0.3.0-SNAPSHOT"
+qupath-fxtras   = "0.3.0"
 qupath-training = "0.1.0"
 qupath-djl      = "0.4.0"
 
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -47,7 +47,7 @@ openslide       = "4.0.0.6"
 
 picocli         = "4.7.7"
 
-qupath-fxtras   = "0.2.1"
+qupath-fxtras   = "0.3.0-SNAPSHOT"
 qupath-training = "0.1.0"
 qupath-djl      = "0.4.0"
 
--- a/qupath-gui-fx/src/main/java/qupath/lib/gui/QuPathGUI.java
+++ b/qupath-gui-fx/src/main/java/qupath/lib/gui/QuPathGUI.java
@@ -23,6 +23,127 @@
 
 package qupath.lib.gui;
 
+import com.google.common.collect.BiMap;
+import com.google.common.collect.HashBiMap;
+import ij.IJ;
+import javafx.application.Application;
+import javafx.application.HostServices;
+import javafx.application.Platform;
+import javafx.beans.binding.Bindings;
+import javafx.beans.binding.BooleanBinding;
+import javafx.beans.binding.StringBinding;
+import javafx.beans.property.BooleanProperty;
+import javafx.beans.property.ObjectProperty;
+import javafx.beans.property.ReadOnlyBooleanProperty;
+import javafx.beans.property.ReadOnlyObjectProperty;
+import javafx.beans.property.SimpleBooleanProperty;
+import javafx.beans.property.SimpleObjectProperty;
+import javafx.beans.value.ChangeListener;
+import javafx.beans.value.ObservableValue;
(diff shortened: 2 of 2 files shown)


alanocallaghan:
I really, really like how this *should* work, but... when I open it and hit Ctrl+L, I just see Ctrl. Other modifiers work as expected.

I notice that Super (Windows key/apple key, I think...?) doesn't show up as a modifier. I guess it's unlikely we ever handle this, but just noting it.

petebankhead:
> I really, really like how this should work, but... when I open it and hit Ctrl+L, I just see Ctrl. Other modifiers work as expected.

Hmmm, Ctrl+L (Windows) and Cmd+L (Mac) and working for me as expected -  but I haven't tried Linux.

Do you mean Ctrl+L *specifically* fails, or Ctrl fails no matter what letter it is paired with? If it's just Ctrl+L then I guess it's a tolerable (and perhaps Linux-specific?) window-focussing issue.

Windows key [probably just needs to be added here](https://github.com/qupath/qupath-fxtras/blob/9104072bcd365d9b5bc57911dea98827bd4b8764/src/main/java/qupath/fx/controls/InputDisplay.java#L113).

alanocallaghan:
I *think* we shouldn't use the system/super key anyway, as traditionally it is for system-level shortcuts, and we don't currently support user keymaps.

The control issue is odd, the line
```
currentKeys.put(code, getText(event));
```
collects normal text when alt/shift/etc are depressed, but when ctrl is down, the text gets mangled, probably by some Linux quirk, so the current key map looks like this:

```
{A=}
```
or for multiple keys:
```
{A=, D=}
```

This then gets displayed as (what looks like?) a blank space in the input display.

I'll try solve this as I suspect you are not as keen to understand Linux weirdness as I am

alanocallaghan:
I think Super == Windows so should be alright

petebankhead:
I was using `includeBuild` successfully here. You’ll need that or your strategy to check any additional changes, but since our PR is now in the latest maven snapshot then using `./gradlew run --refresh-dependencies` should be enough to test it (and `--refresh-dependencies` might not be needed).

alanocallaghan:
Looks good, but num lock for me has this value (should be 144), which is "not blank" but does show up as an invisible block:


image


Could modify the if statement that returns `code.getName();` to guard against invalid chars generally, but I'm tempted to ignore this for the time being as the numbers still show up on the input display with num lock on...

```
        private String getText(KeyCode code) {
            if (code.isLetterKey())
                return code.getName().toUpperCase();
            else {
                if (showSymbolsProperty.get()) {
                    var symbol = getSymbol(code);
                    if (symbol != null)
                        return symbol;
                }
                if (code.isModifierKey() || code.getChar().isBlank() || code.isFunctionKey())
                    return code.getName();
                else
                    return code.getChar();
            }
        }
```

Rylern (approved):
Looks good on Linux (Fedora) and MacOS

finglis (approved):
Some issues pointed out in call:
1. Windows specific - scroll arrow remains even after release (but will eventually randomly disappear)
2. When pressing a combo (eg ctrl + shift + c) and then pressing another keyboard key after. It looks like the ctrl + shift was pressed in combination with it instead (eg v on its own looks like ctrl + shift + v).
3. Shift + numbers will not show the punctuation key that corresponds, only that shift + number has been pressed. Possibly not ideal for cross platform demonstrating since shift +2 is not the same output on windows and mac.

Issue #2005, opened 2025-10-01
ImageJ Roi names are ignored during import


petebankhead:
### Describe the bug
When importing objects an ImageJ TIFF with an overlay, the ROI name is no longer used in QuPath v0.6.0.
It *was* used to set the annotation name in QuPath v0.5.

See https://forum.image.sc/t/different-behaviour-when-importing-objects-from-file-in-qupath-v0-5-1-vs-v0-6-0/116784

### To Reproduce
1. Create a project
2. Import the example image from the linked forum post
3. Ensure that 'Import objects' is selected
4. See that the annotations do not have names set

### Expected behavior
The QuPath annotations have names that match the ImageJ Rois.

### Screenshots
See forum post.

### Desktop (please complete the following information):
 - OS: All
 - QuPath Version: v0.6.0

### Additional context
This is a tricky one because the decision *not* to use `Roi.getName()` was intentional - see https://github.com/qupath/qupath/commit/ebeb854a6c3f778990290486207150c8313feeab

The reasoning seems to be concerned mostly with ROIs that are passed from QuPath and then retrieved later: we set the name so that it appears in ImageJ, but we don't want to then grab that back into QuPath. If we did, then the simple act of sending a ROI to ImageJ and getting it back again would result in an object having a new name on the QuPath side.

In the specific case of importing objects, we could easily fix things by making a change at https://github.com/qupath/qupath/blob/58acb8ed40c6c0100e1bcb24ce6cfffc07eeb03d/qupath-core-processing/src/main/java/qupath/imagej/images/servers/ImageJServer.java#L323

This *doesn't* handle the case of importing ROIs from an existing image other ways though, and we need to check what happens when the name is set as part for a serialized `.roi` file or ` RoiSet.zip`.

PR #2034, opened 2025-11-07
Improve conversion of QuPath objects to/from ImageJ Rois

petebankhead:
- Fixes https://github.com/qupath/qupath/issues/2005 by providing a way to distinguish between Roi names that are set on the QuPath side from those that are set elsewhere.
- In `IJTools`, return `SortedMap<Integer, PathObject` instead of `SortedMap<Number, PathObject>` because it turns out that `Number` does not implement `Comparable`.
- Add `IJTools.setObjectType` methods to standardize how an object type can be saved as a `Roi` property.


alanocallaghan:
> Number does not implement Comparable

My gut reaction to this was one of exasperated confusion but it does make sense given the existence of complex numbers, NaN, etc.

PR #2035, opened 2025-11-07
Log an exception in ImageServerProvider for debugging

alanocallaghan:
I found this change useful when debugging the CZI extension.

Either the log message should be at a lower level (eg, debug) to signal that it's not unusual behaviour, or we should be propagate the full exception into the log.


alanocallaghan:
Yeah that alternative sounds good. I did consider that in usual use a server not working may be fine, but then imo the warning should be at debug level anyway. However we do need some way to propagate errors when it's needed (as you point out, when nothing works).

alanocallaghan:
I'll close this and implement the alternative proposal unless you wish to do it

PR #2036, opened 2025-11-07
Log at warn only if all supports failed

alanocallaghan:
From discussion in #2035. This improves the logging for image server providers, only warning if no supported servers can open and then warn logging all exceptions encountered, and debug logging exceptions in the normal case, where advanced users/developers may want to view the exceptions but in normal use they are irrelevant.


petebankhead:
Looks fine to me, although I can't figure out a way to have it called - any suggestions?

Seems safe enough to just merge, but I haven't been able to check what is actually logged. When I attempt to open a .czi image on my Apple Silicon Mac it's travelling down some other code path, so I don't see how the output from this looks.

alanocallaghan:
I ran into it by declaring an image server that claims to have top tier support for czi files but doesn't actually work 🙈 I'll retest that and report back

PR #2038, opened 2025-11-12
Reduce navigation speed using arrow keys in viewer

petebankhead:
This makes panning easier using the arrow keys by decreasing the effect of a single keypress - which previously could pan too far.

I found the old behavior *deeply* annoying when trying to pan to a precise location, but it turns out it can already be adjusted with a couple of preferences. So this PR just changes one of them.

Pressing and holding an arrow still makes it possible to gradually accelerate when wanting to move further (at least on macOS) - although there's a preference for this too..

You don't need to build this PR to check it - in QuPath v0.6.0, just reduce the preference value from `100` to `10`:

Screenshot 2025-11-12 at 11 12 34

The navigation code is horribly complicated and should probably be revisited, but to me this change is the simplest way to get more intuitive default behavior.

### DIFF
--- a/qupath-gui-fx/src/main/java/qupath/lib/gui/prefs/PathPrefs.java
+++ b/qupath-gui-fx/src/main/java/qupath/lib/gui/prefs/PathPrefs.java
@@ -23,6 +23,31 @@
 
 package qupath.lib.gui.prefs;
 
+import javafx.beans.binding.BooleanBinding;
+import javafx.beans.property.BooleanProperty;
+import javafx.beans.property.DoubleProperty;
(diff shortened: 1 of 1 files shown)


finglis (approved):

alanocallaghan (approved):
It's better, but on Linux at least the behaviour when holding is to go medium speed -> slow speed -> faster, and it's hard to stop things before the viewer shoots off into space.

Aside from my fixing this (presumably platform-specific) weirdness in a future release, we could expose an acceleration rate parameter in future also.

Rylern (approved):

PR #2065, opened 2026-01-13
Flush measurement writer after each row

alanocallaghan:
While digging into the measurement code to see about other formats, I noticed we don't flush the output writer. Surely this is why export fails for huge numbers of measurements...? This way we only need to hold a single row in memory, whereas otherwise I'd guess there's one huge string being formed. I'd expect text files to be written line by line anyways.

### DIFF
--- a/qupath-gui-fx/src/main/java/qupath/lib/gui/tools/MeasurementExporter.java
+++ b/qupath-gui-fx/src/main/java/qupath/lib/gui/tools/MeasurementExporter.java
@@ -37,7 +37,6 @@ import qupath.lib.objects.TMACoreObject;
 import qupath.lib.projects.ProjectImageEntry;
 
 import java.awt.image.BufferedImage;
-import java.io.BufferedOutputStream;
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


petebankhead:
> Surely this is why export fails for huge numbers of measurements...? 

I wasn't aware that it does - do you have any examples where this has been an issue?

Putting together code from various lines, it looks like we have
```java
new PrintWriter(new OutputStreamWriter(new BufferedOutputStream(new FileOutputStream(file))))
```
For `OutputStreamWriter` a [buffer will already be used](https://docs.oracle.com/javase/8/docs/api/java/io/OutputStreamWriter.html), and if you really wanted to flush a `PrintWriter` after each line it would make sense to [use the constructor with `autoFlush`](https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/io/PrintWriter.html#%3Cinit%3E(java.io.Writer,boolean)).

So I'm not sure if this solves a real or predicted problem, and I'm not sure it's the right (simplest, most performant) solution.

The `OutputStreamWriter` javadocs suggests wrapping within a `BufferedWriter` - so if looking to improve this, then it'd be worth seeing if we should have a more optimal use of streams/writers.

alanocallaghan:
Also, 

> The PR description suggested it was a known failure

I saw that forum post before and I thought we had also discussed the issue with many measurements reported by Ava/Hugh previously. When I then ran into the issue myself, I thought "Oh, I guess I've hit the limit that we are somewhat aware of"

Issue #2028, opened 2025-10-29
Add a flip image parameter


Rylern:
## Feature request

Currently, it is only possible to flip an image with a [script](https://forum.image.sc/t/flipping-an-image-in-qupaths-gui/85110/4).

It would be better if there was an option to do it when importing images to a project, for example below `Rotate image` on this dialog:

<img width="1010" height="1506" alt="Image" src="https://github.com/user-attachments/assets/acc977d1-3cef-47e4-ad9f-a1421ef2349f" />

The new parameter could be called `Flip image` and have a dropdown to select between ` `, `X axis`, and `Y axis`.

The implementation could reuse the script mentionned above, or a new `FlippedImageServer` which may be more efficient.

See https://forum.image.sc/t/flipping-an-image-in-qupaths-gui/85110 for more context.

PR #2075, opened 2026-01-21
Add a flip parameter on ProjectImportImagesCommand

Rylern:
Implements #2028:

image

image


Rylern:
On [LuCa-7color_Scan1.qptiff or Patient_1.ome.tiff](https://qupath.readthedocs.io/en/0.6/docs/intro/acknowledgements.html), the [benchmark script](https://gist.github.com/petebankhead/ea559265bb5e87ede3eb9c700b26c584) results are:
* Around 21 seconds with no rotation.
* Around 30.3 seconds with the following rotation:
```groovy
server = new RotatedImageServer(server, RotatedImageServer.Rotation.ROTATE_180)
```
* Around 35 seconds with the following rotation:
```groovy
var transform = new AffineTransform()
transform.rotate(Math.PI, server.getWidth() / 2, server.getHeight() / 2)
server = new TransformedServerBuilder(server)
    .transform(transform)
    .build()
```

So yes it might be better to create a new `FlipImageServer`.

PR #2113, opened 2026-03-27
Improve speed for object/cell measurements

petebankhead:
Cell measurements can be *extremely* slow when running InstanSeg with highly multiplexed images - slower than running InstanSeg itself.

## Previously

The key code is in `ObjectMeasurements`, which is where the cell measurements were made.

Previously, it was possible to only measure one cell at a time. In a script, the code might be parallelized as
```groovy
import qupath.lib.analysis.features.*

var server = getCurrentServer()
var cells = getCellObjects()
double downsample = 1.0
var measurements = ObjectMeasurements.Measurements.values() as List
var compartments = ObjectMeasurements.Compartments.values() as List

println "Num cells: ${cells.size()}"

cells.parallelStream().forEach { c ->
    ObjectMeasurements.addIntensityMeasurements(server, c, downsample, measurements, compartments)        
}
```

One reason this was slow was that it would involve creating a cropped image for every single cell, then measuring each channel separately.

## This PR
This PR aims to improve things in several ways. The main ones:

1. Permit requesting measurements for lists of cells. The code will then partition these into batches of cells that are near to one another, so as to reduce the number of image requests.
2. When queried cells all fall within a single tile, grab the whole tile - because this avoids extra (slow) image cropping.
3. When measuring the same (labeled) mask across multiple channels, create an array containing all the non-zero pixels when processing the first channel. Previously, we needed to iterate through *all* pixels of the mask for *all* channels and *all* compartments. This can drastically cut down the number of pixels we need to visit. 

## Example

```groovy
import qupath.lib.analysis.features.*

var server = getCurrentServer()
var cells = getCellObjects()
double downsample = 1.0
var measurements = ObjectMeasurements.ALL_MEASUREMENTS
var compartments = ObjectMeasurements.ALL_COMPARTMENTS

println "Num cells: ${cells.size()}"

try (var pool = java.util.concurrent.Executors.newFixedThreadPool(6)) {
    ObjectMeasurements.addIntensityMeasurements(server, cells, downsample, measurements, compartments, pool)    
}
```

It's also possible to use
```groovy
ObjectMeasurements.addIntensityMeasurements(server, cells, downsample, measurements, compartments)    
```
but currently this won't use multiple threads for any aspect. It may still be faster than the older (parallelized) code.

## Results
For ~28,000 cells in the LuCa 3x3 image, this code reduced processing time from >20 seconds to ~5 seconds on my older M1 MacBook Pro. The image has 8 channels.

I think / hope / expect that any speedup will be more noticeable on images with more channels.

## Still to do
We should figure out how best to test this, and ensure that the same results are obtained as before.

Then we should merge the PR and look to include the improvements in the InstanSeg and StarDist extensions.

---

## Extra info

Initially, I used a `QuadTree` for the spatial cache.

Switching to `STRTree` helped *enormously*... because the `QuadTree` was returning too much on early queries, and I wasn't checking the results really *did* overlap with the queried region.

This may be work investigating if there are other performance-critical parts of QuPath where `QuadTree` is used.

### DIFF
--- a/qupath-core-processing/src/main/java/qupath/lib/analysis/features/ObjectMeasurements.java
+++ b/qupath-core-processing/src/main/java/qupath/lib/analysis/features/ObjectMeasurements.java
@@ -31,7 +31,6 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
-import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
@@ -51,7 +50,7 @@ import org.locationtech.jts.geom.Lineal;
 import org.locationtech.jts.geom.Polygon;
 import org.locationtech.jts.geom.Polygonal;
 import org.locationtech.jts.geom.util.AffineTransformation;
-import org.locationtech.jts.index.quadtree.Quadtree;
+import org.locationtech.jts.index.strtree.STRtree;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import qupath.imagej.tools.IJTools;
@@ -185,7 +184,7 @@ public class ObjectMeasurements {
 
 
 	private static <T> Set<T> toOrderedUnmodifiableSet(T[] elements) {
-        var set = new LinkedHashSet<T>(Arrays.asList(elements));
+        var set = new LinkedHashSet<>(Arrays.asList(elements));
 		return Collections.unmodifiableSet(set);
 	}
 	
@@ -529,7 +528,7 @@ public class ObjectMeasurements {
 																double downsample) {
 
 		// Create a set of objects that have ROIs (probably all of them, but good to be sure)
-		Set<PathObject> objectsToMeasure = pathObjects.stream().filter(PathObject::hasROI).collect(Collectors.toCollection(HashSet::new));
+		Set<PathObject> objectsToMeasure = pathObjects.stream().filter(PathObject::hasROI).collect(Collectors.toCollection(LinkedHashSet::new));
 
 		// See if a single request will do the job
 		int pad = (int)Math.ceil(downsample * 2);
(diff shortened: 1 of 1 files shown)


qupath-core-processing/src/main/java/qupath/lib/analysis/features/ObjectMeasurements.java
```diff
@@ -396,96 +404,394 @@ private static void addShapeMeasurements(MeasurementList ml, Geometry geom, Stri
...
+		}
+
+		// Batch together nearby objects... we could try to optimize this cleverly, but it's probably not worth it
+		int batchTileSize = (int)Math.round(1024 * downsample);
+		for (int y = globalRequest.getMinY(); y < globalRequest.getMaxY(); y += batchTileSize) {
+			for (int x = globalRequest.getMinX(); x < globalRequest.getMaxX(); x += batchTileSize) {
+				var tempRequest = RegionRequest.createInstance(
+						server.getPath(), downsample, x, y, batchTileSize, batchTileSize, globalRequest.getImagePlane()
+				);
+				@SuppressWarnings("unchecked") // Quadtree isn't typed - but we know what we put in
```
alanocallaghan, line 561: STRtree, not quadtree?

alanocallaghan (approved):
Interestingly this code is slower on OS-2, about 15s to 20s with 270k cells. On a 70+ channel image, for 33k cells with 23 threads it goes from 3mins to 13s, which is massive.

The code looks good to me

Unrelated I assume, but I keep managing to corrupt my bioformats readers which seems to make reading tiles impossible. Presumably I could re-open the image to refresh the reader pool before it manages to crash. Is this a known issue, or should I investigate?

petebankhead:
Thanks, I'm curious about the slowdown but I suppose a modest slowdown for some images is tolerable when there's a massive speedup for others. OS-2 is presumably more like a single-channel request, since everything comes back in an ARGB array.

There isn't a known issue with corrupting Bio-Formats readers. There *were* issues, but these were addressed in recent-ish commits https://github.com/qupath/qupath/commits/c42e0d09a1922b4645e013a4f9f26e4c98e58f25/qupath-extension-bioformats/src/main/java/qupath/lib/images/servers/bioformats/BioFormatsImageServer.java
(Where I meant 'ensure' rather than 'unsure'...)

alanocallaghan:
Hmm. I should be using all those commits. I'll see if I can reproduce on an open licence image for a start

PR #2119, opened 2026-04-01
Enter key should confirm values in annotation properties dialog

malwilley:
Before this change, pressing enter in the annotation properties dialog did not confirm the values.

This is similar to the logic in ImageDetailsPane: https://github.com/qupath/qupath/blob/c42e0d09a1922b4645e013a4f9f26e4c98e58f25/qupath-gui-fx/src/main/java/qupath/lib/gui/panes/ImageDetailsPane.java#L634 

Before:

https://github.com/user-attachments/assets/471c5fe1-c06c-4d1a-9f62-0418996b7547

After:

https://github.com/user-attachments/assets/483f8416-7923-45db-95d1-72d0f8182891

### DIFF
--- a/qupath-gui-fx/src/main/java/qupath/lib/gui/tools/GuiTools.java
+++ b/qupath-gui-fx/src/main/java/qupath/lib/gui/tools/GuiTools.java
@@ -1061,7 +1061,7 @@ public class GuiTools {
 
 		var btnApply = dialog.getDialogPane().lookupButton(ButtonType.APPLY);
 		var enterPressed = new KeyCodeCombination(KeyCode.ENTER);
-		dialog.getDialogPane().addEventFilter(KeyEvent.KEY_PRESSED, e -> {
+		dialog.getDialogPane().addEventHandler(KeyEvent.KEY_PRESSED, e -> {
 			if (enterPressed.match(e)) {
 				btnApply.fireEvent(new ActionEvent());
 				e.consume();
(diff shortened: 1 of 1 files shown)


petebankhead:
Thanks, I tried this. It has the problem that it's no longer possible to press `Enter` when adding a description, therefore making it impossible to have multiple lines in the description.

If the workflow you want is to quickly enter a name and then close the dialog, this is already possible (at least on macOS):
* Draw an annotation
* Press enter to show the properties dialog
* Type the name
* Press `Shift`+`Tab` to move focus backwards to the 'Apply' button
* Press `Enter`

I'd prefer this approach, partly because it requires no changes but mostly because I feel like adding custom behavior like this to specific dialogs could quickly escalate and add a lot of complexity.

I think that the 'Set image type' dialog is a special case because it will confront _every_ user when they start using QuPath, _every_ time they open an image. Also, because it mostly contains buttons (and no text input controls), there is no risk of conflict.

malwilley:
@petebankhead thanks for taking a look so quickly! I didn't test the description field, my bad.

I pushed a commit that will handle both the title and description correctly, but I see your point:

> I feel like adding custom behavior like this to specific dialogs could quickly escalate and add a lot of complexity.

I'll close this PR, but it would be nice to have proper enter behavior in all the dialogs! This PR came about from watching my wife struggle through some analysis and having to resort to the mouse a little too often 😅

PR #2122, opened 2026-04-13
Reduce small fragments and holes created by wand

petebankhead:
Possible alternative to https://github.com/qupath/qupath/pull/2120

For consideration and testing; the changes it makes are quite modest, so I'm not sure if they achieve enough to justify the complexity. We'd need to be on the lookout for any performance penalty with very large geometries, but I think it should be ok.

The general idea:
* When expanding an annotation with the wand, check the holes in the ROI before and after adding the new piece.
* Remove any new holes that are thinner than a threshold determined from the viewer downsample (which is what is influencing the wand's resolution)

The 'thinness' threshold warrants some more investigation, so for now this is a draft.


petebankhead:
I'll close this, as I think it is no longer required thanks to
* https://github.com/qupath/qupath/pull/2125
* https://github.com/qupath/qupath/pull/2126

PR #2127, opened 2026-04-22
Improve methods on `PathObject` to remove child objects

petebankhead:
Fixes https://github.com/qupath/qupath/issues/2111

This also deprecates `PathObject.clearChildObjects()` and replaces it with `PathObject.removeAllChildObjects()` for consistency with `QP` since https://github.com/qupath/qupath/pull/1713

Pinging @alanocallaghan since you led that PR.

This means that we now have
```java
void removeChildObjects(Collection<? extends PathObject> pathObjects) // Remove objects in the collection

void removeAllChildObjects() // New method: Remove all the child objects

void clearChildObjects() // Deprecated method, inconsistent naming
```

Previously, we *could* call `removeChildObjects()` in a script and Groovy would merrily replace it with `removeChildObjects(null)`... which was a potential source of confusion, i.e. https://github.com/qupath/qupath/issues/2111

An alternative we could do is to have
```java
void removeChildObjects(Collection<? extends PathObject> pathObjects) // Remove objects in the collection

void removeChildObjects() // New method: Remove all the child objects... without including the word 'all'

void clearChildObjects() // Deprecated method, inconsistent naming
```
but I haven't done that, since I figure that it's better to require the calling code to be more explicit and specify when 'all' objects are intended.


alanocallaghan:
The remove object stuff looks good, but it seems like you've accidentally smuggled in [some window changes](https://github.com/petebankhead/qupath/blob/ffa6c99b8928fe2eaca48ad241da5c5f8478a4b7/qupath-gui-fx/src/main/java/qupath/lib/gui/measure/ui/SummaryMeasurementTable.java) too?

petebankhead:
Indeed I did... I think I cherry-picked my way out of the debacle adequately with https://github.com/qupath/qupath/pull/2129

PR #2135, opened 2026-05-12
Update gradle, fix geometry conversion tests (hopefully)

petebankhead:
Attempts to investigate/fix problem that surfaced in https://github.com/qupath/qupath/pull/2134
This includes Gradle and workflow updates.

The source of the problem traces back to tests that fail (at least on macOS) using Temurin 25.0.3 (and 26.0.1) but did *not* fail on 25.0.2.

They relate to converting a `java.awt.geom.Area` to a `org.locationtech.jts.geom.Geometry`.

The failures are strange. The immediately relevant code is where `AreaROI` lazily creates a `Geometry`

```java
	@Override
	public Geometry getGeometry() {
		if (geometry == null) {
			synchronized(this) {
				if (geometry == null)
					geometry = super.getGeometry();
			}
		}
		return geometry;
	}
```

This fails by creating an invalid `Geometry` on 25.0.3, but works if I create the `Geometry` every time with
```java
	@Override
	public Geometry getGeometry() {
		return super.getGeometry();
	}
```

It also works if I remove the synchronisation
```java
	@Override
	public Geometry getGeometry() {
		if (geometry == null) {
			geometry = super.getGeometry();
		}
		return geometry;
	}
```

All the above calls ultimately lead back to [`GeometryTools.convertAreaToGeometry`](https://github.com/qupath/qupath/blob/31312a5708d1c396cba0a12db306358524c1e8c0/qupath-core/src/main/java/qupath/lib/roi/GeometryTools.java#L1179) and the use of `Polygonizer`.

Because of the non-deterministic behavior, my best guess is that the bug relates to https://github.com/locationtech/jts/issues/1063

Because `AreaROI` is really a legacy class to enable ROIs to be read from *very* old QuPath releases (possibly v0.2.0) and has been deprecated since v0.6.0, so one of the more hack-y workarounds above would address the failing tests... although `GeometryTools.convertAreaToGeometry` can still be used to convert manually-draw polygons, so we'd want it to be as reliable as possible.

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       matrix:
         platform: [ubuntu-latest, macos-latest, windows-latest]
-        fail-fast: false
+      fail-fast: false
     runs-on: ${{ matrix.platform }}
     steps:
     - uses: actions/checkout@v6
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -14,6 +14,7 @@ jobs:
     strategy:
       matrix:
         platform: [ubuntu-latest, macos-latest, windows-latest]
+        fail-fast: false
     runs-on: ${{ matrix.platform }}
     steps:
     - uses: actions/checkout@v6
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -16,13 +16,23 @@ jobs:
         platform: [ubuntu-latest, macos-latest, windows-latest]
     runs-on: ${{ matrix.platform }}
     steps:
-    - uses: actions/checkout@v4
-    - uses: actions/setup-java@v4
+    - uses: actions/checkout@v6
+    - uses: actions/setup-java@v5
       with:
         java-version: 25
         distribution: temurin
     - name: Setup Gradle
-      uses: gradle/actions/setup-gradle@v4
+      uses: gradle/actions/setup-gradle@v6
 
     - name: Execute build
       run: ./gradlew build
+
+    - name: Report if failed
+      if: failure()
+      uses: actions/upload-artifact@v4
+      with:
+        name: reports
+        path: |
+          **/build/reports/
+          **/build/test-results/
+        retention-days: 7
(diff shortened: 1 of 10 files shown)


petebankhead:
As seen above, the tests fail on macOS, but pass on Ubuntu and Windows. So... whatever is causing the problem seems very obscure.

petebankhead:
So I was wrong to blame `Polygonizer` here: the issue appears to trace back to a change to `AreaOp` made in commit `ba8b995` to OpenJDK, which should address

> 8364373: Transform Affine transformations

The reason it fails on macOS only is that it is using 25.0.3, whereas on Ubuntu and Windows the tests were still run with 25.0.2.

In QuPath's code, a `Path2D` is constructed and then converted to an `Area`. Previously, this worked without any evident issue. The OpenJDK change now results in the `Area` object sometimes now having an incorrect output, as can be seed by drawing the `Area` with `Graphics2D`.

### Before

lines-correct

### After

lines-wrong

petebankhead:
The issue turns out to be really problematic in practice, because the relevant code is used when converting a `PolygonROI` into a `GeometryROI`.

If I create a *really* complicated `PolygonROI`

image

In QuPath v0.7.0, I could start to edit part of this as a brush; lines disappeared, but the same area was represented and everything behaved as it should.

image

With Java 25.0.3, as soon as I started my edit I see

image

My guess is that it relates to how [`AreaOp` now tries to reduce memory by 'consuming subcurves'](https://github.com/openjdk/jdk/blame/ba8b9958b7d239cdabddc8b0dd7e1395827b76e9/src/java.desktop/share/classes/sun/awt/geom/AreaOp.java#L416), but it appears to fundamentally change the behavior in a way that causes us a lot of trouble.

petebankhead:
This should now be fixed, although we should still be on the lookout for any new failures.

I created a polygon that I thought might fail (due to coincident lines), but it seems to behave as it should - both with this PR and in v0.7.0.
```groovy
var p = ROIs.createPolygonROI(
    [0, 100, 100, 0, 0, 100, 100, 0, 50, 50] as double[],
    [0, 0, 100, 100, 200, 200, 100, 100, 50, 150] as double[]
)
addObject(PathObjects.createAnnotationObject(p, getPathClass("Testing")))
```

PR #2138, opened 2026-05-25
Add raster brush preference

petebankhead:
This prevents the brush from creating ROIs that contain diagonal connections between coordinates. Because coordinates are constrained to be integers, this results an ROIs that rasterize more predictably.

## Example

The top nucleus is annotated with the new preference turned **on**, while the bottom nucleus is annotated with the preference turned **off** (the previous behavior).

image

## Extra

A new 'experimental' method `GeometryTools.rasterize(Geometry)` has been added to enable any arbitrary polygonal geometry to be rasterised. *However* its performance is poor for large and complex geometries, and it requires further testing for small ones.

It exists for now as a 'better than nothing' option for use in scripts. In the end, the code for this was not used to implement the raster brush option.

### DIFF
--- a/qupath-core/src/main/java/qupath/lib/roi/GeometryTools.java
+++ b/qupath-core/src/main/java/qupath/lib/roi/GeometryTools.java
@@ -22,6 +22,7 @@
 package qupath.lib.roi;
 
 import org.locationtech.jts.algorithm.Orientation;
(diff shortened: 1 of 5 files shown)


petebankhead:
Note that the raster brush is currently **off by default**, but this may change in the future (based upon user feedback, or what we discover works better in practice).

My guess is that turning it on may be a sensible default one day.

PR #2144, opened 2026-06-12
fix: size filter on pixel-classifier object creation ignores holes

### LINKED ISSUE #2136: The size filter on object creation from pixel classifier seems to not work properly
### Describe the bug
The size filter on object creation from pixel classifier is not working properly. A minimum object area is set to 5 um^2 is set, but object of size < 5um ^2 are still remaining.

### To Reproduce
1. Add to the QuPath project the following image: https://omero-public.epfl.ch/webclient/?show=image-568578
2. Get the following pixel classifier (should also work with another one, but I made this dummy one to have a lot of small objects) [classifier-test.json](https://github.com/user-attachments/files/27807453/classifier-test.json)
3. Draw a rectangle 
4. Apply the pixel classifier on this rectangle
5. Create annotation objects out of this, with minimum object size = 5 um^2, and SPLIT OBJECTS option activated
6. Open the annotation measurements, and sort by area.
7. Objects with size < 5 um^2 are remaining.

### Expected behavior
Having all objects with area < 5um^2  removed from the analysis

### Screenshots

Image

Image

### Desktop:
 - OS: Windows
 - QuPath Version: 0.7.0

mvanhorn:
## Summary

The minimum-object-size filter in pixel-classifier object creation now measures net geometry area (outer ring minus retained holes) and runs after hole handling, so objects whose net area is below the threshold are removed even when their outer ring exceeds it.

## Why this matters

In #2136 you diagnosed the cause: the area test is applied to the outer ring of each polygon and ignores holes, so with "split objects" enabled, a 5 um^2 minimum still lets through ring-shaped fragments whose outer ring passes but whose net area does not. You proposed exactly this reordering (handle holes first, then apply the area test to what remains). The shared filtering in `RoiTools` carries the corrected semantics so the behavior is consistent, while the change is limited to the pixel-classifier creation path since you flagged that "Remove fragments & holes" carries an intentional, changelog-documented behavior from years back.

## Testing

New cases in `TestPixelClassifierTools` cover ring-shaped objects below/above the net-area threshold with and without hole removal, and `TestRoiTools` covers the net-area computation. No JVM was available in the implementation environment, so CI is the verification run for the suites.

Fixes #2136

### DIFF
--- a/qupath-core-processing/src/main/java/qupath/opencv/ml/pixel/PixelClassifierTools.java
+++ b/qupath-core-processing/src/main/java/qupath/opencv/ml/pixel/PixelClassifierTools.java
@@ -43,6 +43,7 @@ import qupath.lib.objects.hierarchy.PathObjectHierarchy;
 import qupath.lib.regions.ImagePlane;
 import qupath.lib.regions.RegionRequest;
 import qupath.lib.roi.GeometryTools;
+import qupath.lib.roi.RoiTools;
 import qupath.lib.roi.interfaces.ROI;
 import qupath.opencv.ml.pixel.PixelClassifiers.ClassifierFunction;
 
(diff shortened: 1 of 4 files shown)


petebankhead:
I am going to guess that an LLM contributed to this, and I'm afraid we don't accept or review AI-generated contributions. As I described on the issue, this is not a bug requiring a fix but rather a matter of balancing different desirable outcomes with user expectations and backwards compatibility.

Issue #2149, opened 2026-06-24
Bug report: NegativeArraySizeException when opening ARGOS AVS files via OpenSlide 4.0.1


CrackerJack10:
Hi QuPath team,

I am writing to report a bug in QuPath v0.7.0 when attempting to open ARGOS AVS whole slide images using the updated OpenSlide 4.0.1 library.

**Environment**
- QuPath version: 0.7.0 (Windows)
- OpenSlide DLL: libopenslide-1.dll from openslide-bin 4.0.1.2 (manually updated in app/openslide-4.0.0.6-win32-x86_64.jar)
- OS: Windows 11 Pro
- File format: ARGOS AVS (.avs) — newly supported in OpenSlide 4.0.1

**What happens**
OpenSlide successfully recognises and opens the .avs file (a WARN is logged that openslide.objective-power is not available, which is expected for this format). However, QuPath immediately crashes with a NegativeArraySizeException during thumbnail generation inside OpenslideImageServer.<init>, before the image can be displayed or added to a project.

The crash occurs both when importing via the project import dialog and when using the scripting API (ImageServerProvider.getPreferredUriImageSupport). It is not possible to bypass the thumbnail generation step through either route.

**Stack trace**

java.lang.NegativeArraySizeException: -1091131160
    at java.desktop/java.awt.image.DataBufferInt.<init>(Unknown Source)
    at java.desktop/java.awt.image.Raster.createPackedRaster(Unknown Source)
    at java.desktop/java.awt.image.DirectColorModel.createCompatibleWritableRaster(Unknown Source)
    at java.desktop/java.awt.image.BufferedImage.<init>(Unknown Source)
    at qupath.lib.images.servers.AbstractTileableImageServer.createDefaultRGBImage(AbstractTileableImageServer.java:254)
    at qupath.lib.images.servers.AbstractTileableImageServer.createRGBImage(AbstractTileableImageServer.java:497)
    at qupath.lib.images.servers.AbstractTileableImageServer.readRegion(AbstractTileableImageServer.java:301)
    at qupath.lib.images.servers.AbstractTileableImageServer.readRegion(AbstractTileableImageServer.java:59)
    at qupath.lib.images.servers.AbstractImageServer.getDefaultThumbnail(AbstractImageServer.java:313)
    at qupath.lib.images.servers.openslide.OpenslideImageServer.<init>(OpenslideImageServer.java:267)
[...]

PR #2150, opened 2026-06-24
Update openslide to v4.0.1.2

alanocallaghan:


petebankhead:
Looks good, I confirm this allows me to open `Argos-1.avs` on macOS.

Some tile misalignment, but suspect it is nothing we can influence - think you can update the OpenSlide version in the changelog and merge this.

image

petebankhead:
I think this is the most we can do to resolve https://github.com/qupath/qupath/issues/2149

alanocallaghan:
Yes, I noticed some popin when zooming, but agree I don't know if we can do much about it (or even openslide can to be fair, I don't know the format)

petebankhead:
The black-to-white background changes when zooming aren't ideal, but a separate issue.

image

image

The issue is that [there is no background color stored in the properties](https://github.com/qupath/qupath/blob/main/qupath-extension-openslide/src/main/java/qupath/lib/images/servers/openslide/OpenslideImageServer.java#L252).

If we default to white (instead of null, leading to black) then it displays better.

Since OpenSlide *can* show fluorescence images (but usually doesn't), I'm not sure if defaulting to white would cause compatibility problems for someone.

petebankhead:
Created an issue at https://github.com/qupath/qupath/issues/2151 so you can merge this if you're happy with it.

PR #2152, opened 2026-06-25
Support `--background-color` for OpenSlide

### LINKED ISSUE #2151: Support for custom background color property for OpenSlide images
### Is your feature request related to a problem? Please describe.
OpenSlide has a standard `openslide.background-color` property, but [it's not always present](https://openslide.org/docs/properties/).

Where it *is* available, [QuPath will use it](https://github.com/qupath/qupath/blob/main/qupath-extension-openslide/src/main/java/qupath/lib/images/servers/openslide/OpenslideImageServer.java#L252) to pre-fill the `BufferedImage` before painting the pixels on top.

The background color only makes a difference when the alpha indicates that the pixels to be drawn are transparent (i.e. the region is unscanned). 

The issue is that no background color is specified with Argos images (introduced in OpenSlide 4.0.1), and since the default is to *not* fill the `BufferedImage`... meaning that the background appears black.

This can result in weird affects when zooming, and will mess with thresholds. 

See https://github.com/qupath/qupath/pull/2150#issuecomment-4791395458

### Describe the solution you'd like
I'd like 2 things:
* Make the default fill `Color.WHITE`. Since OpenSlide is almost always used with brightfield images, this will almost always be correct... or irrelevant, since all pixels are scanned.
* Make it possible to pass a string argument to override the default color. Possibly this could take 2 forms: override the default color *always*, or override the default color *only when it isn't specified under `openslide.background-color`*. (If only one of these options is provided, it should be the second)

### Describe alternatives you've considered
1. Leave as it is. But that is bad for Argos.
2. Make the default fill white, with no option to override this.

### Additional context
I expect this will matter for `.mrxs` and *maybe* other formats (`.bif`, `.czi`?).

Making this change could conceivably cause version compatibility issues, although I'm not aware of any times when the background color is specified to be anything other than white *and* there are transparent pixels. Still, we should check the OpenSlide test data and also look for fluorescence examples.

petebankhead:
Fixes https://github.com/qupath/qupath/issues/2151

Now the background value will be chosen using the following descending order of preference:

1. An optional string argument, in the format `--background-color=black` or `--background-color white` (i.e. with or without equals; valid values also include `red`, `green`, `blue` or a hex like `#ff2288`)
2. From `openslide.background-color` (a [recognized OpenSlide property](https://openslide.org/docs/properties/)
3. White (previously it default to black, on account of no background filling being performed)

This results in Argos sample images displaying properly with white backgrounds (the original problem), while mrxs OpenSlide samples appear to continue to display correctly. However, they can be coerced into displaying badly if you like, for example by passing `--background-color red` to see what it does

image

### DIFF
--- a/qupath-extension-openslide/src/main/java/qupath/lib/images/servers/openslide/OpenslideImageServer.java
+++ b/qupath-extension-openslide/src/main/java/qupath/lib/images/servers/openslide/OpenslideImageServer.java
@@ -4,7 +4,7 @@
  * %%
  * Copyright (C) 2014 - 2016 The Queen's University of Belfast, Northern Ireland
  * Contact: IP Management (ipmanagement@qub.ac.uk)
- * Copyright (C) 2018 - 2024 QuPath developers, The University of Edinburgh
+ * Copyright (C) 2018 - 2026 QuPath developers, The University of Edinburgh
  * %%
  * QuPath is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as
@@ -24,6 +24,7 @@
 package qupath.lib.images.servers.openslide;
 
 import com.google.gson.GsonBuilder;
+import java.util.Optional;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import qupath.lib.common.GeneralTools;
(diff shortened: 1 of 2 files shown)


alanocallaghan:
Sorry, I fat fingered Copilot

alanocallaghan (approved):
Not run it but looks good, only minor thought is to catch the number format exception from Color.decode and go with an empty optional, but I think that throwing the exception is the right thing to do.

PR #2160, opened 2026-07-20
Use scaled length in ellipse shape measurement

### LINKED ISSUE #2158: 'Add shape features' wrongly gives 'Length µm' in pixels
The length generated through "Add shape features" for ellipses is incorrect. It appears to be not scaled by pixel size. Image below has a pixel size of 0.3254 um.

Image
 

Test:
1) Draw an ellipse
2) `Analyze > Add features > Add shape features`, click Length
3) "Length um" will equal "Perimeter um" / pixel size. 

To convince myself I understand what length is supposed to be, I looked through the Object Measurements code. I think maybe I found the issue? [This ](https://github.com/qupath/qupath/blob/main/qupath-core-processing/src/main/java/qupath/lib/analysis/features/ObjectMeasurements.java#L336)calls getLength(), not getScaledLength(). ObjectMeasurements leads back to [AbstractPathROI](https://github.com/qupath/qupath/blob/main/qupath-core/src/main/java/qupath/lib/roi/AbstractPathROI.java#L217), which sets the scale as 1,1. 

Found in 0.7 and 0.6.

alanocallaghan:
Resolve #2158

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 * Drawing/moving annotations can cause jiggly class list scrolling behavior (https://github.com/qupath/qupath/issues/2101)
 * Removing child objects can give an unexpected exception in scripts (https://github.com/qupath/qupath/issues/2111)
 * Nested detections render poorly when filled (https://github.com/qupath/qupath/issues/2112)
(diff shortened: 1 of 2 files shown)


petebankhead:
Thanks, looks good - I use updated the changelog to link to the bug report.

Issue #1055, opened 2022-09-12
Support saving & reloading settings when training pixel classifiers


petebankhead:
## Feature request

**Is your feature request related to a problem? Please describe.**
When training a pixel classifier, various settings can be customized (including features). These automatically return to their defaults when the dialog is closed.

See original discussion at https://forum.image.sc/t/change-default-settings-on-pixel-classifier/70371

**Describe the solution you'd like**
A more efficient way to set these values than through clicking in the UI.

**Additional context**
This could be through a saved parameter file, or recovering the settings from an existing saved classifier (reloading a saved classifier is also a frequent request).

Neither is entirely straightforward, given that the available settings change based upon image properties and not all settings are currently serialised in the JSON for a classifier.

Potentially, this request could be relevant for object classifiers and density maps.

drbergman:
As an intermediate step, could qupath come natively with support for json slurper? So that we can run `import groovy.json.JsonSlurper` to make for programmatic editing of classifiers?

petebankhead:
QuPath comes natively with Gson - would that do instead?

drbergman:
oh probably? I figured out how to add groovy-json-4.0.19.jar as an extension, so I'm using that now

roynov01:
I understand that resolution and available channels are image-specific, which makes it tricky to define universal presets.

That said, I was wondering if it might be possible to add support for custom feature/scale presets that are independent of image properties? For example, allowing a user-defined entry in the Features dropdown (alongside "Default multiscale features") with all 12 features and scales like [0.5, 1, 2].

Right now, having to click the same 20 buttons every time is both time-consuming and frustrating — being able to save and reuse a custom preset would really improve the workflow.

PR #2162, opened 2026-07-29
Improved pixel classification UI

petebankhead:
This PR extensively refactors the pixel classifier UI, adding many improvements.

_Important! This all requires more testing and refinement!_

## New dialog with tabs

The preview image is now part of a tab pane.

Screenshot 2026-07-29 at 14 56 41

In addition to a preview image, this provides access to other useful info that was previously hidden under advanced options (if it was visible at all in the UI).

Screenshot 2026-07-29 at 14 56 45

## Easy access to feature importance

OpenCV's `RTrees` classifiers can output feature importance values.
Previously these could be logged, but only if training an `RTrees` classifier for prediction.

Now, the `Features` tab lists the features that are chosen, and an `Update feature importance` button calculates importance values for display. The table can be sorted to easily see a ranking.

Screenshot 2026-07-29 at 14 57 09

## Cross-validation metrics

Previously, there was no straightforward way to determine the accuracy of a classifier during training, other than by subjectively assessing if it seemed to be getting better or worse.

Now, confusion matrices can be generated using cross-validation.

**Warning! For single-image training, these will be overly-optimistic!**
When training from a single image, the cross-validation is calculated at a pixel level. This means that you might train on one pixel and validate on a neighboring pixel. Features will be highly correlated, and the results will likely seem better than they should.

*However*, with `Load training` you can train across multiple images. In this case, cross-validation splits across different images. The metrics should therefore be more meaningful, and help judge how changes impact a classifier even on an annotated image that isn't currently open.

Screenshot 2026-07-29 at 14 57 50

## Smaller dialog by default
The extra features don't have to take up a lot of screen space.
They are hidden by default, and you can toggle them on and off.

Screenshot 2026-07-29 at 14 56 31

## Setting persistence (within a QuPath session)
Previously, the pixel classification dialog was created anew every time *Classify → Pixel classification → Train pixel classifier...* was called.

This meant that it 'forgot' previous settings. This wasn't easy to address, because the training dialog took some information from the current image (i.e. channel names and resolutions) at the time it was created. This meant that if you wanted to train a classifier for a completely different image, you would just close the dialog and open it again: the new dialog would be initialised for the current image.

The bad thing was that sometimes you might customise a lot of settings, and opening a new dialog meant restoring all of those. This was quite a common complaint https://github.com/qupath/qupath/issues/1055

The current PR doesn't support saving settings in preferences, however it *does* make settings persistent within a QuPath session. If you close the dialog it is only hidden, so that when you run *Classify → Pixel classification → Train pixel classifier...* it reappears.

The problem remains that you might switch between completely different images (e.g., different channels or resolutions). When an incompatibility is detected, a warning dot appears with a tooltip to explain the situation.

Screenshot 2026-07-29 at 15 05 41

## Multiscale features 3D

Previously, there was a weird drop-down menu for `Features` that contained a single item: `Default multiscale features 2D`.

The reason what the QuPath's pixel classifier was designed to support different ways to calculate features... it just didn't include any alternatives.

Now, `Multiscale features 3D` has been added. This makes it possible to calculate common features (e.g., for colors, ridges, blobs) in 3D. 

*Note that QuPath wasn't designed primarily for 3D... currently, this doesn't scale to huge images but it should be fine with small stacks (e.g., a couple of hundred MB). For really large images, you probably want a more specialised tool, like ilastik*.

Screenshot 2026-07-29 at 15 04 39

## See the JSON

The classifiers can be saved as JSON. If you *really* want to see the details, you can view what is stored in the JSON with syntax highlighting.

Screenshot 2026-07-29 at 14 57 55

## Lots more details

There are many other smaller changes... with potentially more to come in the future.


alanocallaghan:
1. The left/right buttons for splits were not visible by default, I had to make the window larger first, although I've not managed to reproduce this so it may have been caused by me already resizing the window...
2. After closing and re-opening, the window is larger. I prefer the larger size as things feel more well-space, but it is something to be aware of. Also generally I'm on a high res display so am happy to have larger windows generally so we should consider being more adaptive to this.
3. I like the persistence, but it would also be nice to have an easy way to reset to default, or to open a new window starting from default as before.

It is really good overall; it would be nice to be able to edit the classifier parameters in the classifier tab but it makes basically no difference and will presumably add a decent amount of code so I wouldn't bother.

petebankhead:
Thanks for checking!

1. I tried to replicate the 'splits' issue, but failed... although did manage to make the buttons jump to the left in a way that I couldn't reproduce. So there's definite weirdness there that might be fixable [inside `GuiTools`](https://github.com/qupath/qupath/pull/2162/changes/2afab0938c5c045545625c8fde153373b93fa76e#diff-1e6ff224f4387e34da57d27064e50b3cf121dea9ccd3fa408adc4950d4267502), but I'm tempted to leave it for now because I can't figure it out.

2. Do you mean after pressing the `More` button? Then I get a slightly clunky resize, because `Stage.sizeToScene()` is being called. If this button isn't ever pressed, I don't see a resize. Current behavior is more a product of luck than design, but I don't have a clear idea how to improve it.

3. I wondered about this, but couldn't figure out a good way. Options I thought of:
   * Hold down shift when running the command (but I don't know how to get the shift state)
   * Add a reset button somewhere in the UI (but where?)
   * Ask the user whether to keep the settings when they press to close the dialog (is that more helpful or annoying?)
  
Do any of these seem worth it to you, or is there a better way?

alanocallaghan:
For 1. yeah I'd just leave it, if it happens again I'll dig but it was pretty minor. For 2. I mean open the window, click more, do stuff, close window, re-open window. To be clear the new size is good, but it is noticably bigger.

For 3, I think a reset would be best but also not sure where. Maybe menubar?

petebankhead:
Merging this, while agreeing we should revisit these things in a moment of inspiration.

PR #2166, opened 2026-08-05
Simplify use of include-extra

petebankhead:
Support for an `include-extra` file made it easier to set up QuPath for extension development: https://github.com/qupath/qupath/pull/1710

The contents might look like this:
```
[includeBuild]
../qupath-extension-djl

[dependencies]
io.github.qupath:qupath-extension-djl
```

The `[includeBuild]` is needed to include the build (naturally), but this alone wouldn't get the extension picked up properly within QuPath. That's where `[dependencies]` was required.

The trouble is that the need for 2 lines added in 2 different sections is annoying if you have a long list of extensions.

This PR makes it possible to dispense with the sections, and instead use
```
../qupath-extension-djl
io.github.qupath:qupath-extension-djl
```

Basically, we try to figure out whether a line refers to something for `includeBuild`, `dependencies` or (rarely-used) `includeFlat`, by checking for the existing of a file/directory and (if not found) presence of a `:`.

The old strategy of using sections should still work (if you want to avoid the guesswork), but I think the new strategy is handy for me at least. I tend to have a long list of all my extensions inside `include-extras.properties`, then comment out the ones I don't currently need.. and doing this in 2 consecutive lines is much easier than finding the corresponding lines in 2 separate sections.

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -1,4 +1,5 @@
 import org.apache.tools.ant.taskdefs.condition.Os
+import java.nio.file.Paths
 
 pluginManagement {
     plugins {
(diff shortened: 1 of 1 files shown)


alanocallaghan:
Looks good. Small question for my own sake: is the below effectively a switch with a `default`, or?

```groovy
            "[includeBuild]" -> search = searchIncludeBuild
            "[includeFlat]" -> search = searchIncludeFlat
            "[dependencies]" -> search = searchDependencies
            else -> {
                [...]
            }
```

petebankhead:
I believe so, Kotlin-style: https://kotlinlang.org/docs/control-flow.html#when-expressions-and-statements

PR #2167, opened 2026-08-05
Ask to remove projects that are not found

alanocallaghan:
Very minor QoL improvement probably only really appreciated by me


petebankhead:
Looks good, I think it helps. Any objections to me inserting this?

```java
.contentText(MessageFormat.format(QuPathResources.getString("QuPathGUI.cannotFindProject"),
								uri == null ? null : URLDecoder.decode(uri.toString(), StandardCharsets.UTF_8)))
```

I was getting the ol' `%20` in my dialog, and elsewhere we do some decoding to reduce that.

alanocallaghan:
Looks good yeah

PR #2169, opened 2026-08-05
Escape dash in regex

alanocallaghan:


petebankhead:
Closing, fixed in https://github.com/qupath/qupath/pull/2173

PR #2178, opened 2026-08-17
Clean up contributing doc

alanocallaghan:
I was looking for a job and was disappointed to see the ad is no longer live.


petebankhead (approved):
Please stop looking for other jobs!
Otherwise, looks good :)

PR #2182, opened 2026-08-20
Default to the extension of the currently-selected filter in snapshots

### LINKED ISSUE #2181: Export snapshot fails silently on Linux if the extension is not part of the filename
* [x] I've definitely found a bug (it you're not sure, please use [image.sc](https://forum.image.sc/tags/qupath) instead)
* [x] I've checked https://qupath.github.io for a new release that might already have fixed the issue
* [x] I've checked the [Changelog](https://github.com/qupath/qupath/blob/master/CHANGELOG.md) to see if the bug has already been fixed in the next release
* [x] I've checked for existing GitHub issues describing the same problem

### Describe the bug
On Linux, export snapshot commands only works if the extension is specified in the filename and fails silently otherwise.

### To Reproduce
1. Go to File -> Export snapshots... -> <Any of the 4 commands>
2. If the file format doesn't end with .png, .tif or .jpg a file is saved as expected. If not, no file is saved and running the command fails silently.

### Expected behavior
The command should add the extension automatically to save the snapshot. If not possible it should output a warning asking the user to add the desired extension into the filename.
This behavior is different on Windows, where the extension is automatically added to the provided filename.

 - OS: Ubuntu 24.04.2 LTS
 - QuPath Version: 0.7.0 (also tested with 0.6.0)

### Additional context
On Linux, if the provided name ends with .png, .tif or .jpg, the file will be saved in this format whatever the selected output format in the GUI.

alanocallaghan:
Resolve #2181 

Draft PR as untested on anything but Linux

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@
 * Adding single points can cause unhelpful 'empty' point annotations to remain (https://github.com/qupath/qupath/issues/2155)
 * Calling `ImageServer.readRegion(RegionRequest)` can return the wrong pixels if `RegionRequest.getPath()` is wrong (https://github.com/qupath/qupath/issues/2164) 
 * The image overview doesn't rescale for very 'tall' images (https://github.com/qupath/qupath/issues/2163)
+* Export snapshot on Linux fails if the extension is not part of the filename (https://github.com/qupath/qupath/issues/2181)
 
 ### Dependency updates
 * Bio-Formats 8.5.0
--- a/qupath-gui-fx/src/main/java/qupath/lib/gui/commands/Commands.java
+++ b/qupath-gui-fx/src/main/java/qupath/lib/gui/commands/Commands.java
@@ -977,7 +977,11 @@ public class Commands {
 		lastSnapshotDirectory = fileOutput.getParentFile();
 
 		String ext = GeneralTools.getExtension(fileOutput)
-				.orElse(chooser.getSelectedExtensionFilter().getExtensions().getFirst());
+				.orElse(null);
+		if (ext == null) {
+			ext = chooser.getSelectedExtensionFilter().getExtensions().getFirst();
+			fileOutput = new File(fileOutput + ext);
+		}
 		List<ImageWriter<BufferedImage>> compatibleWriters = ext == null ? Collections.emptyList() :
 				ImageWriterTools.getCompatibleWriters(BufferedImage.class, ext);
 		if (compatibleWriters.isEmpty()) {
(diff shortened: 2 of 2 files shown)


petebankhead:
What happens on Linux? If you type `my-file` and select `.png`, does it write `my-file` or `my-file.png`?

In any case, this seems fine on macOS: it wasn't fussy about requiring an extension before, and I don't observe any change after this PR.

On macOS, it *always* appends the selected extension from the chooser, even if the user explicitly types a filename that contains a different extension. For example, a user typing `my-file.bmp` would get `my-file.bmp.png` on macOS. It looks funny, but I think it's reasonable behavior.

If the behavior is ok on Linux and Windows, then I think it's fine to merge - but we should check the 'user gives an unexpected extension' scenario.

(Minor point: we request the extension [again at line 992](https://github.com/alanocallaghan/qupath/blob/195e497decc96e435de46f40b8d310df5c6eee6d/qupath-gui-fx/src/main/java/qupath/lib/gui/commands/Commands.java#L992), which looks redundant.... but might not be, if Linux/Windows is actually allowing the user to sneak in a file extension that we hadn't explicitly allowed.)

petebankhead:
Addendum: I've just tried writing `sneaky.ome` on macOS, with `TIFF` selected. I do indeed get `sneaky.ome.tif` as the output, using the `OMEPyramidWriter`. It isn't exactly intended behavior, but I can't think of any harm in it.

alanocallaghan:
Yeah my-file becomes my-file.png rather than my-file (which would fail because we dont know what image writer to use with no extension, which I think is good behaviour).

The double extension thing is not ideal but the "real" double extensions make me think it's fine to pass some nonsense through occasionally

petebankhead:
Intriguing, if `fileOutput` doesn't have an extension then from looking at the code I can't see where one would be added. If I run the script

```groovy
var path = "/path/to/anything"
var img = GuiTools.makeSnapshot()
var writer = ImageWriterTools.getCompatibleWriters(BufferedImage.class, "png")[0]
writer.writeImage(img, path)
```
then I get `anything` and not `anything.png`. This PR only ensures we get a sensible `ImageWriter`, but the `ImageWriter` (at least on macOS) doesn't seem to add its own extension.

Anyway, conclusion remains the same: behavior on macOS is fine with this PR.

alanocallaghan:
Oh true yeah, I guess my DE was just smart enough to twig it was an image and open it anyways. Will test on windows and remedy that

alanocallaghan:
So the current main on Windows behave the same as Mac; this is a Linux-specific issue.

Tested this branch and it correctly produces a file with the correct extension and should therefore behave much the same as the other platforms.

petebankhead:
I confirm this PR still seems fine on macOS. Any reason not to merge it?

alanocallaghan:
Nope, apologies forgot to re-test on Linux until now

PR #2183, opened 2026-08-28
Split `OpenCVClassifiers`, reduce dependence on `StatModel`

petebankhead:
`ObjectClassifier` and `PixelClassifier` interfaces are provided to make object and pixel classifier implementations flexible.

However, _interactively trainable_ object and pixel classifiers were *not* flexible because they required an instance of `OpenCVStatModel`, which would invariably wrap an OpenCV `StatModel` class.
This was to standardize training based on `Mat` and JSON serialisation.

The logic was implemented within a mega `OpenCVClassifiers` class, which contained internal classes for specific `StatModel` implementations.

This PR 
1. Splits the mega class to make it easier to manage.
2. Revises the `OpenCVStatModel` interface to remove the `StatModel` dependency.

The goal of 2 is to reduce the 'OpenCV' element to requiring that the training and prediction data is provided as `Mat`, while relaxing the requirement that the model itself must be a `StatModel`.
This would allow the use of different interactively trainable models using different machine learning libraries, as long as we can convert the data sensibly.

It's currently a draft PR because this needs more thought, and `OpenCVStatModel` should be renamed to something more general.

Of note, any new `OpenCVStatModel` will need to be JSON serialializable, which will be one of the tricky things to integrate.

> Sidenote: The draft code also adds `EnsembleClassifier`, but we should probably drop this as having alternatives to `StatModel` means we no longer need to try to squeeze very bit of performance we can out of OpenCV's limited ML capabilities.


petebankhead:
Major refactoring. There are now 3 key interfaces/classes arranged as follows:

`PredictionModel` &rarr; `TrainableModel` &rarr; `OpenCVStatModel`

* When loading a pre-trained pixel or object classifier, only a `PredictionModel` is required - so this is kept simple.
* For interactive training, the additional methods of `TrainableModel` are required.
* `OpenCVStatModel` extends `TrainableModel` and is a close match to what was previously the only implementation.

The idea is that new `PredictionModel` or `TrainableModel` instances could be made JSON serializable via `OpenCVTypeAdapters`, and therefore provide drop-in replacements... although this doesn't currently work.

So the PR remains a draft until I get to the bottom of that. Additionally, we'll need a way to add the extra model types to the pixel and object classifier UIs.

petebankhead:
I've now added tests to check pixel/object classifier serialisation and deserialization.

A script used to create the test data in v0.7.0 is at https://gist.github.com/petebankhead/797fe63ae7ec5f4866231f4176c3c5a7

petebankhead:
`OpenCVStatModel` has become `qupath.opencv.ml.models.statmodel.TrainableStatModel` and the actual wrappers to `StatModel` have been made package-private to simplify the API (and give us more flexibility to change things).
 
To achieve that, `PredictionModel` and `TrainableModel` have had some extra methods added to extract UI-worthy model details and feature importance values (if available).

So the general idea for adding a different kind of model:
 
* Create a new implementation of `PredictionModel` or `TrainableModel` (trainable being more interesting)
  * Ensure that it's JSON-serializable.
* Call `OpenCVTypeAdapters.registerPredictionModel(Class)` so that it can be used within object and pixel classifiers
* (Later) somehow get it into the training windows for object or pixel classification

PR #2185, opened 2026-08-29
Ignore brush/wand/point ctrl clicks

petebankhead:
Ignore viewer clicks with ctrl (or cmd) key pressed and the brush, wand or points tool active.

This makes it possible to bring a viewer into focus without inadvertently creating a new annotation.

It's especially important when a different window was focused and any of these tools were active.

The user often needs to bring a viewer into focus so they can interact with it (and even have keypresses recognized), and this needs to be possible without accidentally creating an annotation that then needs to be deleted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 * Support to close more windows with Ctrl/Cmd+W (https://github.com/qupath/qupath/issues/2107)
 * Lots more strings externalized (https://github.com/qupath/qupath/pull/2104)
 * Ask to remove projects that are not found (https://github.com/qupath/qupath/pull/2167)
+* Hold down `Ctrl` (`Cmd` on macOS) when clicking viewer to focus it without creating an annotation with the brush, wand or points tool (https://github.com/qupath/qupath/pull/2185)
 
 ### Bug fixes
 * StackOverflowError at ROITypeAdapters.writeCoordinates when exporting some geometries (https://github.com/qupath/qupath/issues/2115)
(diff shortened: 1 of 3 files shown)


petebankhead:
This really annoyed me when training a pixel classifier with lots of annotations and 'Auto-set' selected for the classification. I would adjust the brightness/contrast, then want to return to the viewer.

Because the unwanted annotations that I created every time I brought the viewer into focus were classified, they would trigger the classifier to be retrained.

My workaround was to click outside the image, but this isn't always possible (e.g., when zoomed in a lot, so that image completely fills the viewer).

MichaelSNelson:
For what it's worth, I think I used spacebar for this when running into similar issues.

petebankhead:
Yeah, the spacebar was supposed to solve this, but it doesn't. There's no way to query whether the spacebar is pressed whenever you like in JavaFX, but rather you need to catch the event that occurred when it was pressed, and then wait for the event that signals it's released.

That's fine if you only have one window, but if you have multiple windows (or users might even switch to other apps, and press the spacebar in them, and keep it down when switching back to QuPath...) then it becomes difficult / impossible to have a totally reliable way to handle this.

Additionally, if the viewer isn't in focus and instead some other control that handles the spacebar a different way *is* in focus (e.g., a text field), then accidental inputs can still occur.

This means we're limited to only what we can [query directly on the `MouseEvent`])https://openjfx.io/javadoc/25/javafx.graphics/javafx/scene/input/MouseEvent.html) and which doesn't correspond to a key we're already using for something else.

PR #2191, opened 2026-09-06
Add command finder to header bar

petebankhead:
This makes use of the [`HeaderBar`](https://docs.oracle.com/en/java/java-components/javafx/26/docs/javafx.graphics/javafx/scene/layout/HeaderBar.html) preview feature of JavaFX, in anticipation of it being a full feature in JavaFX 27.

It adds a version of *View → Show command list* to the header bar, making it more discoverable.

The search box can be focused by pressing `Ctrl/Cmd+F` and, in combination with `Up`/`Down`/`Enter` makes many QuPath commands easily runnable from the keyboard.

Also, while it provides a searchable way to access all menu commands, the search box also can be used to access previously-run commands. To do this, type `Up` or `Down` without any search text entered. It maintains a `Set` of past commands, to keep the list short (i.e. duplicates are removed, while ensuring that the commands run most recently are at the top of the list).

header-search

While `HeaderBar` is in preview, using this requires
```
./gradlew run --args="-Djavafx.enablePreview=true"
```
but I'm hoping JavaFX 27 will make that simpler.

### Notes

* I could only test this on macOS so far. Feedback on behavior on other platforms would be very welcome.
* This PR removes the old feature of having a command finder attached to a viewer. I'm not aware of anyone using (or liking) that feature, and it adds complexity.

old-finder


zindy:
Hello,

just wanted to say, your instructions worked on Windows 11 and I do get the search text box.

There seems to be a bug I can mostly repeat by doing the following. Also I have other extensions (some badly written by me) installed, not sure whether that would cause the problem or not:
* open qupath with your gradlew run instruction,
* type `instan` and all you see `...` instead of the menu entry string. But the line points to the right entry.
* type space -> No content in table
* type backspace and the search results appear correctly.
* repeat the last two steps and you may get a different outcome (the `...` again, or not).

Cheers,
Egor

Screenshot 2026-09-08 204536

Screenshot 2026-09-08 205828

image

alanocallaghan:
Works great on Linux; interestingly 0on GNOME desktop it seems to remove the slightly unsightly default top bar:

without:

image

with:

image

petebankhead:
Thanks @zindy and @alanocallaghan :)

I saw the annoyingly narrow column issue during development, but haven't seen it with the PR and can't replicate it on macOS or Windows 10.

I've updated the PR to use Java 27, which might also help a bit.

Since it seems to be (at least) _mostly_ working, I'll merge it for now so that there's more time to be on the look-out for misbehaving columns.

PR #2195, opened 2026-09-09
Write v0.5 OME Zarrs

alanocallaghan:
This attempts to refactor the OME Zarr writer to write v0.5 OME Zarrs, and mostly succeeds. For example:

```groovy
import qupath.lib.classifiers.pixel.PixelClassificationImageServer
import qupath.lib.scripting.QP
import qupath.lib.images.writers.ome.zarr.PyramidalOMEZarrWriter

var server = new PixelClassificationImageServer(QP.getCurrentImageData(), QP.getProject().pixelClassifiers.get("temp"));

var pozw = new PyramidalOMEZarrWriter.Builder(server).build("/home/alan/tmp.ome.zarr")
pozw.writeImage()
```

writes a (seemingly correct) OME Zarr that can be read using https://github.com/BioImageTools/ome-zarr-fiji-java/

This is still a draft because Codec choice is not implemented, nor is writing attributes, and writing pyramidal zarrs relies on being able to read the first pass using bioformats, and bioformats doesn't read v0.5 Zarrs yet: https://github.com/alanocallaghan/qupath/blob/efa186d761f17fe8fbeba44a90ced82a88a798b3/qupath-extension-bioformats/src/main/java/qupath/lib/images/writers/ome/zarr/PyramidalOMEZarrWriter.java#L163-L169

Moving this from the bioformats extension would also require dealing with the fact that we have both OME tiff and OME Zarr writers etc, and I can't say I understand how to do that without just breaking everything for zero gain beyond a nominal support of 0.5 OME Zarr writing over 0.4


alanocallaghan:
...also this understandably breaks all of the lovely tests that Leo wrote

